PR #586, opened 2022-02-08
Make necessary changes for compatibility with cPaaS

Molter73:
## Description

This PR refactors some actions previously executed by the drivers compilation script into a separate wrapper script. This is done in order to allow builds done in the RH infrastructure to use a separate script for setting up the environment, while still maintaining most of the build logic in a common script executed by both infrastructures.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Building of drivers should keep working after the changes, so running CI with `build-legacy-probes`, `no-cache` and `dockerized-no-cache` should be enough.

### DIFF
--- a/kernel-modules/build/build-wrapper.sh
+++ b/kernel-modules/build/build-wrapper.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -exuo pipefail
+
 # This script handles decompressing the bundles tarballs and setting environment
 # variables needed by the kernel driver compilation script.
 
--- a/kernel-modules/build/build-wrapper.sh
+++ b/kernel-modules/build/build-wrapper.sh
@@ -1,0 +1,96 @@
+#!/usr/bin/env bash
+
+# This script handles decompressing the bundles tarballs and setting environment
+# variables needed by the kernel driver compilation script.
+
+KERNEL_SRC_DIR=""
+
+extract_bundle() {
+    local kernel_version="$1"
+
+    KERNEL_SRC_DIR="/scratch/kernel-src/${kernel_version}"
+
+    if [[ ! -d "$KERNEL_SRC_DIR" ]]; then
+        rm -rf /scratch/kernel-src/* 2> /dev/null || true
+        mkdir -p "$KERNEL_SRC_DIR"
+        tar -C "$KERNEL_SRC_DIR" -xzf "/bundles/bundle-${kernel_version}.tgz"
+    fi
+
+    [[ -f "${KERNEL_SRC_DIR}/BUNDLE_BUILD_DIR" ]] || {
+        echo "No BUNDLE_BUILD_DIR entry found in kernel source bundle!"
+        return 1
+    }
+
+    [[ -f "${KERNEL_SRC_DIR}/BUNDLE_UNAME" ]] || {
+        echo "No BUNDLE_UNAME entry found in kernel source bundle!"
+        return 1
+    }
+}
+
+export_env_variables() {
+    export bundle_uname
+    export bundle_version
+    export bundle_major
+    export bundle_distro
+    export kernel_build_dir
+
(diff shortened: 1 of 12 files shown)


kernel-modules/build/build-wrapper.sh
```diff
@@ -0,0 +1,105 @@
...
+    export bundle_version
+    export bundle_major
+    export bundle_distro
+    export kernel_build_dir
+
+    bundle_uname="$(cat "${KERNEL_SRC_DIR}/BUNDLE_UNAME")"
+    bundle_version="$(cat "${KERNEL_SRC_DIR}/BUNDLE_VERSION")"
+    bundle_major="$(cat "${KERNEL_SRC_DIR}/BUNDLE_MAJOR")"
+    bundle_distro="$(cat "${KERNEL_SRC_DIR}/BUNDLE_DISTRO")"
+    kernel_build_dir="${KERNEL_SRC_DIR}/$(cat "${KERNEL_SRC_DIR}/BUNDLE_BUILD_DIR")"
```
Stringy, line 46: I'm not familiar with this kind of bash. Does this basically mean that these variables are exported with the values set in this function? Apologies if silly question 😅

kernel-modules/build/build-wrapper.sh
```diff
@@ -0,0 +1,105 @@
...
+        if ! extract_bundle "${kernel_version}"; then
+            echo >&2 "Failed to extract kernel bundle for version '${kernel_version}'"
+            exit 1
+        fi
+
+        export_env_variables
+
+        failure_output_file="${FAILURE_DIR}/${kernel_version}/${module_version}/${probe_type}.log"
+        mkdir -p "$(dirname "$failure_output_file")"
+        if ! build-kos "$kernel_version" "$module_version" "${probe_type}" 2> >(tee "$failure_output_file" >&2); then
```
Stringy, line 74: where is the `build-kos` function? are we missing a `source` somewhere?

kernel-modules/build/build-wrapper.sh
Molter73: It does, yes. After they are defined here they will be picked up by any spawned subshells

kernel-modules/build/build-wrapper.sh
Molter73: This one is a little tricky and might need a little refactoring. There is a `kernel-modules/build/build-kos` script, [when the builders are created it gets copied into `/usr/bin`](https://github.com/stackrox/collector/blob/92bd6681510393e27e69778fe7dd9cf2e99637ac/kernel-modules/build/Dockerfile#L40) so I can consistently call it without worrying on what the path to it actually is.

Molter73:
@JoukoVirtanen I'll create the test branch now so we can compare errors, but since they are all in the `dockerized` build I wouldn't put too much weight on it, WDYT?

Molter73:
Looks like a bunch of bpf probes that used to work are now not being added to the final images, (the same errors are found when running on master but are somehow ignored), I'm looking into it.

Molter73:
For whatever reason, running `build-kos` as a separate script was causing the `make` command on it to exit prematurely, which was not the case when it was being called as a function. I have changed the script so that now, the wrapper scripts sources it and uses the `build_ko` function directly, this seems to fix the problem. I'm currently re-running CPaaS with these changes to verify everything works as expected.

Molter73:
The build on CPaaS worked properly, so we should be good to go.

JoukoVirtanen (approved):
LGTM!

Molter73:
I'm merging the PR, regarding the failures:
- Flatcar ITs were broken up to yesterday, fixed by #657.
- Dockerized kernel modules have been broken for some time now and have been removed as part of the checks by #637.
- The COS failure is due to a connectivity issue while running the tests, looks like a flake.

PR #602, opened 2022-02-22
Jv/rox 9457 enable core dumps

JoukoVirtanen:
## Description

If core dumps are enabled we should see them when Collector crashes. Core dumps should be enabled for integration tests and it should be possible to turn them on optionally.

## Checklist
- [x] See core dump when there is a purposeful SIGABRT locally
- [x] See core dump when there is a purposeful SIGABRT in CircleCI
- [x] Don't see core dump when core dumps are disabled locally

**Automated testing**
No added testing

## Testing Performed

See checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -865,6 +865,8 @@ jobs:
         name: "Running integration tests"
         no_output_timeout: 45m
         command: |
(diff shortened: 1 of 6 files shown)


Stringy (approved):
LGTM!

PR #617, opened 2022-03-17
ROX-9462 Collect baselines

erthalion:
Introduce a script to manage performance baselines numbers. The baseline
is essentially a json file, containing previous benchmark results. Its
size is limited, the exact threshold has to be figured out depending on
the actual variance of data.

The script supposed to collect baseline, and provide way to compare
current benchmark data with it. Due to unavoidable noise try to be fancy
and employ t-test for mean values to figure out if the difference we've
got is significant, or just a noise.

This change also introduces a slight change in `all-perf.json` format,
before it was a list of json documents concatenated together, like:

    {"aaa": 1, "bbb": 2, ...} {"aaa": 3, "bbb": 4, ...}

This was fine for jq, but it's not a valid json. Not it will be:

    [{"aaa": 1, "bbb": 2, ...}, {"aaa": 3, "bbb": 4, ...}]

jq apparently didn't notice a thing.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Only local testing.

### DIFF
--- a/.circleci/test-scripts/baseline/compare-and-update-baseline.sh
+++ b/.circleci/test-scripts/baseline/compare-and-update-baseline.sh
@@ -10,11 +10,7 @@ export BASELINE="${CI_ROOT}/test-scripts/baseline/"
     | sort \
     | awk -f "${BASELINE}"/format.awk > "${WORKSPACE_ROOT}"/benchmark.md
 
-"${BASELINE}"/main.py --update "${WORKSPACE_ROOT}"/all-perf.json
-
-# The baseline has to be updated only in the main branch, uncomment that before
-# merging.
-#
-#if [[ "$CIRCLE_BRANCH" == "master" || -n "$CIRCLE_TAG" ]]; then
-    #${BASELINE}/main.py --update ${WORKSPACE_ROOT}/all-perf.json
-#fi
+# The baseline has to be updated only in the main branch.
+if [[ "$CIRCLE_BRANCH" == "master" || -n "$CIRCLE_TAG" ]]; then
+    "${BASELINE}"/main.py --update "${WORKSPACE_ROOT}"/all-perf.json
+fi


.circleci/test-scripts/baseline/main.py
```diff
@@ -0,0 +1,289 @@
...
+
+    processed = [
+        {
+            "kernel": record.get("VmConfig"),
+            "collection_method": record.get("CollectionMethod"),
+            "timestamp": record.get("Timestamp"),
+            record["TestName"]: record.get("Metrics").get("hackbench_avg_time")
+        }
+        for record in data
+    ]
```
Molter73, line 182: Couldn't these two list comprehensions be combined if we moved the if condition of the first one to the last one?

integration-tests/integration_test.go
```diff
@@ -430,7 +432,7 @@ func (s *RepeatedNetworkFlowTestSuite) SetupSuite() {
 	s.clientIP, err = s.getIPAddress("nginx-curl")
 	s.Require().NoError(err)
 
-	totalTime := (s.sleepBetweenCurlTime * s.numIter + s.sleepBetweenIterations) * s.numMetaIter + s.afterglowPeriod + 10
+	totalTime := (s.sleepBetweenCurlTime*s.numIter+s.sleepBetweenIterations)*s.numMetaIter + s.afterglowPeriod + 10
```
Molter73, line 435: [nit] What happened to the spacing here? Seems like some auto-formatting might be playing some nasty tricks

Molter73 (approved):
Just some minor comments that wouldn't be blocker IMO. LGTM!

integration-tests/integration_test.go
erthalion: Looks like an interesting [quirk](https://github.com/golang/go/issues/1206) of gofmt. When I run `gofmt` directly over this file it doesn't report any difference, and it seems like it does remove spaces sometimes "to express binding strength".

.circleci/config.yml
```diff
@@ -1073,14 +1073,15 @@ jobs:
...
-          cat ~/workspace/benchmark.csv | sort | awk -v FS="," 'BEGIN{print "|Kernel|Method|Without Collector Time (secs)|With Collector Time (secs)|";print "|---|---|---|---|"}{printf "|%s|%s|%s|%s|%s",$1,$2,$3,$4,ORS}' > ~/workspace/benchmark.md
+          pip install --upgrade scipy google-cloud-storage
+
+          jq -s 'flatten' ${WORKSPACE_ROOT}/*perf.json > ${WORKSPACE_ROOT}/all-perf.json
+
+          export BASELINE="${CI_ROOT}/test-scripts/baseline/"
+          ${BASELINE}/main.py --test ${WORKSPACE_ROOT}/all-perf.json \
+            | sort \
+            | awk -f ${BASELINE}/format.awk > ${WORKSPACE_ROOT}/benchmark.md
+          ${BASELINE}/main.py --update ${WORKSPACE_ROOT}/all-perf.json
```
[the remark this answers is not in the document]
erthalion: It's not really a full re-write, there are going to be some follow up steps for sure. But I don't have anything against moving it into a script :)

erthalion:
There is only one thing that bothers me at the moment -- since the baseline file is simply an array of benchmark results, there is no explicit connections between `baseline_benchmark` and corresponding `collector_benchmark`. It works now via sorting results  by timestamps and correlating things in this way, which is a fragile approach. I believe this could be improved as a follow-up, when we get more data to see if it's a problem or not.

.circleci/test-scripts/baseline/compare-and-update-baseline.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -eo pipefail
```
Molter73, line 2: Might be worth it to add `-u` in order for the script to fail explicitly if some env variable isn't defined.
```suggestion
set -euo pipefail
```

.circleci/test-scripts/baseline/compare-and-update-baseline.sh
```diff
@@ -0,0 +1,20 @@
...
+    | awk -f "${BASELINE}"/format.awk > "${WORKSPACE_ROOT}"/benchmark.md
+
+"${BASELINE}"/main.py --update "${WORKSPACE_ROOT}"/all-perf.json
+
+# The baseline has to be updated only in the main branch, uncomment that before
+# merging.
+#
+#if [[ "$CIRCLE_BRANCH" == "master" || -n "$CIRCLE_TAG" ]]; then
+    #${BASELINE}/main.py --update ${WORKSPACE_ROOT}/all-perf.json
+#fi
```
Molter73, line 20: Why was this left as a comment? Is it replacing line 13 before merging the PR?

Molter73 (approved):
Re-approving after the latest changes. Added 2 comments but they are super-nit things that shouldn't prevent this from being merged.

.circleci/test-scripts/baseline/compare-and-update-baseline.sh
erthalion: Yep, that's the goal.

Stringy (approved):
re-approving, looks good!

PR #601, opened 2022-02-22
ROX-9409: sets DRIVER_NAME in collector CMakeLists

Stringy:
## Description

A detailed explanation of the changes in your PR.

Feel free to remove this section if it is overkill for your PR, and the title of your PR is sufficiently descriptive.

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Updated documentation accordingly

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

If any of these don't apply, please comment below.

## Testing Performed

TODO(replace-me)
Use this space to explain how you tested your PR, or, if you didn't test it, why you did not do so. (Valid reasons include "CI is sufficient" or "No testable changes")
In addition to reviewing your code, reviewers **must** also review your testing instructions, and make sure they are sufficient.

For more details, ref the [Confluence page](https://stack-rox.atlassian.net/wiki/spaces/StackRox/pages/855998488/Proposal+Explicitly+List+Testing+Steps+on+PRs) about this section.

PR #625, opened 2022-03-24
Configure ring buffer depth at build time

Molter73:
## Description

This PR leverages stackrox/falcosecurity-libs#11, which makes it possible for us to configure the desired ring buffer depth when compiling the collector userspace component and drivers. Right now the depth is set in 2 separate places, we might want to refactor our build system to make it possible to configure both the userspace and driver builds from the same point in the repo.

Merge order for the PR:
- [x] Merge #601 into master.
- [ ] Merge stackrox/falcosecurity-libs#11 into master
- [ ] Change this PR to point to the new commit in the master branch of stackrox/falcosecurity-libs and merge into master.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [ ] Double check the configured ring buffer depth is the same when building userspace and drivers.
- [ ] Rebuilding all drivers works.
- [ ] Drivers themselves work.


Molter73:
This PR is outdated, #1343 will already enable this.

PR #644, opened 2022-04-01
Push images to the public registry

gavin-stackrox:
## Description

Now that StackRox is open source 🎉 , there are a number of other repos and processes that depend on images being found under `quay.io/stackrox-io`. This PR pushes collector images there also. At some future point, it should be possible to remove the private docker.io/stackrox and quay.io/rhacs-eng registries but that would probably break other workflows and processes, so for now the change is yet another repo. (Which entails a somewhat kludgey 2nd login to quay.io in 100-push-images.sh).

## Checklist
- [x] Investigated and inspected CI test results

## Testing Performed

CI is sufficient + validate that images appear at quay.io/stackrox-io.

PR #645, opened 2022-04-04
Push builder image to quay

erthalion:
## Description

For building open sourced StackRox, builder image has to be pushed to
the public quay as well. Do this similarly to #644.

## Checklist

## Testing Performed

CI results are sufficient.

### DIFF
--- a/.circleci/builder/save-collector-builder-image.sh
+++ b/.circleci/builder/save-collector-builder-image.sh
@@ -38,6 +38,6 @@ for repo in "${image_repos[@]}"; do
         docker tag "stackrox/collector-builder:${COLLECTOR_BUILDER_TAG}" "${repo}:cache"
 
         docker image inspect "${repo}:cache" > /dev/null
-        "${WORKSPACE_ROOT}/go/src/github.com/stackrox/collector/scripts/push-as-manifest-list.sh" "${repo}:cache"
+        "${SOURCE_ROOT}/scripts/push-as-manifest-list.sh" "${repo}:cache"
     fi
 done
--- a/.circleci/builder/save-collector-builder-image.sh
+++ b/.circleci/builder/save-collector-builder-image.sh
@@ -3,11 +3,34 @@ set -eo pipefail
 
 BRANCH=$1
 
-if [[ "${COLLECTOR_BUILDER_TAG}" != "cache" ]]; then
-    # Push cache only if creating a new builder
-    docker push "stackrox/collector-builder:${COLLECTOR_BUILDER_TAG}"
-fi
-if [[ "$BRANCH" == "master" ]]; then
-    docker tag "stackrox/collector-builder:${COLLECTOR_BUILDER_TAG}" stackrox/collector-builder:cache
-    docker push stackrox/collector-builder:cache
-fi
+image_repos=(
+    "${DOCKER_REPO}/collector-builder"
+    "${QUAY_REPO}/collector-builder"
+    "${PUBLIC_REPO}/collector-builder"
+)
+
+for repo in "${image_repos[@]}"; do
+    if [[ "$repo" == "${PUBLIC_REPO}/collector-builder" ]]; then
+        # Relogin on quay is needed
+        docker login -u "$QUAY_STACKROX_IO_RW_USERNAME" -p "$QUAY_STACKROX_IO_RW_PASSWORD" quay.io
+    fi
+
+    if [[ "${COLLECTOR_BUILDER_TAG}" != "cache" ]]; then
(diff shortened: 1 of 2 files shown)


.circleci/builder/save-collector-builder-image.sh
```diff
@@ -3,11 +3,34 @@ set -eo pipefail
...
+
+        image="${repo}:${COLLECTOR_BUILDER_TAG}"
+        echo "Pushing image ${image}"
+        docker image inspect "${image}" > /dev/null
+        "${WORKSPACE_ROOT}/go/src/github.com/stackrox/collector/scripts/push-as-manifest-list.sh" "${image}"
+
+    fi
+
+    if [[ "$BRANCH" == "master" ]]; then
+        image="${repo}:${COLLECTOR_BUILDER_TAG}"
```
Molter73, line 30: [ultra-nit] This same assignment is done in line 22, maybe set `image` once outside of the `if` conditions and use it directly?

Molter73 (commented):
Thanks for this change @erthalion, aside from the few comments I left I would suggest re-running CI using the `build-builder-image` label in the PR to ensure the image is properly created and pushed.

.circleci/builder/save-collector-builder-image.sh
```diff
@@ -3,11 +3,41 @@ set -eo pipefail
...
+    if [[ "${COLLECTOR_BUILDER_TAG}" != "cache" ]]; then
+        # Push cache only if creating a new builder
+        echo "Pushing image ${image}"
+
+        # Only Docker image is build at this point, retag it for others.
+        # Tagging Docker image doesn't make any changes.
+        docker tag "stackrox/collector-builder:${COLLECTOR_BUILDER_TAG}" "${image}"
+
+        docker image inspect "${image}" > /dev/null
+        "${WORKSPACE_ROOT}/go/src/github.com/stackrox/collector/scripts/push-as-manifest-list.sh" "${image}"
```
Molter73, line 29: I think we can simplify this a bit.
```suggestion
        "${SOURCE_ROOT}/scripts/push-as-manifest-list.sh" "${image}"
```

.circleci/builder/save-collector-builder-image.sh
```diff
@@ -3,11 +3,41 @@ set -eo pipefail
...
+
+    if [[ "$BRANCH" == "master" ]]; then
+        echo "Pushing image ${repo}:cache"
+
+        # Only Docker image is build at this point, retag it for others.
+        # Tagging Docker image doesn't make any changes.
+        docker tag "stackrox/collector-builder:${COLLECTOR_BUILDER_TAG}" "${repo}:cache"
+
+        docker image inspect "${repo}:cache" > /dev/null
+        "${WORKSPACE_ROOT}/go/src/github.com/stackrox/collector/scripts/push-as-manifest-list.sh" "${repo}:cache"
```
Molter73, line 41: Same as previous comment.
```suggestion
        "${SOURCE_ROOT}/scripts/push-as-manifest-list.sh" "${repo}:cache"
```

Molter73 (approved):
Just a couple minor nitpicks, I also see the builder is being properly pushed to `quay.io/stackrox-io`, so everything is looking good IMO.

.circleci/builder/save-collector-builder-image.sh
```diff
@@ -3,11 +3,34 @@ set -eo pipefail
...
+image_repos=(
+    "${DOCKER_REPO}/collector-builder"
+    "${QUAY_REPO}/collector-builder"
+    "${PUBLIC_REPO}/collector-builder"
+)
+
+for repo in "${image_repos[@]}"; do
+    if [[ "$repo" == "${PUBLIC_REPO}/collector-builder" ]]; then
+        # Relogin on quay is needed
+        docker login -u "$QUAY_STACKROX_IO_RW_USERNAME" -p "$QUAY_STACKROX_IO_RW_PASSWORD" quay.io
```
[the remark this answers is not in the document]
Stringy: Could we have a comment here about this, because it's not that obvious at a glance as to why the relogin is needed.

.circleci/builder/save-collector-builder-image.sh
Stringy, line 41: I think we have `${SOURCE_ROOT}` for slight simplification of this path

```suggestion
        "${SOURCE_ROOT}/scripts/push-as-manifest-list.sh" "${repo}:cache"
```

Stringy (approved):
Just a couple of minor comments, but otherwise looks good

PR #648, opened 2022-04-05
Run syscall latency measurements on CI

Stringy:
## Description

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


.circleci/run-integration-tests.sh
```diff
@@ -7,18 +7,72 @@ BUILD_NUM=$3
...
+}
+
+function integration_tests_no_measurements() {
+    make -C "${SOURCE_ROOT}" integration-tests-repeat-network \
+                             integration-tests-missing-proc-scrape \
+                             integration-tests-image-label-json \
+                             integration-tests-baseline \
+                             integration-tests-benchmark \
+                             integration-tests \
+                             integration-tests-report
```
Molter73, line 31: I'm wondering why we don't have a `ci-integration-tests` target that simply lists all of the required targets as dependencies so we could do:
```sh
make -C "${SOURCE_ROOT}" ci-integration-tests
```

Molter73, line 31: I'm wondering why we don't have a `ci-integration-tests` target that simply lists all of the required steps as dependencies so we could do:
```sh
make -C "${SOURCE_ROOT}" ci-integration-tests
```

integration-tests/Makefile
```diff
@@ -22,43 +22,43 @@ docker-clean:
 tests: docker-clean
 	go version
 	set -o pipefail ; \
-	go test -timeout 30m -count=1 -v \
+	go test -timeout 60m -count=1 -v \
```
Molter73, line 25: We might want to bump the `no_output_timeout` value in the CI configuration to something like 75 minutes. Though I have to say that a full hour to run some tests seems a little excessive for regular testing, any way we could have this configurable? Also, if this is not failing in CI then we might not need to bump the timeout at all 🤷🏻‍♂️ 
https://github.com/stackrox/collector/blob/452355b4d3c90247d176f16c392d7529b72a753b/.circleci/config.yml#L983

integration-tests/container/perf/tools/do_syscall_64.py
```diff
@@ -47,9 +52,15 @@ def exit_handler(bpf, output, *args):
         # plain Exception so it is not possible to be more refined.
         print(f"[*] detaching failed: {e}")
 
+    print(f"possibly lost {g_lost_count} events")
```
Molter73, line 55: I may be missing this, but do we extract the logs for the container running this script as part of the artifacts?

integration-tests/integration_test.go
```diff
@@ -37,7 +37,7 @@ const (
 
 func TestCollectorGRPC(t *testing.T) {
 	suite.Run(t, new(ProcessNetworkTestSuite))
-	suite.Run(t, new(BenchmarkCollectorTestSuite))
+	// suite.Run(t, new(BenchmarkCollectorTestSuite))
```
Molter73, line 40: [nit] Maybe remove this if we don't need it anymore.

Molter73 (commented):
Looking good! Aside from a few comments, it looks like there might be some conflicts with #674, so a rebase might be in order.

integration-tests/Makefile
Stringy: This is purely a fail safe, and one that we inconsistently hit on CI when running the (fairly heavy weight) syscall latency measurements, so it feels reasonable to bump it. I think the no_output_timeout should also remain the same since the tests should be outputting some info pretty frequently and no output would suggest some breakage somewhere.

integration-tests/container/perf/tools/do_syscall_64.py
Stringy: yeah, this is capture and will end up in the integration tests log, so should be available.

integration-tests/integration_test.go
Stringy: I'm unsure if we need this test at all anymore since it'll just run the ProcessNetworkTestSuite, which is duplicated elsewhere. I'll remove the whole thing.

.circleci/run-integration-tests.sh
Stringy: Good point, I'll see about adding one

Molter73 (approved):
LGTM!

PR #649, opened 2022-04-05
Add CircleCI badge

erthalion:
## Description

It displays status of the last pipeline, which includes build and tests.


Molter73 (approved):
LGTM! There is a failure in CI that seems unrelated to the change, but please try re-running it before merging the PR just in case.

PR #654, opened 2022-04-08
Refactors pre-commit hooks to use https://pre-commit.com

Stringy:
## Description

Refactors the pre-commit hooks into a https://pre-commit.com config file to automate the installation & execution of the necessary hooks. 

## Checklist
~- [ ] Investigated and inspected CI test results~

**Automated testing**
  ~- [ ] Added unit tests~

## Testing Performed

Tested locally on a range of trivial formatting mistakes.


Molter73 (commented):
This is pretty cool, once you are done would you mind updating the [setting-up-git-hooks](https://github.com/stackrox/collector#setting-up-git-hooks) section in the README so that people wanting to contribute to the repo can set it up properly?

.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
```
Molter73, line 1: Can we add some comments to what the `ignore` and `select` values are? I think adding at least a short description on what the ignored items are would go a long way, it looks little cryptic otherwise.

Molter73 (commented):
As I said before, this is really cool. My only concern now that I've had some time to think about it is that our CI and the `pre-commit` hooks are now coming from 2 different sources for some of the checks, specifically shellcheck and clang-format were being run from `make` targets and now they come from a pre-commit hook. It is not a major concern, but definitely something to be aware of.

.flake8
Stringy: I've massively simplified this file, it now only ignores E501, which is a PEP8 line length error and is handled by the `max-line-length` field anyway. 

My thinking is that we can add to the ignore list if we find any rules that we're regularly violating (for good enough reasons) but we can get by with the default set to begin with.

Stringy:
> Looks cool indeed, thanks. I don't have any objection, but feel slightly uncomfortable by dependencies on some external repos for linting configuration. Do you think it's feasible to move them into the collector repo, or is it completely groundless?

Yeah I had similar concerns, but we're only deferring to those repos for scripts that run the linters rather than configuration, and we're pinned to a specific revision so we won't need to worry about those repositories updating/changing stuff underneath us.

Molter73 (commented):
Sorry I didn't try this before, but the circleci hook will not work because of [this issue](https://github.com/zahorniak/pre-commit-circleci/issues/4), we can either:
- Try forking the hook and supporting it ourselves.
- Creating a PR upstream to address it and wait for it to be merged in a newer version.
- Keep our own validation and follow up on this on a separate PR.

Stringy:
I have created a PR in the upstream that fixes the private orbs problem: https://github.com/zahorniak/pre-commit-circleci/pull/5

Molter73:
If it takes too long to be merged and we want to move forward with this PR, we could also consider using your fork as the source for the circleci hook.

Stringy:
The upstream PR has been merged now, and the hook is working as expected locally.

Molter73 (approved):
LGTM! Looking forward to getting this implemented in CI!

PR #660, opened 2022-04-12
Check 'circleci' command is available before halt

Molter73:
## Description

A few instances of CI has been failing when halting with a message stating that the command `circleci` is not found, this PR attempts to mitigate that error by wrapping the `circleci step halt` instruction in a script that retries it a few times.

The `circleci` command has also been replaced by `circleci-agent`, as described [here](https://discuss.circleci.com/t/documentation-for-circleci-agent/36594/3), `circleci step` is indirectly handled by `circleci-agent`, so removing the middle man might also help.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

I've run CI 4 times and no error on commands not being found have showed up.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -151,8 +151,8 @@ commands:
           if [[ ! -s ~/workspace/build-tasks ]]; then
             echo "Nothing was built, skipping this job"
 
-            circleci step halt
-            exit 0
+            "${CI_ROOT}/safe-halt.sh"
+            exit
           fi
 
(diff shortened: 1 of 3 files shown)


erthalion (approved):
It probably makes sense to add few comments into the script to justify retries on halt, but otherwise looks good, thanks.

Stringy (approved):
LGTM!

PR #665, opened 2022-04-19
ROX-9802: Replace docker repositories with quay

Molter73:
## Description

In order to remove docker hub images completely from the repository, references have been changed to point to the quay.io alternatives of the repositories.

The logic used to choose between stackrox-io and rhacs-eng is to keep anything that could be used by a community member in stackrox-io (i.e make targets, documentation) and anything used privately by CI in rhacs-eng, although they should be interchangeable so we might want to add some of the missing repos in stackrox-io and move on to use only that repository in a follow up PR.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Every modified CI step needs to be tested using the proper labels in the PR.
Some make targets might need to be run manually.


Molter73:
The dockerized check failures step is expected to fail, everything else seems to be working correctly so I'm opening the PR up for review.

docs/how-to-start.md
```diff
@@ -49,7 +49,7 @@ following configuration:
 version: "3.9"
```
Stringy, line 49: There's another reference that could do with being updated on [line 73](https://github.com/stackrox/collector/blob/mauro/ROX-9802-remove-dockerhub/docs/how-to-start.md?plain=1#L73) of this file

Stringy (commented):
There's an additional reference to the old repo in a comment in [integration-tests/container/jsonlabel/Dockerfile#2](https://github.com/stackrox/collector/blob/mauro/ROX-9802-remove-dockerhub/integration-tests/container/jsonlabel/Dockerfile#L2) but otherwise looks good!

docs/how-to-start.md
Stringy: how come it's not published to `quay.io/stackrox-io`? 

I think you're right that it makes sense to leave it as is if it would otherwise misguide users.

Stringy (approved):

PR #666, opened 2022-04-19
Re-add update-suppport-package CI step

Molter73:
## Description

The CI step seems to have been removed by accident during some refactoring

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [ ] Run PR with `test-support-packages` label.


Molter73:
The `quay-rhacs-eng-readonly` is needed in order to download the `apollo-ci` image as seen here:
https://github.com/stackrox/collector/blob/e7a3f81ea44e711e010c47ad9f488bf7138acfe2/.circleci/config.yml#L21-L25

I'll add it back in and we can work on cutting down on the number of contexts we have on a follow up PR (moving away from dockerhub is a good first step IMO #665)

Stringy (approved):

PR #668, opened 2022-04-20
Re-enable "CPaaS" image build (aka `ubi-image`)

robbycochran:
## Description

Changes
- Use working RH credentials and use an activation key instead of user/pass (https://access.redhat.com/solutions/5558771)
- Clarified package to repo attribution in the docker file and removed gzip from the final image.
- ubi:8.4 to ubi:8.5

## Checklist

**Automated testing**
  - ~[ ] Added unit tests~
  - ~[ ] Added integration tests~
  - ~[ ] Added regression tests~

## Testing Performed

Checked that ubi image build passes and dockerized probe builder works as expected. The kernel errors in the `dockerized-check-failures` job are all expected.


collector/container/Dockerfile.ubi
```diff
@@ -85,32 +87,24 @@ COPY --from=builder ${BUILD_DIR}/src/container/scripts /
...
+      microdnf -y install valgrind; \
     fi && \
     if [ "$ADDRESS_SANITIZER" = "true" ]; then \
-        microdnf -y install libasan; \
+      microdnf -y install libasan; \
     fi && \
     microdnf clean all && \
-    rpm --query --all 'curl' '*rpm*' '*dnf*' '*libsolv*' '*hawkey*' 'yum*' |\
-      xargs rpm -e --nodeps && \
+    rpm -e --nodeps $(rpm -qa curl '*dnf' '*solv*' '*rpm*') && \
```
robbycochran, line 107: The SC2046 hadolint exception is for this unquoted shell evaluation. `xargs` isn't installed in ubi-minimal by default and I didn't want to add it just for this line.

collector/container/Dockerfile.ubi
```diff
@@ -31,12 +27,18 @@ RUN dnf -y update && \
...
+    dnf config-manager --enable codeready-builder-for-rhel-8-x86_64-rpms && \
+    dnf -y install \
+        gmock-devel \
+        gtest-devel && \
+    dnf config-manager --enable amq-clients-2-for-rhel-8-x86_64-rpms && \
+    dnf -y install \
+        jsoncpp-devel && \
+    dnf config-manager --enable rhocp-4.6-for-rhel-8-x86_64-rpms && \
+    dnf -y install \
+        jq-devel && \
```
Molter73, line 41: I like this, it makes it quite clear which content sets are required for which packages. Would it make sense to use the `--enablerepo` flag in the install commands so the scope is even more clear?
```suggestion
    dnf -y --enablerepo codeready-builder-for-rhel-8-x86_64-rpms  install \
        gmock-devel \
        gtest-devel && \
    dnf -y --enablerepo amq-clients-2-for-rhel-8-x86_64-rpms install \
        jsoncpp-devel && \
    dnf -y --enablerepo rhocp-4.6-for-rhel-8-x86_64-rpms  install \
        jq-devel && \
```

kernel-modules/dockerized/Dockerfile
```diff
@@ -33,7 +31,8 @@ RUN subscription-manager register --username $REDHAT_USERNAME --password $REDHAT
 		git \
 		python3 \
 		kmod && \
-	ln -s /usr/bin/gcc /usr/bin/gcc-8
+	ln -s /usr/bin/gcc /usr/bin/gcc-8 && \
+	subscription-manager unregister
```
Molter73, line 35: Out of curiosity, why are we unregistering here but not in `Dockerfile.ubi`?

Molter73 (approved):
Just a couple of minor comments, but other than that LGTM!

kernel-modules/dockerized/Dockerfile
robbycochran: Because we need to copy the rhsm entitlements. `microdnf` doesn't support subscription management (as far as I can tell).

collector/container/Dockerfile.ubi
robbycochran: Done, thanks!

PR #670, opened 2022-04-25
Giles/rox 10009 custom ebpf probe

Stringy:
## Description

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


collector/lib/SysdigService.cpp
```diff
@@ -133,7 +133,10 @@ sinsp_evt* SysdigService::GetNext() {
   if (event->get_category() & EC_INTERNAL) return nullptr;
 
   // TODO (rc) not needed when syscall filtering implemented in the eBPF probe
-  if (useEbpf && !global_event_filter_[event->get_type()]) return nullptr;
+  // if (useEbpf && !global_event_filter_[event->get_type()]) {
```
Stringy, line 136: note: verify that this truly can be removed, and remove before merging.

Stringy:
This PR now contains the custom probe source, to consolidate collector-specific parts of the codebase and abstract ourselves slightly from the libs repo. Previous PR comments on the probe itself can still be viewed at stackrox/falcosecurity-libs#14

kernel-modules/build/build-kos
```diff
@@ -85,13 +95,11 @@ build_ko() (
...
-        fi
 
-        echo "Building collector eBPF probe for kernel version ${kernel_version} and module version ${module_version}."
-
-        KERNELDIR="$kernel_build_dir" BUILD_ROOT="${KERNEL_SRC_DIR}" make -j 6 -C bpf
-
-        collector_probe="bpf/probe.o"
+            echo "Building collector eBPF probe for kernel version ${kernel_version} and module version ${module_version}."
+            KERNELDIR="$kernel_build_dir" BUILD_ROOT="${KERNEL_SRC_DIR}" FALCO_DIR="$(pwd)/bpf" make -j 6 -C /probe
+            collector_probe="/probe/probe.o"
```
Molter73, line 101: [nit] We could eliminate duplicated code if we had a variable `probe_directory` set by default to `/probe` and changed to `bpf` when compiling RHEL 7 kernels, then the compilation command could be:
```make
KERNELDIR="$kernel_build_dir" BUILD_ROOT="${KERNEL_SRC_DIR}" FALCO_DIR="$(pwd)/bpf" make -j 6 -C "${probe_directory}"
collector_probe="${probe_directory}/probe.o"
```

The `FALCO_DIR` variable would be ignored by the catch-all probe and we could go back to having a single command for all eBPF probes, WDYT?

kernel-modules/probe/Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+# DEBUG = -DBPF_DEBUG
+
+# Begin StackRox
+ifeq ($(shell grep -sq "^\s*struct\s\+audit_task_info\s\+\*audit;\s*$$" $(KERNELDIR)/include/linux/sched.h ; echo $$? ), 0)
+	KBUILD_CPPFLAGS+= -DCOS_73_WORKAROUND
+endif
+# clang-7 does not support -fmacro-prefix-map
+KBUILD_CPPFLAGS:=$(filter-out -fmacro-prefix-map=%,$(KBUILD_CPPFLAGS))
+# End StackRox
```
Molter73, line 29: Think we could clean this up a little and make it look the same way as the upstream change? (We should be pulling that soon).

Molter73 (commented):
Just a few minor comments, I'm really happy to see progress being made on the custom probe!

kernel-modules/build/build-kos
Stringy: That seems a lot neater! It also looks like RHEL-7 is using the new probe at the moment, which is odd. I'll look into it when making these changes.

kernel-modules/probe/Makefile
Stringy: Good point - I'd completely forgotten about that change, I'll update it.

kernel-modules/probe/collector_probe.c
```diff
@@ -0,0 +1,280 @@
...
+ *        The function will exit 0 (zero) regardless of outcome.
+ *
+ * @param id the syscall id
+ * @param ctx the eBPF context as provided by the kernel
+ *
+ * @return 0 (regardless of outcomes)
+ */
+static __always_inline int exit_probe(long id, struct sys_exit_args* ctx) {
+  const struct syscall_evt_pair* sc_evt = NULL;
+  struct sysdig_bpf_settings* settings = NULL;
```
Molter73, line 247: Just a heads up, `sysdig_bpf_settings` has been renamed in the latest changes of falco, along with a number of sysdig references, so we will need to be renaming them as well pretty soon.

Molter73 (approved):
This is huge step forward for us, thanks @Stringy for all your hard work with this PR!

kernel-modules/probe/collector_probe.c
Stringy: Good point! I think we also need more metadata sections defined at the bottom of this file because they're all validated at runtime by the falco userspace IIRC

kernel-modules/probe/collector_probe.c
```diff
@@ -0,0 +1,340 @@
...
+    id = bpf_syscall_get_nr(ctx);
+  }
+
+  sc_evt = get_syscall_info(id);
+  if (sc_evt == NULL) {
+    return 0;
+  }
+
+  evt_type = sc_evt->exit_event_type;
+  drop_flags = sc_evt->flags;
```
Stringy, line 328: I've just realised that this doesn't quite work now, with the possibility of this being part of a sys_enter/sys_exit probe. I'll need to check the flags in case this even is unused.

collector/lib/SysdigService.cpp
Stringy: looks like this is still required because of RHEL 7 still using sys_enter and sys_exit. However, it doesn't seem to have an effect on other platforms. I'll update the comment to reflect this situation.

Stringy: Need to work out a good way of doing this, because it's still relevant to RHEL 7 , which is using sys_enter/sys_exit still.

Molter73: Maybe we can simply add an `isRHEL7` variable somewhere, set it at startup with a new heuristic and check it here alongside `useEbpf`? There might be an even easier solution, but it eludes me right now.

kernel-modules/build/build-kos
```diff
@@ -68,6 +71,11 @@ build_ko() (
             return 1
         }
 
+        # directory in which the probe source lives.
+        probe_directory="/probe/"
+        # resulting object file from the build
+        collector_probe="/probe/probe.o"
+
```
Molter73, line 78: Are these changes still needed? My understanding is that they were here to fallback to the previous version of the probe, but I don't see them being modified anymore.

kernel-modules/probe/.clang-format
```diff
@@ -0,0 +1,25 @@
+BasedOnStyle: Google
```
Molter73, line 1: I see this file is based on `collector/.clang-format`, would it make sense to use a symlink to that file so we don't need to maintain 2 versions of the same file? I know that might be troublesome for docker builds, but the tradeoff might be worth it.

Molter73 (approved):
LGTM!

kernel-modules/probe/.clang-format
Stringy: Might have to check actually how clang-format works because I have a feeling if we put it in the repo root directory then it'll affect all files recursively (except those with another .clang-format file in their local directory). If that is the case I'll move it to the repo root so we don't have to worry about symlinks

kernel-modules/build/build-kos
Stringy: Agreed, they aren't needed any longer. I'll adjust this.

collector/lib/SysdigService.cpp
Stringy: I'm sure there's actually an isRHEL7 method in the HostInfo class so I can probably just use that to enable this filter.

kernel-modules/probe/.clang-format
Stringy: Looks like that doesn't work, unfortunately. A symlink might be our only option, so my thinking now is to have the file at the root of the repository and then symlink that into `kernel-modules/probe` and `collector`. What do you think?

Stringy: Apparently I spoke too soon - putting the file in the root directory does appear to work

collector/lib/SysdigService.cpp
```diff
@@ -132,8 +133,15 @@ sinsp_evt* SysdigService::GetNext() {
...
 
-  // TODO (rc) not needed when syscall filtering implemented in the eBPF probe
-  if (useEbpf && !global_event_filter_[event->get_type()]) return nullptr;
+  HostInfo& host_info = HostInfo::Instance();
+
+  // This additional userspace filter is a guard against additional events
+  // from the eBPF probe. This can occur when using sys_enter and sys_exit
+  // tracepoints rather than a targeted approach, which we currently only do
+  // on RHEL7 with backported eBPF
+  if (useEbpf && host_info.IsRHEL76() && !global_event_filter_[event->get_type()]) {
```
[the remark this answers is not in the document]
robbycochran: Feel free to do this in a follow-up PR so that we verify there is no impact.

robbycochran:
We should also verify that the `build-legacy-probes` workflow is still working. Can you add that label and test rebuild? It might be worth bumping the parallelism. When this is merged we will rely on that path for building the updates for recently released kernels.

Stringy:
Performance regression fixed. Now to fix legacy builds. The changes to build-kos disregard the legacy sources, but this can be fixed easily enough.

Stringy:
dropping labels and adding `skip-integration-tests` now, following the full successful test run (https://app.circleci.com/pipelines/github/stackrox/collector/9277/workflows/2eaac76f-9e6b-4a6a-bc67-ea6416d0ad91) 

I need to revert the parallelism before we merge, so this is to avoid several more hours of CI execution

robbycochran:
> I need to revert the parallelism before we merge, so this is to avoid several more hours of CI execution

We have CircleCI credits to burn so might as well keep it at 32 for now.

kernel-modules/probe/Makefile
```diff
@@ -0,0 +1,61 @@
...
+		-I $(FALCO_DIR) \
+		-D__KERNEL__ \
+		-DKBUILD_MODNAME=\"collector\" \
+		-D__BPF_TRACING__ \
+		-Wno-gnu-variable-sized-type-not-at-end \
+		-Wno-address-of-packed-member \
+		-fno-jump-tables \
+		-fno-stack-protector \
+		-Wno-tautological-compare \
+		-O2 -g -emit-llvm -c $< -o $(patsubst %.o,%.ll,$@)
```
robbycochran, line 60: Let's remove `-g` or make it optional. Per @erthalion 's suggestion, I saw some [significant reduction](https://github.com/stackrox/collector/pull/723) in probe size when removing debug symbols on the current probes.

robbycochran (approved):
Awesome work @Stringy! LGTM!

kernel-modules/probe/Makefile
robbycochran: Feel free to remove in followup PR.

PR #641, opened 2022-03-31
Jv/rox 9961 create nightly collector circleci job

JoukoVirtanen:
## Description

Create a nightly job

## Checklist
- [x] Investigated and inspected CI test results
- [x] Updated documentation accordingly

PR #673, opened 2022-05-03
Bump Ubuntu VMs to 22.04 LTS in tests

Molter73:
## Description

Version 22.04 is the next LTS release for Ubuntu, this PR bumps the previous short term support VMs used in integration tests to this version instead.

## Checklist
- [ ] ~Updated documentation accordingly~

**Automated testing**
  - [ ] ~Added unit tests~
  - [ ] ~Added integration tests~
  - [ ] ~Added regression tests~

## Testing Performed

- [x] Integration tests on Ubuntu 22.04 LTS pass.
- [x] Automated performance PR comment is properly generated.
- [x] Dockerized tests work with eBPF and is not run with kernel module.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -915,7 +915,7 @@ jobs:
           sles-15-sp2-sap,
           ubuntu-1804-lts,
           ubuntu-2004-lts,
-          ubuntu-2110,
+          ubuntu-2204-lts,
           ubuntu-pro-1804-lts
         ]
       image_name:
@@ -1401,12 +1401,12 @@ workflows:
         vm_type: ubuntu-os
         matrix:
           parameters:
-            image_family: [ubuntu-1804-lts, ubuntu-2004-lts, ubuntu-2110]
+            image_family: [ubuntu-1804-lts, ubuntu-2004-lts, ubuntu-2204-lts]
             collection_method: [module, ebpf]
             dockerized: [false, true]
           exclude:
             # Failing due to GLIC 2.33 not available on ubi 8
-            - image_family: ubuntu-2110
+            - image_family: ubuntu-2204-lts
               collection_method: module
               dockerized: true
     - integration-test:
@@ -1502,8 +1502,8 @@ workflows:
         - test-ebpf-ubuntu-1804-lts
         - test-module-ubuntu-2004-lts
         - test-ebpf-ubuntu-2004-lts
-        - test-module-ubuntu-2110
-        - test-ebpf-ubuntu-2110
+        - test-module-ubuntu-2204-lts
+        - test-ebpf-ubuntu-2204-lts
         - test-ebpf-sles-15
         - test-module-sles-15
         - test-module-sles-12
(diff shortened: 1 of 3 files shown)


Molter73:
I've been having trouble with the new benchmarking script, since it is unable to handle both added and removed tests without manually running an update to the baseline file. However, because we only keep one copy of said file updating it for this PR would break the `integration-test-data` step in any other branch other than this one.

In order to make it easier to add and remove tests without need of running scripts manually or breaking CI on other branches, I'm trying to make it so any test that is not found in both the baseline and the new measurement is ignored by the reporting script, here is my train of thought for this:
- The script is only run after all integration tests have passed (enforced by our CI).
- If nothing was changed since the last baseline was created, all tests will be found in both the baseline and the new measurement.
- If a new test was added, there is no baseline to compare against until it gets merged to master, (it would be nice to have some metric reported for it but this could mean quite a significant change to the current script and is outside the scope of this PR). Once the new test is merged, it will be run in master and the metrics will begin to be accumulated in the baseline script.
- If a test is removed, it will show up in the baseline but there will be no new measurement to compare it against, we can ignore it, (again, it would be nice to have an indication the test was skipped, but falls off the current PRs scope). Once the PR with the removed test is merged to master, baseline updating will not persist any measurements for the removed test in order to clean it up.

The solution is not ideal, I admit the script was a lot safer before these changes, however I would personally prefer it to have the addition/removal of tests to be seamless, specially considering CI is already making sure tests have been properly run and the data should be available.

@JoukoVirtanen, I know you've been having a similar problem in #641, I believe this approach would also work with a reduced number of tests since it effectively filters out tests that are not found in both data sets, WDYT?

erthalion:
> In order to make it easier to add and remove tests without need of running scripts manually or breaking CI on other branches, I'm trying to make it so any test that is not found in both the baseline and the new measurement is ignored by the reporting script, here is my train of thought for this:
> 
> * The script is only run after all integration tests have passed (enforced by our CI).
> * If nothing was changed since the last baseline was created, all tests will be found in both the baseline and the new measurement.
> * If a new test was added, there is no baseline to compare against until it gets merged to master, (it would be nice to have some metric reported for it but this could mean quite a significant change to the current script and is outside the scope of this PR). Once the new test is merged, it will be run in master and the metrics will begin to be accumulated in the baseline script.
> * If a test is removed, it will show up in the baseline but there will be no new measurement to compare it against, we can ignore it, (again, it would be nice to have an indication the test was skipped, but falls off the current PRs scope). Once the PR with the removed test is merged to master, baseline updating will not persist any measurements for the removed test in order to clean it up.

Thanks for working on this, that would be my approach as well. I have only few suggestions in [a5a8687](https://github.com/stackrox/collector/pull/673/commits/a5a868758923516b807c69ed5b9bed8362de8e81), I think it still makes sense to verify some bits (without insisting on compatibility between baseline and the new data), what do you think? Those changes are just a proposal, of course -- feel free to adjust if needed.

Molter73:
I think the changes make sense, if the data gathering step works we can probably go ahead merging these changes (unless there is any other objection from someone else).

Stringy (approved):
LGTM!

PR #674, opened 2022-05-03
Have CI build all necessary QA containers

Molter73:
## Description

This PR makes it so all containers needed to run integration tests and benchmarking are built by our CI and pushed to quay.io periodically.

The containers will not be built in PRs if the build-qa-containers label is not present. Layer caching is enabled in the job in charge of building the containers.

This changes have been extracted from #665 as a pre-factor in order to make both PRs easier to review.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [ ] All QA containers are properly built.
- [ ] All integration tests run correctly.


integration-tests/executor.go
```diff
@@ -70,6 +70,15 @@ func ReadBoolEnvVar(env string) bool {
 	return e
 }
 
+func GetQATag(base_tag string) string {
```
Stringy, line 73: [nit] this might be more appropriately placed in the utils.go file

integration-tests/integration_test.go
```diff
@@ -400,9 +400,11 @@ func (s *RepeatedNetworkFlowTestSuite) SetupSuite() {
 	err = s.collector.Launch()
 	s.Require().NoError(err)
 
+	scheduled_curls_image := "quay.io/rhacs-eng/qa:" + GetQATag("collector-schedule-curls")
```
Stringy, line 403: another slight nit but this pattern is repeated so might benefit from its own helper

```go
func QaImage(name string, tag string) string {
    return name + ":" + GetQaTag(tag)
}

QaImage("quay.io/rhacs-eng/qa", "collector-schedule-curls")
```

Stringy (approved):
Just a couple of minor nits but otherwise looks good!

integration-tests/executor.go
Molter73: I had a pretty hard time deciding where to place this function and somehow completely missed `utils.go`, I'll move it there.

integration-tests/integration_test.go
Molter73: Makes sense, I'll make this change.

PR #677, opened 2022-05-06
ROX-10044: Add SHA256 checksum to support packages

Molter73:
## Description

Added a small step after creating the support package to have the sha256 checksum for said files to be output into an accompanying file with the same name and the `.sha256` extension. Content of the file looks like this:

```
130bfe6e3c95e924675714ac6cd31c7466b241fd27101bdaadd7ef3fb3138455  support-pkg-b6745d-latest.zip
```

So downloading it in the same directory as the support package and running `sha256sum -c support-pkg-b6745d-latest.zip.sha256` directly validates the file.

The checksum will also be added to the published index of packages for interested parties to be able to verify the contents of them.

## Checklist

**Automated testing**
  - [ ] ~Added unit tests~
  - [ ] ~Added integration tests~
  - [ ] ~Added regression tests~

## Testing Performed

- [x] Manually download and validate the generated support packages and checksum files.
- [x] Check the index.html is rendered correctly.

### DIFF
--- a/kernel-modules/support-packages/05-create-index.py
+++ b/kernel-modules/support-packages/05-create-index.py
@@ -94,12 +94,12 @@ def load_support_packages(output_dir, mod_md_map):
         except (FileNotFoundError, PermissionError):
             support_pkg_file_latest = None
 
-        support_pkg_sig = f'{support_pkg_file}.sig'
+        support_pkg_sig = f'{support_pkg_file}.sha256'
         if not os.path.isfile(os.path.join(mod_out_dir, support_pkg_sig)):
             # No signature available for package
             continue
 
-        support_pkg_sig_latest = f'{support_pkg_file_latest}.sig'
+        support_pkg_sig_latest = f'{support_pkg_file_latest}.sha256'
         if not os.path.isfile(os.path.join(mod_out_dir, support_pkg_sig_latest)):
             # No signature available for package
             support_pkg_sig_latest = None
--- a/kernel-modules/support-packages/05-create-index.py
+++ b/kernel-modules/support-packages/05-create-index.py
@@ -19,15 +19,15 @@ class SupportPackage(object):
                  file_name,
                  latest_file_name,
                  last_update_time,
-                 sig_name,
-                 latest_sig_name):
+                 chk_name,
+                 latest_chk_name):
         self.module_version = module_version
         self.rox_version_ranges = rox_version_ranges
         self.file_name = file_name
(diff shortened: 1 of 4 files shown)


Molter73:
I manually checked the generated sha256 checksum files and it looks to be working properly
```sh
[vagrant@fedora 1.0.0]$ sha256sum -c support-pkg-1.0.0-latest.zip.sha256
support-pkg-1.0.0-latest.zip: OK
[vagrant@fedora 1.0.0]$ sha256sum -c support-pkg-1.0.0-20220506142445.zip.sha256
support-pkg-1.0.0-20220506142445.zip: OK
[vagrant@fedora 1.0.0]$ cd ../b6745d795b8497aaf387843dc8aa07463c944d3ad67288389b754daaebea4b62/
[vagrant@fedora b6745d795b8497aaf387843dc8aa07463c944d3ad67288389b754daaebea4b62]$ sha256sum -c support-pkg-b6745d-latest.zip.sha256
support-pkg-b6745d-latest.zip: OK
[vagrant@fedora b6745d795b8497aaf387843dc8aa07463c944d3ad67288389b754daaebea4b62]$ sha256sum -c support-pkg-b6745d-20220506142732.zip.sha256
support-pkg-b6745d-20220506142732.zip: OK
[vagrant@fedora b6745d795b8497aaf387843dc8aa07463c944d3ad67288389b754daaebea4b62]$
```

The index file also looks OK to me (though I'm not a UI designer 🤷🏻‍♂️ ).
image

kernel-modules/support-packages/05-create-index.py
```diff
@@ -73,11 +91,28 @@ def load_support_packages(output_dir, mod_md_map):
         support_pkg_file_latest = 'support-pkg-%s-latest.zip' % (mod_ver[:6])
```
robbycochran, line 91: It looks like this truncation could create a collision if we were to have versions `1.10.1` and `1.10.10`.

robbycochran:
Should we add instructions for using the checksum after the note of the stable link? 

Something like **To check the integrity of the support package, download the checksum to the same directory as the support package and use `sha256sum -c checksum-file`.**

kernel-modules/support-packages/05-create-index.py
Molter73: Good catch! I completely missed this when we changed the versioning of drivers! Think I can easily fix it, but I will do so in a follow up PR so we don't need to re-review this one.

robbycochran (approved):

PR #682, opened 2022-05-10
Use absolute overhead, not relative

erthalion:
## Description

Benchmark numbers have high variance, which makes relative overhead vary
significantly as well. Use absolute overhead value numbers, as they seem
more stable.

## Checklist

## Testing Performed

Local testing with the baseline data, captured so far.

### DIFF
--- a/.circleci/test-scripts/baseline/format.awk
+++ b/.circleci/test-scripts/baseline/format.awk
@@ -3,6 +3,6 @@ BEGIN {
     print "|---|---|---|---|---|---|---|"
 }
 {
-    warn = ($7==1) ? " :question:" : "";
-    printf "|%s|%s|%s|%s|%s|%s|%s%s|%s",$1,$2,$3,$4,$5,$6,$7,warn,ORS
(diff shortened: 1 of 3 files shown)


.circleci/test-scripts/baseline/format.awk
```diff
@@ -3,6 +3,6 @@ BEGIN {
     print "|---|---|---|---|---|---|---|"
 }
 {
-    warn = ($7<=0.05) ? " :question:" : "";
+    warn = ($7==1) ? " :question:" : "";
     printf "|%s|%s|%s|%s|%s|%s|%s%s|%s",$1,$2,$3,$4,$5,$6,$7,warn,ORS
```
Molter73, line 7: [ultra-nit] Since this is a true/false check now, maybe we can do something like this?
```suggestion
    warn = ($7==1) ? ":green_circle:" : ":red_circle:";
    printf "|%s|%s|%s|%s|%s|%s|%s|%s",$1,$2,$3,$4,$5,$6,warn,ORS
```

Molter73 (approved):
Just a small nit no displaying results, it looks good otherwise.

PR #691, opened 2022-05-17
Keep original baseline data

erthalion:
## Description

The original implementation dropped benchmark data for all tests, that
are not present either in the baseline or in the test. This approach
leads to data overwriting in situations when different jobs have
different set of tests (e.g. a normal testing flow, vs extended
"nightly" run). To prevent that, keep the original set of baseline data
untouched from run to run.

As a side effect this means that "dead" tests that were removed at some
point will stay in the baseline until manually removed.

## Checklist

## Testing Performed
Local testing.


Molter73 (approved):
LGTM!

PR #700, opened 2022-05-20
Jv/3.6.x fix skip prepare legacy sources

JoukoVirtanen:
## Description

prepare-legacy-sources.sh should not run when a tag is used.

## Checklist

## Testing Performed

See above


Molter73 (approved):
In order to speed up release I'm approving this change, we will take a closer look at how to properly decouple code preparation from tag sweeping in a follow up pr for the master branch after discussing possibilities with the team.

PR #703, opened 2022-05-25
Add a make target to build full dev images

Molter73:
## Description

The new make target builds on top of the image-dev target, the kernel builders and the Dockerfile used to create the full images to create a collector image with embedded kernel modules and eBPF probe compiled by the system the target is run on.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

It would be appreciated if reviewers could run the new `image-dev-full` target in different systems to make sure it works on most developer environments.


Stringy (approved):
This is fantastic! Just a brief note that the script will need to be updated when the custom probe lands in master.

PR #713, opened 2022-06-01
ROX-11192 the collector can be built using podman

ovalenti:
## Description

We want to support podman as the container provider for the build process, and keep compatibility with Docker

## Checklist
- [ ] ~~Updated documentation accordingly~~

## Testing Performed
The build image and collector builds fine with this patch

### DIFF
--- a/builder/install/10-protobuf.sh
+++ b/builder/install/10-protobuf.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-curl -L "https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOBUF_VERSION}/protobuf-cpp-${PROTOBUF_VERSION}.tar.gz" | tar xz
+curl -L "https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOBUF_VERSION}/protobuf-cpp-${PROTOBUF_VERSION}.tar.gz" | tar xz --no-same-owner
 cd "protobuf-${PROTOBUF_VERSION}"
 cp LICENSE "${LICENSE_DIR}/protobuf-${PROTOBUF_VERSION}"
 


Molter73:
The change looks good, I'm testing the build right now. Could you add the `build-builder-image` label to the PR and re-trigger CI so we can double check everything works correctly?

builder/install/10-protobuf.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-curl -L "https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOBUF_VERSION}/protobuf-cpp-${PROTOBUF_VERSION}.tar.gz" | tar xz
+curl -L "https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOBUF_VERSION}/protobuf-cpp-${PROTOBUF_VERSION}.tar.gz" | tar xz --no-same-owner
```
erthalion, line 5: Could you elaborate more (maybe in the comment, or the commit message) in which way is it related to the podman vs docker compatibility?

ovalenti: Sure, this is related to running the container as non-root and the corresponding mapping to the actual user id.
If the [commit log](https://github.com/stackrox/collector/commit/f0facf5cc09077ac6efa628e2bea86a1c38e3764) should go further in the details, please tell me and I will elaborate.

ovalenti:
Rebased on master to work on depending PR

ovalenti:
@erthalion I re-triggered the CI pipeline as siggested and it looks like the failing test is related to a failing kernel module build, which is deactivated in _master_
`ERROR: modpost: "cc_mkenc" [/scratch/module-src/1.1.0/collector.ko] undefined!`

Would you like me to rebase to have it all green before approving ?

PR #715, opened 2022-06-02
Update ubi image version

erthalion:
## Description

Bump from 8.5 to 8.6

## Checklist

### DIFF
--- a/.circleci/dockerized-drivers/unify-images.sh
+++ b/.circleci/dockerized-drivers/unify-images.sh
@@ -22,8 +22,8 @@ fi
 
 if [[ "${COLLECTOR_DRIVERS_TAG}" != "${COLLECTOR_DRIVERS_CACHE}" ]]; then
     # Create an empty base image if we are not building the cache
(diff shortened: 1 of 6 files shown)


Molter73 (commented):
Think there are a few more places were we need to make this bump

https://github.com/stackrox/collector/blob/0fe760ad19f6f6d11054fa5603e3fd482a51075a/collector/container/Dockerfile#L3

This ones are not that important, since are just empty images that get discarded:
https://github.com/stackrox/collector/blob/0fe760ad19f6f6d11054fa5603e3fd482a51075a/.circleci/dockerized-drivers/unify-images.sh#L25-L27
https://github.com/stackrox/collector/blob/0fe760ad19f6f6d11054fa5603e3fd482a51075a/.circleci/dockerized-pull-cache-image/10-pull-cache-image.sh#L8

Also, could you add the `run-dockerized-steps` label before running CI to check those images are working properly? Some kernel build failures will happen, but I won't to make sure the steps still run correctly.

Molter73 (approved):
LGTM!

PR #725, opened 2022-06-13
Removes -g from default eBPF makefile build

Stringy:
Adds some documentation about a new procedure for handling MODULE_VERSION, both during development and for releases.

On top of this, MODULE_VERSION is incremented to 2.0.0-rc1 and `-g` is removed from the eBPF probe, as per #723


robbycochran:
You will need to bump the probe version for this to have an effect on all probes.

Molter73 (commented):
Just have a few comments on the release instructions with the new rc drivers, other than that it is looking good!

PR #731, opened 2022-06-17
Release branch/tag creation tool

Stringy:
Adds a tool for creating/pushing the necessary release tags/branches based on the procedure outlined in https://github.com/stackrox/collector/blob/05aa8b4045ec0ff19957a825b4be19764d4306a4/docs/release.md


utilities/release.py
```diff
@@ -0,0 +1,96 @@
...
+
+class Repository:
+    """
+    Encapsulates some logic for running git commands within the repository,
+    particularly related to creating release tags/branches.
+    """
+    def __init__(self, root='.', dry_run=False):
+        self.root = root
+        self.dry_run = dry_run
+        self.git = git.bake('--no-pager', C=root)
```
Molter73, line 17: This is sort of a nitpick, but maybe we can have the class save the current branch in the repo and check it out in the destructor. It's a minor thing but I personally don't like scripts that move things around and dump me somewhere different than where I was 🙂

utilities/release.py
```diff
@@ -0,0 +1,96 @@
...
+def main(version: VersionInfo, dry_run: bool, push: bool):
+    repo = Repository(dry_run=dry_run)
+    repo.reset()
+    repo.make_release_tag(version)
+    repo.checkout_release_branch(version)
+    repo.make_release_tag(version, patch='0')
+
+    if push:
+        repo.push_release_tag()
+        repo.push_release_branch()
```
Molter73, line 80: Ultra nitpick here, but I think pushing the branch first makes more sense, otherwise the tag could point to a non-existent commit

Molter73 (approved):
This looks really cool. Maybe as a follow up we could put the `Repository` class into its own module and refactor `tag-bumper.py` to use it

utilities/release.py
Stringy: Excellent point - I'll definitely put something like that in. Perhaps making it a context manager would be neat?

utilities/release.py
Stringy: this also doesn't actually work because I don't pass the versions in... that's what I get for only testing dry runs and local changes! Will fix!

Molter73 (commented):
Left a few more comments, but overall it is looking awesome!

utilities/release.py
```diff
@@ -0,0 +1,112 @@
...
+        try:
+            return self.git(command, *args).stdout.decode()
+        except ErrorReturnCode as e:
+            print(f'Failed to run "git {command} {" ".join(args)}": {e}')
+            sys.exit(1)
+
+
+def main(version: VersionInfo, dry_run: bool, push: bool):
+    with Repository(dry_run=dry_run) as repo:
+        repo = Repository(dry_run=dry_run)
```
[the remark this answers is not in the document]
Stringy: because I am an idiot! 😆

PR #738, opened 2022-06-22
ROX-11211-collector-ci-migration-images

JoukoVirtanen:
## Description

Migrate images job to Openshift CI

## Checklist

## Testing Performed

See above

### DIFF
--- a/.openshift-ci/build/Dockerfile.build_root
+++ b/.openshift-ci/build/Dockerfile.build_root
@@ -1,0 +1,8 @@
+FROM quay.io/stackrox-io/apollo-ci:collector-0.3.42
+
+# oc
+RUN sudo wget --no-verbose -O oc.tgz https://github.com/openshift/origin/releases/download/v3.11.0/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz
+RUN sudo tar -xf oc.tgz
+RUN sudo install openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit/oc /usr/local/bin
+RUN sudo rm -rf openshift-* oc.tgz
+RUN command -v oc
--- a/.openshift-ci/build/Dockerfile.drivers
+++ b/.openshift-ci/build/Dockerfile.drivers
@@ -1,0 +1,13 @@
+FROM quay.io/centos/centos:stream8
+
+ENV SRC_ROOT_DIR=/go/src/github.com/stackrox/collector
+WORKDIR $SRC_ROOT_DIR
+
+RUN cp "${SRC_ROOT_DIR}"/.openshift-ci/build/install-gcloud-sdk.sh /
+RUN "${SRC_ROOT_DIR}"/.openshift-ci/build/install-gcloud-sdk.sh /
+RUN ls
+RUN /install-gcloud-sdk.sh
+
+RUN gsutil -m cp -r 'gs://collector-modules/612dd2ee06b660e728292de9393e18c81a88f347ec52a39207c5166b5302b656/1.0.0/collector-3.10.0-1160.62.1.el7.x86_64.ko.gz' kernel-modules/
+
+RUN ls "${SRC_ROOT_DIR}/kernel-modules/"
--- a/.openshift-ci/build/install-gcloud-sdk.sh
+++ b/.openshift-ci/build/install-gcloud-sdk.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+dnf update -y
+dnf install epel-release -y
+
+tee -a /etc/yum.repos.d/google-cloud-sdk.repo << EOM
+[google-cloud-sdk]
+name=Google Cloud SDK
+baseurl=https://packages.cloud.google.com/yum/repos/cloud-sdk-el7-x86_64
+enabled=1
+gpgcheck=1
+repo_gpgcheck=1
+gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg
+       https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
+EOM
+
+cat -A /etc/yum.repos.d/google-cloud-sdk.repo
+
+dnf install google-cloud-sdk -y
--- a/.circleci/images/70-build-collector-slim-and-base-images.sh
+++ b/.circleci/images/70-build-collector-slim-and-base-images.sh
@@ -23,11 +23,13 @@ fi
     "-" \
     "$SOURCE_ROOT/${COLLECTOR_BUILD_CONTEXT}"
 
+make -C "$SOURCE_ROOT" container-dockerfile
+
 docker build \
     -t "${QUAY_REPO}/collector:${COLLECTOR_VERSION}-base" \
     -t "${QUAY_REPO}/collector:${COLLECTOR_VERSION}-slim" \
     -t "${PUBLIC_REPO}/collector:${COLLECTOR_VERSION}-base" \
     -t "${PUBLIC_REPO}/collector:${COLLECTOR_VERSION}-slim" \
     "${build_args[@]}" \
-    -f "${SOURCE_ROOT}/collector/container/Dockerfile" \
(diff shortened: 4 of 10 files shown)


ovalenti:
It's a good idea to use build-collector.sh in the mounted volume; since it avoids some of the "stale builder" image issues when updating this script.

I am wondering whether something is still using build-collector.sh except the Makefile. In this case, maybe we could just bundle all this logic in the Makefile itself ?

robbycochran:
@JoukoVirtanen Please update this PR to point to an updated commit on master.

JoukoVirtanen:
> It's a good idea to use build-collector.sh in the mounted volume; since it avoids some of the "stale builder" image issues when updating this script.
> 
> I am wondering whether something is still using build-collector.sh except the Makefile. In this case, maybe we could just bundle all this logic in the Makefile itself ?

This was actually done in a previous PR. It should not have really appeared in "Files Changed", but did because of the way this branch was created. Openshift CI can't really use the Makefile to build or run images. I don't really see any reason to change the Makefile.

collector/container/Dockerfile
```diff
@@ -3,15 +3,43 @@ ARG BASE_IMAGE=ubi8/ubi-minimal
...
+RUN cp "${ROOT_DIR}"/scripts/* /
+
 WORKDIR /bundle
 RUN ./extract-bundle.sh
 
+RUN echo "${collector_version}" > /COLLECTOR_VERSION
+RUN echo "${module_version}" > /bundle/kernel-modules/MODULE_VERSION.txt
+
+WORKDIR /go/src/github.com/stackrox/collector
+RUN if [[ "${collector_version}" == "xxx" ]]; then \
```
robbycochran, line 27: I would expect this and the following RUN step to occur in `Dockerfile.create_bundle`. Can you move them there?

collector/container/scripts/bootstrap.sh
```diff
@@ -71,6 +71,12 @@ function main() {
     # Export SNI_HOSTNAME and default it to sensor.stackrox
     export SNI_HOSTNAME="${SNI_HOSTNAME:-sensor.stackrox}"
 
+    # Get collector and module versions from files and set corresponding environment variables
+    collector_version="$(cat /COLLECTOR_VERSION)"
```
robbycochran, line 75: I would prefer that we do not create a new dependency in this bootstrap script on COLLECTOR_VERSION AND MODULE_VERSION files.

robbycochran (changes requested):

collector/container/scripts/bootstrap.sh
JoukoVirtanen: That is my preference as well, but I had difficulty setting COLLECTOR_VERSION and MODULE_VERSION and this is what worked. collector/container/Dockerfile is split into two phases and those two environment variables can only be determined in the first phase and need to be passed into the second. There are alternate ways of passing environment variables from one image to the next, but I think in this case where the environment variables are sometimes passed in as an ARG and sometimes set from a command the only way of persisting them that I found to work was saving them to a file.

.openshift-ci/build/Dockerfile.build_root
```diff
@@ -0,0 +1,8 @@
+FROM quay.io/stackrox-io/apollo-ci:collector-0.3.42
+
+# oc
```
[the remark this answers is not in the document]
JoukoVirtanen: Yes, but during development it is much easier to modify things here then to try and do another PR in another repo.

collector/container/Dockerfile
erthalion: Could you please elaborate, why would you expect them there?

.openshift-ci/build/Dockerfile.build_root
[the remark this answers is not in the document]
robbycochran: Ok, sounds good.

collector/container/Dockerfile
```diff
@@ -1,17 +1,45 @@
...
+RUN mkdir /bundle
+
+RUN cp "${ROOT_DIR}/bundle.tar.gz" /
+RUN cp "${ROOT_DIR}/extract-bundle.sh" /bundle/
+RUN cp "${ROOT_DIR}"/scripts/* /
+
 WORKDIR /bundle
 RUN ./extract-bundle.sh
 
+RUN echo "${collector_version}" > /COLLECTOR_VERSION
```
robbycochran, line 30: I think this will always overwrite the `COLLECTOR_VERSION` file in `.openshift-ci/build/Dockerfile.create_bundle`

collector/container/scripts/bootstrap.sh
robbycochran: Your logic makes sense @JoukoVirtanen. However, I do think that having multiple ways of instantiating these values introduces the potential for errors. For example, we won't log or exit here if the files are not found and the env vars are not already defined. 

Perhaps we should use the same technique as [Dockerfile.envsubst](https://github.com/stackrox/stackrox/blob/master/image/rhel/Dockerfile.envsubst) in stackrox and generate the Dockerfile with the correct env vars injected.

collector/container/Dockerfile
robbycochran: I thought it was cleaner and would make the code more straightforward, but I'm not sure how much it helps.

JoukoVirtanen:
There is a problem here since COLLECTOR_VERSION and MODULE_VERSION.txt are created by Dockerfile.create_bundle, but that dockerfile is not used when building collector outside of Openshift CI. That is why the images job failed. I think that is the wrong place to try to create those files. What about trying to create them in builder/build/build-collector.sh?

.openshift-ci/build/Dockerfile.build_root
```diff
@@ -0,0 +1,8 @@
...
+RUN sudo wget --no-verbose -O oc.tgz https://github.com/openshift/origin/releases/download/v3.11.0/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz
+RUN sudo tar -xf oc.tgz
+RUN sudo install openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit/oc /usr/local/bin
+RUN sudo rm -rf openshift-* oc.tgz
+RUN command -v oc
```
erthalion, line 8: Does this image exist only for `oc`? There seems to be a way to inject existing `oc` into the steps [[1]], would it be easier?

[1]: https://docs.ci.openshift.org/docs/architecture/step-registry/#injecting-the-oc-cli

.openshift-ci/build/Dockerfile.drivers
```diff
@@ -0,0 +1,13 @@
...
+
+ENV SRC_ROOT_DIR=/go/src/github.com/stackrox/collector
+WORKDIR $SRC_ROOT_DIR
+
+RUN cp "${SRC_ROOT_DIR}"/.openshift-ci/build/install-gcloud-sdk.sh /
+RUN "${SRC_ROOT_DIR}"/.openshift-ci/build/install-gcloud-sdk.sh /
+RUN ls
+RUN /install-gcloud-sdk.sh
+
+RUN gsutil -m cp -r 'gs://collector-modules/612dd2ee06b660e728292de9393e18c81a88f347ec52a39207c5166b5302b656/1.0.0/collector-3.10.0-1160.62.1.el7.x86_64.ko.gz' kernel-modules/
```
erthalion, line 11: Do I understand correctly, that the images are build only with this version here? What would require to build with everything (in follow up PRs)?

.openshift-ci/build/install-gcloud-sdk.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+dnf update -y
+dnf install epel-release -y
+
+tee -a /etc/yum.repos.d/google-cloud-sdk.repo << EOM
```
erthalion, line 7: Giles have correctly mentioned, that we have quite a few different approaches implemented to install gcloud sdk. Just a note to reconcile them in the future.

erthalion (commented):
Overall looks good, thanks! Few comments below.

erthalion (approved):

PR #740, opened 2022-06-23
Process Arguments integration tests and refactor

Stringy:
## Description

Adds some tests for process args to catch the missing args bug recently encountered. This also refactors some of the tests to verify all parts of the returned data. This relies on some changes to mock gRPC server: https://github.com/stackrox/stackrox/pull/2177

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Tested locally with module version 2.0.0-rc2 (which does not have the proc args bug) and 1.1.0 (which does) and the bug is caught as expected.


integration-tests/integration_test.go
```diff
@@ -260,52 +260,82 @@ func (s *ProcessNetworkTestSuite) TearDownSuite() {
...
+	for _, expected := range expectedLineages {
+		val, err := s.GetLineageInfo(expected.Name, "0", processLineageInfoBucket)
+		s.Require().NoError(err)
+		lineage, err := NewProcessLineage(val)
+		s.Require().NoError(err)
+
+		assert.Equal(expected.Name, lineage.Name)
+		assert.Equal(expected.ExePath, lineage.ExePath)
+		assert.Equal(expected.ParentUid, lineage.ParentUid)
+		assert.Equal(expected.ParentExePath, lineage.ParentExePath)
```
Molter73, line 337: Think we could create a couple of helper functions that take the full maps and compares them internally? I think it would make it even easier to read. Same goes for assertion of `ProcessInfo` structs. (There's probably a way to implement comparison of structs directly, but I'm not that familiar with go 🤷🏻‍♂️ )
```suggestion
                AssertProcessLineage(expected, lineage);
```

integration-tests/process.go
```diff
@@ -0,0 +1,75 @@
...
+	Gid int
+	Pid int
+	Args string
+}
+
+func NewProcessInfo(line string) (*ProcessInfo, error) {
+	parts := strings.SplitN(line, ":", 6)
+
+	if len(parts) != 6 {
+		return nil, fmt.Errorf("not enough parts to gRPC process info string: %s", line)
```
Molter73, line 22: I think I like the error message on `NewProcessLineage` a little more, could we rewrite this one?
```suggestion
		return nil, fmt.Errorf("invalid gRPC string for process info: %s", line)
```

Molter73 (commented):
The refactoring looks awesome! Much easier to read! Thanks for the PR @Stringy! I'm not approving so we don't accidentally merge the wrong mock version ;)

integration-tests/integration_test.go
```diff
@@ -260,52 +260,82 @@ func (s *ProcessNetworkTestSuite) TearDownSuite() {
...
+		val, err := s.Get(expected.Name, processBucket)
+		s.Require().NoError(err)
+		processInfo, err := NewProcessInfo(val)
+		s.Require().NoError(err)
+
+		assert.Equal(expected.Name, processInfo.Name)
+		assert.Equal(expected.ExePath, processInfo.ExePath)
+		assert.Equal(expected.Uid, processInfo.Uid)
+		assert.Equal(expected.Gid, processInfo.Gid)
+		assert.True(processInfo.Pid >= 0)
```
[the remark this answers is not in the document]
Stringy: I'm sure in my head this made sense at the time, but you're right and I'll fix it :)

integration-tests/integration_test.go
Stringy: Looks like `assert.Equal()` might handle this for us, it will eventually do `reflect.DeepEqual()` which may or may not be suitable. I'll try it and see.

Molter73 (approved):

integration-tests/integration_test.go
```diff
@@ -260,52 +261,84 @@ func (s *ProcessNetworkTestSuite) TearDownSuite() {
...
 
-	processName = "sleep"
-	exeFilePath = "/bin/sleep"
-	expectedProcessInfo = fmt.Sprintf("%s:%s:%d:%d", processName, exeFilePath, 0, 0)
-	val, err = s.Get(processName, processBucket)
-	s.Require().NoError(err)
-	assert.Equal(s.T(), expectedProcessInfo, val)
+	sort.Slice(actualProcesses, func(i, j int) bool {
+		return actualProcesses[i].Name < actualProcesses[j].Name
+	})
```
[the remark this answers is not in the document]
Stringy: We could, but I thought to avoid any uncertainty in the testing we could sort them so that it's guaranteed determinism

PR #745, opened 2022-06-29
ROX-17097 Tests to exercise NetworkStatusNotifier at gRPC level.

ovalenti:
## Description

We want to minimize the risk of regression with the ongoing changes on the collector reporting algorithm.
These tests are able to mock the connection with Sensor down to the level of proto gRPC objects.

### DIFF
--- a/collector/lib/CollectorConfig.h
+++ b/collector/lib/CollectorConfig.h
@@ -105,7 +105,7 @@ end
 
   std::shared_ptr<grpc::Channel> grpc_channel;
 
- private:
+ protected:
   bool use_chisel_cache_;
   int scrape_interval_;
   int snap_len_;
--- a/collector/test/NetworkStatusNotifierTest.cpp
+++ b/collector/test/NetworkStatusNotifierTest.cpp
@@ -278,8 +278,6 @@ TEST(NetworkStatusNotifier, UpdateIPnoAfterglow) {
   // the same server connection normalized and grouped in a known subnet
   Connection conn3("containerId", Endpoint(Address(), 1024), Endpoint(IPNet(Address(139, 45, 0, 0), 16), 0), L4Proto::TCP, true);
 
-  UnorderedMap<Address::Family, std::vector<IPNet>> known_networks = {{Address::Family::IPV4, {IPNet(Address(139, 45, 0, 0), 16)}}};
-
   config.DisableAfterglow();
 
   // the connection is always ready
--- a/collector/test/NetworkStatusNotifierTest.cpp
+++ b/collector/test/NetworkStatusNotifierTest.cpp
@@ -25,6 +25,8 @@ You should have received a copy of the GNU General Public License along with thi
 #include <mutex>
 #include <string>
 
+#include <google/protobuf/util/time_util.h>
+
 #include "internalapi/sensor/network_connection_iservice.grpc.pb.h"
 
 #include "CollectorConfig.h"
@@ -43,6 +45,7 @@ using grpc_duplex_impl::Status;
 using ::testing::Invoke;
 using ::testing::Return;
 using ::testing::ReturnPointee;
+using ::testing::UnorderedElementsAre;
 
 // until we use C++20
 class Semaphore {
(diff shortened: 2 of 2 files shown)


collector/test/NetworkStatusNotifierTest.cpp
```diff
@@ -111,6 +126,80 @@ class MockNetworkConnectionInfoServiceComm : public INetworkConnectionInfoServic
...
+          !conn.has_close_timestamp());
+    }
+  }
+
+  const std::unordered_map<Connection, bool, Hasher>& get_updated_connections() const { return updated_connections_; }
+
+ private:
+  std::unordered_map<Connection, bool, Hasher> updated_connections_;
+
+  static IPNet IPNetFromProto(const sensor::NetworkAddress& na) {
```
erthalion, line 152: Are there any intersections with the implementation of `NetworkSignalHandler::GetConnection` here? Is it possible to reuse some parts of it?

collector/test/NetworkStatusNotifierTest.cpp
```diff
@@ -156,14 +251,114 @@ TEST(NetworkStatusNotifier, SimpleStartStop) {
...
+  Semaphore sem(0);  // to wait for the service to accomplish its job.
+
+  // the connection as scrapped (public)
+  Connection conn1("containerId", Endpoint(Address(10, 0, 1, 32), 1024), Endpoint(Address(139, 45, 27, 4), 999), L4Proto::TCP, true);
+  // the same server connection normalized
+  Connection conn2("containerId", Endpoint(Address(), 1024), Endpoint(Address(255, 255, 255, 255), 0), L4Proto::TCP, true);
+  // the same server connection normalized and grouped in a known subnet
+  Connection conn3("containerId", Endpoint(Address(), 1024), Endpoint(IPNet(Address(139, 45, 0, 0), 16), 0), L4Proto::TCP, true);
+
+  UnorderedMap<Address::Family, std::vector<IPNet>> known_networks = {{Address::Family::IPV4, {IPNet(Address(139, 45, 0, 0), 16)}}};
```
erthalion, line 281: Maybe I'm missing something, where is this variable being used?

collector/test/NetworkStatusNotifierTest.cpp
```diff
@@ -111,6 +126,80 @@ class MockNetworkConnectionInfoServiceComm : public INetworkConnectionInfoServic
...
+              L4ProtocolFromProto(conn.protocol()),
+              conn.role() == sensor::ROLE_SERVER),
+          !conn.has_close_timestamp());
+    }
+  }
+
+  const std::unordered_map<Connection, bool, Hasher>& get_updated_connections() const { return updated_connections_; }
+
+ private:
+  std::unordered_map<Connection, bool, Hasher> updated_connections_;
```
erthalion, line 150: Could you pls add few words to clarify the content here? Mostly to mention that the boolean value indicates presence of "closed" timestamp.

erthalion (commented):
Looks good, thanks! Few minor comments below:

collector/lib/CollectorConfig.h
```diff
@@ -105,7 +105,7 @@ end
 
   std::shared_ptr<grpc::Channel> grpc_channel;
 
- private:
+ protected:
```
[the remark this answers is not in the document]
ovalenti: I used this approach because the object seems to be built to be immutable; and I didn't wanted to change that.
I also expect that more members might be subject to change for testing purpose, I making the private section protected was kindof future-proof on this aspect.

Ideally, the configuration object would have getters and setters for all attributes and it would be provided as a const reference in the production code. But I felt that was too big a change. Having remorse now... :)

collector/test/NetworkStatusNotifierTest.cpp
ovalenti: Both are building the same object, which is the internal representation of a Connection. However, they use distinct sources to do so:

- NetworkStatusNotifierTest uses a gRPC object as source.
- NetworkSignalHandler uses a falco probe event

So I agree on the common taste, but I don't see a meaningful way to factorize.

collector/test/NetworkStatusNotifierTest.cpp
ovalenti: You are right, it is not !

I prepared this "known network" variable in order to inject it during the test and re-made it here: https://github.com/stackrox/collector/blob/olivier/ROX-11097-test-pubIP-change-in-NetworkStatusNotifier/collector/test/NetworkStatusNotifierTest.cpp#L327

I will remove this definition.

collector/test/NetworkStatusNotifierTest.cpp
ovalenti: I added a few words. Do you think it makes it clear enough ?

collector/lib/CollectorConfig.h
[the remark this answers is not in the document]
erthalion: Immutability is a good argument, although I wonder if it was implemented like that on purpose.

erthalion (approved):

PR #746, opened 2022-06-29
Rox 11580 collector ci migration push images

JoukoVirtanen:
## Description

Trying to push the images created by Openshift CI to the appropriate repos.

## Checklist

## Testing Performed

See above

### DIFF
--- a/.openshift-ci/build/push-images.sh
+++ b/.openshift-ci/build/push-images.sh
@@ -1,4 +1,0 @@
-#!/usr/bin/env bash
-set -eo pipefail
-
-oc registry login
--- a/.openshift-ci/jobs/push-images/push-images.sh
+++ b/.openshift-ci/jobs/push-images/push-images.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+export PROJECT_DIR=/go/src/github.com/stackrox/collector
+cd "$PROJECT_DIR"
+
+collector_version="$(make tag)"
+export COLLECTOR_VERSION="$collector_version"
+
+export PUBLIC_REPO=quay.io/stackrox-io
+export QUAY_REPO=quay.io/rhacs-eng
+
+shopt -s nullglob
+for cred in /tmp/secret/**/[A-Z]*; do
+    export "$(basename "$cred")"="$(cat "$cred")"
+done
+
+# shellcheck source=SCRIPTDIR/registry_rw_login.sh
+source "${PROJECT_DIR}/.openshift-ci/jobs/push-images/registry_rw_login.sh"
+
+image_repos=(
+    "${QUAY_REPO}"
+    "${PUBLIC_REPO}"
+)
+
+declare -A image_tags
+image_tags["$COLLECTOR_FULL"]="${COLLECTOR_VERSION}-osci"
+#image_tags["$COLLECTOR_BASE"]="${COLLECTOR_VERSION}-base"
+#image_tags["$COLLECTOR_SLIM"]="${COLLECTOR_VERSION}-slim"
+#image_tags["$COLLECTOR_LATEST"]="${COLLECTOR_VERSION}-latest"
+
+oc registry login
+for repo in "${image_repos[@]}"; do
+    registry_rw_login "$repo"
+    for key in "${!image_tags[@]}"; do
+        image="${repo}/collector:${image_tags[$key]}"
+        echo "Pushing image ${image}"
+        osci_image="$key"
+        oc image mirror "$osci_image" "${image}"
+    done
+done
--- a/.openshift-ci/jobs/push-images/registry_rw_login.sh
+++ b/.openshift-ci/jobs/push-images/registry_rw_login.sh
@@ -9,24 +9,25 @@ registry_rw_login() {
     local registry="$1"
 
     case "$registry" in
-        docker.io/stackrox)
-            if [[ -z "$DOCKER_IO_PUSH_USERNAME" ]]; then
-                echo "DOCKER_IO_PUSH_USERNAME is not defined"
-                exit 0 # Change to exit 1 later
-            fi
-            docker login --username "$DOCKER_IO_PUSH_USERNAME" --password-stdin docker.io <<< "$DOCKER_IO_PUSH_PASSWORD"
-            ;;
         quay.io/rhacs-eng)
             if [[ -z "$QUAY_RHACS_ENG_RW_USERNAME" ]]; then
                 echo "QUAY_RHACS_ENG_RW_USERNAME is not defined"
-                exit 0 # Change to exit 1 later
+                exit 1
+            fi
+            if [[ -z "$QUAY_RHACS_ENG_RW_PASSWORD" ]]; then
+                echo "QUAY_RHACS_ENG_RW_PASSWORD is not defined"
+                exit 1
             fi
             docker login --username "$QUAY_RHACS_ENG_RW_USERNAME" --password-stdin quay.io <<< "$QUAY_RHACS_ENG_RW_PASSWORD"
             ;;
         quay.io/stackrox-io)
             if [[ -z "$QUAY_STACKROX_IO_RW_USERNAME" ]]; then
                 echo "QUAY_STACKROX_IO_RW_USERNAME is not defined"
-                exit 0 # Change to exit 1 later
+                exit 1
+            fi
+            if [[ -z "$QUAY_STACKROX_IO_RW_PASSWORD" ]]; then
+                echo "QUAY_STACKROX_IO_RW_PASSWORD is not defined"
+                exit 1
             fi
             docker login --username "$QUAY_STACKROX_IO_RW_USERNAME" --password-stdin quay.io <<< "$QUAY_STACKROX_IO_RW_PASSWORD"
             ;;
--- a/.openshift-ci/jobs/push-images/registry_rw_login.sh
+++ b/.openshift-ci/jobs/push-images/registry_rw_login.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+registry_rw_login() {
+    if [[ "$#" -ne 1 ]]; then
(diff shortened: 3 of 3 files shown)


.circleci/images/70-build-collector-slim-and-base-images.sh
```diff
@@ -23,11 +23,13 @@ fi
     "-" \
     "$SOURCE_ROOT/${COLLECTOR_BUILD_CONTEXT}"
 
+make -C "$SOURCE_ROOT" container-dockerfile
```
robbycochran, line 26: Looks like your PR is pointing to an old version of master.  Please run `smart-rebase` on your branch and force push the changes.

JoukoVirtanen: I thought I had rebased. It is rebased again.

.openshift-ci/build/push-images.sh
```diff
@@ -1,4 +1,51 @@
...
+        die "missing arg. usage: registry_rw_login <registry>"
+    fi
+
+    local registry="$1"
+
+    case "$registry" in
+        docker.io/stackrox)
+            if [[ -z "$DOCKER_IO_PUSH_USERNAME" ]]; then
+                echo "DOCKER_IO_PUSH_USERNAME is not defined"
+                exit 0 # Change to exit 1 later
```
JoukoVirtanen, line 15: exit 0 is used here temporarily, because the push-images job uses the openshift/release master branch, which does not have secrets. As a result the push-images fails in collector but passes in openshift/release without this check. I thought it would be better for a PR review if all of the tests passed.

.openshift-ci/build/push-images.sh
```diff
@@ -1,4 +1,51 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
+registry_rw_login() {
```
Stringy, line 4: Can we put this in a utilities file so it can be used more widely? Same for the secrets loading below.

These are part of https://github.com/stackrox/stackrox/blob/master/scripts/lib.sh so I think we should just copy that over into .openshift-ci/ or similar. I think we already have duplicates of this kind of functionality which will be a pain to maintain in the future

.openshift-ci/build/push-images.sh
```diff
@@ -1,4 +1,51 @@
...
+done
+
 oc registry login
+
+registry_rw_login quay.io/rhacs-eng
+
+cd /go/src/github.com/stackrox/collector
+COLLECTOR_VERSION="$(make tag)"
+
+oc image mirror "$COLLECTOR_FULL" quay.io/rhacs-eng/collector:"$COLLECTOR_VERSION"
```
robbycochran, line 51: Shouldn't we also push to `quay.io/stackrox-io`? This might be a good place to re-use `.circleci/images/100-push-images.sh` WDYT?

.openshift-ci/build/push-images.sh
JoukoVirtanen: Discussed in huddle. Will leave as is.

.openshift-ci/build/push-images.sh
```diff
@@ -1,4 +1,54 @@
...
+done
+
 oc registry login
+
+
+cd /go/src/github.com/stackrox/collector
+COLLECTOR_VERSION="$(make tag)"
+
+registry_rw_login quay.io/rhacs-eng
+oc image mirror "$COLLECTOR_FULL" quay.io/rhacs-eng/collector:"$COLLECTOR_VERSION"
```
robbycochran, line 51: You need to use a different tag so that you don't clash with the circle ci built images.

.openshift-ci/build/push-images.sh
JoukoVirtanen: It might not be possible to use push-as-manifest-list.sh, which is called by .circleci/images/100-push-images.sh. In the stackrox/stackrox version of push-as-manifest-list.sh there is the following line near the top of the script

```
[[ "${OPENSHIFT_CI:-false}" == "false" ]] || { die "Not supported in OpenShift CI"; }
```

The reason is that it might not be possible to create a manifest list with oc image. I asked Gavin about this and he does not seem to think that having manifest lists is important. I would like confirmation on this from other people.

There are other issues with using 100-push-images.sh. In the following
```
oc image mirror "$COLLECTOR_FULL" quay.io/rhacs-eng/collector:"$COLLECTOR_VERSION"
```
COLLECTOR_FULL is an environment variable created by Openshift CI, which is an image created by Openshift CI. If we have multiple collector images, we will have multiple environment variables for them. We need to be able to pair the Openshift CI locations for the images and the destinations we want for them. This would require a fairly major rewrite on 100-push-images.sh.

.openshift-ci/jobs/push-images/push-images.sh
```diff
@@ -0,0 +1,26 @@
...
+source "${PROJECT_DIR}/registry_rw_login.sh"
+
+image_repos=(
+    "${QUAY_REPO}"
+    "${PUBLIC_REPO}"
+)
+
+declare -A image_tags
+image_tags["$COLLECTOR_FULL"]="${COLLECTOR_VERSION}-osci"
+#image_tags["$COLLECTOR_BASE"]="${COLLECTOR_VERSION}-base"
```
JoukoVirtanen, line 13: These lines will be uncommented once the drivers PR is merged.

.openshift-ci/jobs/push-images/push-images.sh
```diff
@@ -0,0 +1,41 @@
...
+#image_tags["$COLLECTOR_SLIM"]="${COLLECTOR_VERSION}-slim"
+#image_tags["$COLLECTOR_LATEST"]="${COLLECTOR_VERSION}-latest"
+
+oc registry login
+for repo in "${image_repos[@]}"; do
+    registry_rw_login "$repo"
+    for key in "${!image_tags[@]}"; do
```
Stringy, line 35: this may be a silly question but what's the `!` for in this line? I tried it locally and it's trying to look up a previous  command

Stringy (approved):
LGTM!

.openshift-ci/jobs/push-images/push-images.sh
```diff
@@ -0,0 +1,41 @@
...
+        image="${repo}/collector:${image_tags[$key]}"
+        echo "Pushing image ${image}"
+        osci_image="$key"
+        oc image mirror "$osci_image" "${image}"
```
Molter73, line 39: Maybe can use `$key` directly? Though `key` is not very descriptive, is it the key to an array of tags?
```suggestion
        oc image mirror "$key" "${image}"
```

.openshift-ci/jobs/push-images/registry_rw_login.sh
```diff
@@ -0,0 +1,37 @@
...
+    local registry="$1"
+
+    case "$registry" in
+        docker.io/stackrox)
+            if [[ -z "$DOCKER_IO_PUSH_USERNAME" ]]; then
+                echo "DOCKER_IO_PUSH_USERNAME is not defined"
+                exit 0 # Change to exit 1 later
+            fi
+            docker login --username "$DOCKER_IO_PUSH_USERNAME" --password-stdin docker.io <<< "$DOCKER_IO_PUSH_PASSWORD"
+            ;;
```
Molter73, line 18: I think we can remove this block, we shouldn't be using dockerhub anywhere in the repo anymore.

.openshift-ci/jobs/push-images/registry_rw_login.sh
```diff
@@ -0,0 +1,37 @@
...
+        quay.io/rhacs-eng)
+            if [[ -z "$QUAY_RHACS_ENG_RW_USERNAME" ]]; then
```
Molter73, line 20: Why are we only checking the username for existence? Can we either check both username and password or place a comment in the script specifying that username existence is enough?

.openshift-ci/jobs/push-images/registry_rw_login.sh
```diff
@@ -0,0 +1,37 @@
...
+                echo "QUAY_RHACS_ENG_RW_USERNAME is not defined"
+                exit 0 # Change to exit 1 later
+            fi
+            docker login --username "$QUAY_RHACS_ENG_RW_USERNAME" --password-stdin quay.io <<< "$QUAY_RHACS_ENG_RW_PASSWORD"
+            ;;
+        quay.io/stackrox-io)
+            if [[ -z "$QUAY_STACKROX_IO_RW_USERNAME" ]]; then
```
Molter73, line 27: Same question as above

Molter73 (approved):
Looking good! Left a few comments but I think they are minor things that shouldn't prevent us from merging the PR.

.openshift-ci/jobs/push-images/push-images.sh
JoukoVirtanen: It gets the keys in the associative array. ! has different meanings in bash in different contexts. Bash does not have great syntax.

.openshift-ci/jobs/push-images/push-images.sh
JoukoVirtanen: The variable names may not be the best. I started off with a Circle CI script where image_tags was an array of image_tags. I changed it to be an associative array where the keys are OSCI images and the values are image tags. I should probably have changed the name of the variable, but I couldn't come up with anything better. I think using osci_image makes it more clear what is going on here.

.openshift-ci/jobs/push-images/registry_rw_login.sh
JoukoVirtanen: Done

.openshift-ci/jobs/push-images/registry_rw_login.sh
JoukoVirtanen: Done

.openshift-ci/jobs/push-images/registry_rw_login.sh
JoukoVirtanen: Done

PR #753, opened 2022-07-13
Update falco

Molter73:
## Description

This PR updates the Falco libraries. Some additional changes are needed in order to fix compilation.

Upstream also changed the schema version, making existing drivers incompatible with userspace, so we also bump our driver version to `2.1.0-rc1`

## Checklist

## Testing Performed

~- [ ] Run on the stackrox main repo.~ The stackrox main repo has no way of running with full images, which would cause tests on it to fail because we changed the drivers version. I will run those tests after I merge these changes.


Molter73:
Local ubi tests fail because the driver is not embedded in the image.

collector/lib/SysdigEventExtractor.h
```diff
@@ -130,6 +128,23 @@ class SysdigEventExtractor {
   FIELD_CSTR(evt_args, "evt.args");
   FIELD_RAW(ppid, "proc.ppid", int64_t);
 
+ public:
+  const uint32_t* get_uid(sinsp_evt* event) {
```
Stringy, line 132: A bit of a nit perhaps, but it might be worth defining another macro for extracting this kind of compound field:

```cpp
#define TINFO_FIELD_TYPE(name, field, type) \
    const type* get_##name(sinsp_evt* event) { \
         if (!event) return nullptr; \
         sinsp_threadinfo* tinfo = event->get_thread_info(true); \
         if (!tinfo) return nullptr; \
         return &tinfo->m_##field; \ 
    }

TINFO_FIELD_RAW(uid, "user.uid", uint32_t);
TINFO_FIELD_RAW(gid, "user.gid", uint32_t);
```

Stringy (approved):
LGTM! just a very minor comment that shouldn't necessarily hold anything up

Stringy (approved):
LGTM!

PR #755, opened 2022-07-13
ROX-9019: pre-flight checks for ebpf permissions

Stringy:
## Description

Adds HostHeuristic for EBPF permissions, to ensure that nothing is stopping us from loading eBPF programs and creating maps. This normally affects RHEL-7 with misconfigured SELinux, but this check will cover all platforms.

If a problem is encountered, kernel module collection is used instead.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)

### DIFF
--- a/collector/lib/HostHeuristics.cpp
+++ b/collector/lib/HostHeuristics.cpp
@@ -20,8 +20,9 @@ You should have received a copy of the GNU General Public License along with thi
 * do not wish to do so, delete this exception statement from your
 * version.
 */
-
 #include "HostHeuristics.h"
+
+#include <linux/bpf.h>
 
 #include "Logging.h"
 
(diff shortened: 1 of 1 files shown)


collector/lib/HostHeuristics.cpp
```diff
@@ -185,13 +186,74 @@ class GardenLinuxHeuristic : public Heuristic {
...
+    union bpf_attr attr = {0};
+
+    // it doesn't matter what type we use here, except for 0 (which is invalid)
+    attr.map_type = BPF_MAP_TYPE_HASH;
+    attr.key_size = sizeof(uint64_t);
+    attr.value_size = sizeof(uint64_t);
+    attr.max_entries = 1;
+    attr.map_flags = 0;
+
+    if ((map_fd = sys_bpf(BPF_MAP_CREATE, &attr, sizeof(attr))) < 0) {
```
ovalenti, line 225: I supposed that closing the fd below will release all the resources ?

collector/lib/HostHeuristics.cpp
```diff
@@ -185,13 +186,74 @@ class GardenLinuxHeuristic : public Heuristic {
...
+    CLOG(DEBUG) << "Attempting to create eBPF map...";
+
+    if (canCreateMap()) {
+      CLOG(DEBUG) << "eBPF map creation successful.";
+      return;
+    }
+
+    CLOG(WARNING) << "Unable to create eBPF map, switching to KernelModule collection";
+    CLOG(WARNING) << "To enable eBPF, check your SELinux settings.";
+    hconfig->SetCollectionMethod("kernel-module");
```
ovalenti, line 210: I could see that we do this in a lot of places, so this is just a general remark: shouldn't we use enumerated type for the collection method, instead of strings ?

ovalenti (approved):
This looks good !

collector/lib/HostHeuristics.cpp
Stringy: yeah, when you close the file descriptor the map is deallocated (unless it is pinned, which this one of course isn't :) )

collector/lib/HostHeuristics.cpp
ovalenti: Yes, I was just thinking loud :)

PR #756, opened 2022-07-16
Rox 11580 collector ci migration push images test

JoukoVirtanen:
Don't review. Just for testing.


.openshift-ci/build/push-images.sh
```diff
@@ -1,4 +1,51 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
+registry_rw_login() {
```
Stringy, line 4: Can we put this in a utilities file so it can be used more widely? Same for the secrets loading below.

These are part of https://github.com/stackrox/stackrox/blob/master/scripts/lib.sh so I think we should just copy that over into `.openshift-ci/` or similar. I think we already have duplicates of this kind of functionality which will be a pain to maintain in the future

PR #759, opened 2022-07-19
ppc64le

sumitd2:
Draft PR, changes to add ppc64le support.


Molter73:
Thanks for the PR @sumitd2! We appreciate all the effort you've put into migrating collector to the ppc64le architecture.

We'll take a closer look into the changes ASAP, but as we discussed previously, we won't be able to merge the PR until we can properly test it into our new CI system.

PR #763, opened 2022-07-25
ROX-11830 Add valijson as a submodule and build it

ovalenti:
## Description

The current falco-lib depends on the third-party valijson library.
By default, this library is compiled from sources fetched from internet (tarball) as a side-effect during the build.

We need to add this dependency as a submodule and pre-install it before falco is built.

This work involves improvements on falco-lib to be able to consume properly the dependency from the system.

## Checklist


Stringy (approved):
LGTM!

PR #774, opened 2022-08-02
Ppc64le multiarch

sumitd2:
Draft PR, ppc64le support with multiarch images


builder/install/50-luajit.sh
```diff
@@ -2,8 +2,14 @@
...
-
-cp COPYRIGHT "${LICENSE_DIR}/LuaJIT-${LUAJIT_VERSION}"
+if [ "$(arch)" = "x86_64" ]; then
+    cd third_party/luajit
+    cp COPYRIGHT "${LICENSE_DIR}/LuaJIT-${LUAJIT_VERSION}"
+else
+    cd third_party
+    git clone https://github.com/openresty/luajit2.git
+    cd luajit2
+    git checkout v"${LUAJIT2_VERSION}"
```
yselkowitz, line 12: luajit2 needs to be imported properly as a submodule.

integration-tests/collector_manager.go
```diff
@@ -60,14 +60,15 @@ func NewCollectorManager(e Executor, name string) *collectorManager {
 	}
 
 	collectorImage := ReadEnvVar("COLLECTOR_IMAGE")
+	grpcServerImage := "quay.io/rhacs-eng/grpc-server:3.70.x-468-g5d733366eb"
```
yselkowitz, line 63: Ditto for this file.

integration-tests/container/benchmark/Dockerfile
```diff
@@ -1,5 +1,13 @@
-FROM phoronix/pts
+FROM --platform=linux/amd64 phoronix/pts as stage-amd64
+FROM --platform=linux/ppc64le quay.io/rhacs-eng/phoronix-pts-ppc64le as stage-ppc64le
```
yselkowitz, line 2: Is this image already available?

integration-tests/container/benchmark/Dockerfile
```diff
@@ -1,5 +1,13 @@
...
 
+# Declare TARGETARCH to make it available
+ARG TARGETARCH
+
+# Select final stage based on TARGETARCH ARG
+FROM stage-${TARGETARCH} as final
```
yselkowitz, line 8: If this is just for CI then this should work there, but tricks such as this will not work for downstream builds.

integration-tests/container/perf/Dockerfile
```diff
@@ -1,5 +1,13 @@
-FROM alpine:3.12 AS base
+FROM --platform=linux/amd64 alpine:3.12 AS stage-amd64
+FROM --platform=linux/ppc64le ppc64le/alpine:3.13 AS stage-ppc64le
```
yselkowitz, line 2: `alpine` is already a multi-arch manifest-listed image, so no need to jump through such hoops in this case.  If for whatever reason you need/want 3.13, probably just best to use it across the board.

integration-tests/container/perf/Dockerfile.bpftrace
```diff
@@ -1,4 +1,13 @@
-FROM quay.io/iovisor/bpftrace:latest
+FROM --platform=linux/amd64 quay.io/iovisor/bpftrace:latest AS stage-amd64
+
+FROM --platform=linux/ppc64le ubuntu:focal AS stage-ppc64le
+RUN apt-get update -y && apt-get install -y bpftrace && bpftrace -V
```
yselkowitz, line 4: Looks like the original image also installs `xz-utils`, but either way, if this is all it is, why not just start with a base image and install bpftrace on all arches?

integration-tests/container/schedule-curls/Dockerfile
```diff
@@ -1,3 +1,10 @@
-FROM pstauffer/curl:latest
+FROM --platform=linux/amd64 pstauffer/curl:latest as stage-amd64
+FROM --platform=linux/ppc64le quay.io/rhacs-eng/pstauffer-curl-ppc64le:latest as stage-ppc64le
```
yselkowitz, line 2: Why not just `docker.io/curlimages/curl` which is both multi-arch and current?

integration-tests/integration_test.go
```diff
@@ -427,7 +427,8 @@ func (s *RepeatedNetworkFlowTestSuite) SetupSuite() {
 	err = s.collector.Launch()
 	s.Require().NoError(err)
 
-	scheduled_curls_image := qaImage("quay.io/rhacs-eng/qa", "collector-schedule-curls");
+	image := "quay.io/rhacs-eng/qa"
```
yselkowitz, line 430: The cumulative changes in this file also end up being just cosmetic; drop them entirely.

kernel-modules/build/Dockerfile
```diff
@@ -18,6 +20,14 @@ RUN apt-get update \
       pkg-config \
  ;
 
+RUN git clone https://github.com/google/googletest.git && \
```
yselkowitz, line 23: Seems like this should also be properly submodule'd.

kernel-modules/build/Dockerfile.fc36
```diff
@@ -16,10 +18,19 @@ RUN dnf update -y || true; \
         kmod \
         wget \
         golang-go \
+        dwarves \
         pkg-config || true ; \
     mkdir -p /output; \
     dnf autoremove -y
 
+RUN git clone https://github.com/google/googletest.git && \
```
yselkowitz, line 26: Ditto.

reload/missing-kernel-objects.sh
```diff
@@ -22,7 +22,6 @@ output_dir="$3"
 mkdir -p "${output_dir}/${collector_version}"
 
 image="quay.io/rhacs-eng/collector:${collector_version}-latest"
-
```
yselkowitz, line 25: Cosmetic change only.

scripts/push-as-manifest-list.sh
```diff
@@ -14,7 +14,12 @@ image="$1"
 [[ -n "$image" ]] || die "No image specified"
 [[ "$image" == *:* ]] || die "Must specify a tagged image reference when using this script"
 
-arch_image="${image}-amd64"
+ARCH=$(arch)
+if [ "$ARCH" = "x86_64" ]; then
+    arch_image="${image}-amd64"
+else
+    arch_image="${image}-$ARCH"
+fi
```
yselkowitz, line 22: Please think long-term here.  ARM also has such a naming disparity (uname: `aarch64`, goarch: `arm64`).

collector/CMakeLists.txt
```diff
@@ -93,6 +93,11 @@ add_test(collector-tests runUnitTests)
 # Falco Wrapper Library
 set(BUILD_DRIVER OFF CACHE BOOL "Build the driver on Linux" FORCE)
 set(USE_BUNDLED_DEPS OFF CACHE BOOL "Enable bundled dependencies instead of using the system ones" FORCE)
+if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64")
+	set(USE_BUNDLED_LUAJIT OFF CACHE BOOL "Disable building of the bundled LuaJIT" FORCE)
+else()
+	set(USE_BUNDLED_LUAJIT ON CACHE BOOL "Enable building of the bundled LuaJIT" FORCE)
+endif()
```
yselkowitz, line 100: Hard to tell from the context; does this change the default on x86?  Where is this variable used?

yselkowitz (changes requested):
A few general points:

- `docker buildx build` cannot be simply followed up by `docker push` or `docker run` etc.; the docker daemon cannot directly access `buildx build` images.  If you're just building the image to see if it will build on multiple arches, then fine, but if you're planning on pushing or running the images, then you need to adapt those steps accordingly as well.
- `docker buildx` and `TARGETPLATFORM` hacks are fine for upstream/CI but not for downstream; it doesn't seem like these dockerfiles could or would be used downstream anyway, but just checking.
- There are a few files noted below which, once squashed, end up with only cosmetic changes; just remove those entirely.
- What about s390x?

ovalenti (commented):
I must admit that the project strongly assumes an x86_64 arch and will benefit from being "enabled" on other platforms. So overall, thanks for the work !

On a strategic level, I think that it would be cleaner to add parameters to the build than to add tests about the architecture as part of the build process.

Using the current approach, for instance, it would become quite complex to target several ppc64 distros in the future, and each arch/distro added will require to do these changes again.

In conclusion, I wish we could keep arch/distro specific stuff only inside the code creating CI/CD pipelines, providing as many parameters as required to handle the specifics of each build (like the base image for the builder, for instance).

.gitmodules
```diff
@@ -50,3 +50,7 @@
 	path = third_party/valijson
 	url = https://github.com/tristanpenman/valijson.git
 	branch = master
+[submodule "third_party/luajit2"]
+	path = third_party/luajit2
+	url = https://github.com/openresty/luajit2.git
+	branch = master
```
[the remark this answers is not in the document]
sumitd2: luajit does not support ppc64le

integration-tests/container/benchmark/Dockerfile
sumitd2: This image will have to be build once for power.

kernel-modules/build/Dockerfile
sumitd2: googletest is already a submodule of this repo. I am not sure what is the proper way to get it into this dockerfile.

kernel-modules/build/Dockerfile.fc36
sumitd2: googletest is already a submodule of this repo. I am not sure what is the proper way to get it into this dockerfile.

builder/install/50-luajit.sh
sumitd2: Done

integration-tests/benchmark_test.go
```diff
@@ -37,6 +37,7 @@ func (b *BenchmarkTestSuiteBase) StartPerfTools() {
 	perf := ReadEnvVar("COLLECTOR_PERF_COMMAND")
 	bpftrace := ReadEnvVar("COLLECTOR_BPFTRACE_COMMAND")
 	bcc := ReadEnvVar("COLLECTOR_BCC_COMMAND")
+	image := "quay.io/rhacs-eng/collector-performance"
```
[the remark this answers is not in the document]
sumitd2: Done

integration-tests/collector_manager.go
sumitd2: Done

integration-tests/container/benchmark/Dockerfile
sumitd2: Not sure why this wont work?

reload/missing-kernel-objects.sh
sumitd2: Done

scripts/push-as-manifest-list.sh
sumitd2: Done

collector/CMakeLists.txt
sumitd2: Done

.gitmodules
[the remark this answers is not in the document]
ovalenti: If luajit2 works for ppc, then maybe we can use it for all archs ?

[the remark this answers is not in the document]
sumitd2: luajit2 is not actively maintained, plus we did not want to modify the current x86 functionality

sumitd2:
> * `docker buildx build` cannot be simply followed up by `docker push` or `docker run` etc.; the docker daemon cannot directly access `buildx build` images.  If you're just building the image to see if it will build on multiple arches, then fine, but if you're planning on pushing or running the images, then you need to adapt those steps accordingly as well.
**We can use neither buildx build --load (doesnt work with multiarch images), nor --push (we cant push after every build). Looks like this PR is a waste.**
> * `docker buildx` and `TARGETPLATFORM` hacks are fine for upstream/CI but not for downstream; it doesn't seem like these dockerfiles could or would be used downstream anyway, but just checking.
**No sure why.**
> * There are a few files noted below which, once squashed, end up with only cosmetic changes; just remove those entirely.
**Done**
> * What about s390x?
**That will be a separate PR from a different team.**

.gitmodules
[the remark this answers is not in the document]
yselkowitz: luajit2 seems to be well maintained?

kernel-modules/build/Dockerfile
yselkowitz: with `WORKDIR` and `COPY` directives

kernel-modules/build/Dockerfile.fc36
yselkowitz: with `WORKDIR` and `COPY` directives

.gitmodules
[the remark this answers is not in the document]
sumitd2: you are right, i confused luajit2 with moonjit, which is another fork. Now using it for both x86 and power.

ovalenti:
> On a strategic level, I think that it would be cleaner to add parameters to the build than to add tests about the architecture as part of the build process.
> ...
> In conclusion, I wish we could keep arch/distro specific stuff only inside the code creating CI/CD pipelines, providing as many parameters as required to handle the specifics of each build (like the base image for the builder, for instance).

@sumitd2 , I would love to hear about your opinion on this subject, as it appears that your are pursuing in the exact same direction as before.

sumitd2:
@ovalenti 
Hi Olivier, apologies for the late reply. I am still trying to comprehend your suggestions. I am responsible for porting the code to power, and don't know much about the code or the functionality it implements. I have asked a Redhat team member (@robbycochran) to analyse your post and post a suitable reply. I am taking the same direction in the meanwhile as we are in a bit of a hurry for the first release, I just thought your suggestions could possibly be implemented in the second one. Thank you.

kernel-modules/build/Dockerfile
sumitd2: Done

kernel-modules/build/Dockerfile.fc36
sumitd2: Done

integration-tests/benchmark_test.go
```diff
@@ -37,6 +37,11 @@ func (b *BenchmarkTestSuiteBase) StartPerfTools() {
...
+
+	if runtime.GOARCH != "amd64" {
+		image = "quay.io/rhacs-eng/" + runtime.GOARCH + "/collector-performance"
```
hbrueckner: What about specifying this image as environment variables as well? This might be beneficial for self-built images or would further abstract arch specifics for now... I would guess that in future, this container image will be built w/ manifest-list / multi-arch support.

integration-tests/benchmark_test.go
```diff
@@ -72,7 +77,12 @@ func (b *BenchmarkTestSuiteBase) StartPerfContainer(name string, image string, a
 }
 
 func (b *BenchmarkTestSuiteBase) RunInitContainer() {
-	init_image := qaImage("quay.io/rhacs-eng/collector-performance", "init")
+	image := "quay.io/rhacs-eng/collector-performance"
+
+	if runtime.GOARCH != "amd64" {
```
hbrueckner: same here .. see above.

utilities/PerformanceTesting/start-secured-cluster.sh
```diff
@@ -9,7 +9,6 @@ else
     collector_image_registry=${2:-quay.io/rhacs-eng/$ARCH}
 fi
 
-
```
hbrueckner: cosmetic

sumitd2: Already fixed. Cant see this in the "Files Changed" tab.

integration-tests/benchmark_test.go
sumitd2: arch specifics have already been removed from here. Cant see this in the "Files Changed" tab.

integration-tests/benchmark_test.go
sumitd2: arch specifics have already been removed from here. Cant see this in the "Files Changed" tab.

PR #789, opened 2022-08-16
Fix collector-probe build

robbycochran:
## Description

- Fix: copy `collector-probe` directory during source archive. (H/T @Stringy for finding this issue)
- Create `/kernel-modules` directory to fix error when`no-cache` label is used. 
- Remove collector versions associated with releases 3.66-3.68 that are [end-of-life](https://access.redhat.com/node/5822721) --also reducing the load on a full rebuild.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests


Stringy (approved):
Looks good! this is certainly ensuring that the correct probe is built

PR #798, opened 2022-08-22
Add slack notifications

erthalion:
## Description

Notify on failed integration tests & unit tests on the main branch.

## Checklist

## Testing Performed

Local testing of the script.

### DIFF
--- a/.openshift-ci/slack/notify-if-needed.sh
+++ b/.openshift-ci/slack/notify-if-needed.sh
@@ -1,19 +1,24 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
+ERROR_MSG=$1
+ERROR_CODE=$2
+
 # shellcheck disable=SC2155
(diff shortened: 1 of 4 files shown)


ovalenti (commented):
I wonder if the ` || (notify)` construct will mask the failed test. Indeed, this clause will likely return 0.
When a test failes, is Make still marking the job as failed with this modification ?

erthalion:
> I wonder if the ` || (notify)` construct will mask the failed test. Indeed, this clause will likely return 0. When a test failes, is Make still marking the job as failed with this modification ?

Right, it will be awkward. I've made the notification wrapper to keep the exit code.

ovalenti (approved):
It is difficult to test this is a realistic environment until it gets actually merged...

The code looks good to me !

erthalion:
> It is difficult to test this is a realistic environment until it gets actually merged...

The testing is indirectly happening here https://github.com/openshift/release/pull/31547

PR #816, opened 2022-09-02
Initial set up for CPaaS support packages

Molter73:
## Description

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


JoukoVirtanen (commented):
I looked at the bucket [here](https://console.cloud.google.com/storage/browser/stackrox-collector-modules-staging/pr-builds/mauro-setup-cpaas-support-packages/1569599354468044800?authuser=1&project=stackrox-collector&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false). It looks like the support packages are there for module version 2.1.0 in the cpaas directory. Should you have support packages for other module versions as well? In addition to the cpaas directory there is another 2.1.0 directory, which has the unzipped drivers. That doesn't seem right. Other than those two things this is looking good.

Molter73:
Hi @JoukoVirtanen, up until last Friday, CPaaS was only building version `2.1.0`, I will rerun the CI to see the newer image with more versions is properly mirrored and processed shortly.

The directory holding the drivers directly is meant to be used as cache for future builds, in case some kernel version stops being maintained we can at least keep delivering the driver that was already built. It is also easier to build a collector version that holds those drivers if we host them like that, because we can directly download them into the image and not worry about unzipping them.

.openshift-ci/drivers/scripts/push-drivers.sh
```diff
@@ -1,6 +1,6 @@
 #! /usr/bin/env bash
 
-set -eo pipefail
+set -exo pipefail
```
erthalion, line 3: Is `-x` needed, maybe a leftover?

.openshift-ci/drivers/scripts/cpaas-support-package.sh
```diff
@@ -0,0 +1,41 @@
...
+/scripts/create-support-packages.sh \
+      /LICENSE \
+      "${target}" \
+      "${METADATA_DIR}" \
+      "${OUTPUT_DIR}"
+
+# Sync support packages
+gsutil -m rsync -r "${OUTPUT_DIR}" "${target}/cpaas"
+
+# Sleep for 120s
```
erthalion, line 34: I've seen this approach couple of times, the purpose here is to wait for syncing to finish, right? Maybe it's a good time to use some more reliable approach, what do you think?

kernel-modules/support-packages/04-create-support-packages.sh
```diff
@@ -40,7 +47,7 @@ for mod_ver_dir in "${MD_DIR}/module-versions"/*; do
 
     package_out_dir="${OUT_DIR}/${mod_ver}"
     mkdir -p "$package_out_dir"
-    if [[ "${mod_ver}" =~ [0-9]+\.[0-9]+\.[0-9]+ ]]; then
+    if [[ "${mod_ver}" =~ [0-9]+\.[0-9]+\.[0-9]+(:?-rc[0-9])? ]]; then
```
erthalion, line 50: That's quite interesting, can you explain to me this shell magic, why the last part is matching more than one digit after `-rc`?

erthalion (commented):
Looks good, thanks! Few minor comments/questions below.

JoukoVirtanen (approved):
LGTM!

.openshift-ci/drivers/scripts/push-drivers.sh
Molter73: It's just foe debug, I'll remove it before merging

kernel-modules/support-packages/04-create-support-packages.sh
Molter73: This is leftover logic to truncate the version of the support packages to 6 characters when they were versioned with sha digests, if it has a semver like versioning, then we use it as is. I think we already dropped support for the last driver versioned with shas, so we can probably drop the check altogether.

.openshift-ci/drivers/scripts/cpaas-support-package.sh
Molter73: I believe this is just waiting for the sync to finish properly, it's been handled this way for some time now and it works reliably, we can look into making this better but in the interest of moving this PR forward I'd prefer to do so on a follow up.

erthalion: Sure, although it would be great to understand if this weird looking part is needed.

erthalion (approved):

PR #817, opened 2022-09-05
Get webhook env directly

erthalion:
## Description

env.sh doesn't exist, pull webhook env directly from the notify script.

## Checklist

## Testing Performed

Testing is possible via CI logs.

### DIFF
--- a/.openshift-ci/slack/notify.sh
+++ b/.openshift-ci/slack/notify.sh
@@ -1,8 +1,10 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
-# shellcheck disable=SC1091
-source .openshift-ci/env.sh
+shopt -s nullglob
+for cred in /tmp/secret/**/[A-Z]*; do
+    export "$(basename "$cred")"="$(cat "$cred")"
+done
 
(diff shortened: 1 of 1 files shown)


.openshift-ci/slack/notify.sh
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
-# shellcheck disable=SC1091
-source .openshift-ci/env.sh
+shopt -s nullglob
+for cred in /tmp/secret/**/[A-Z]*; do
```
Stringy, line 5: not one to fix now, perhaps, but we have this pattern in most (all?) scripts but we also have a function in `.openshift-ci/drivers/scripts/lib.sh` that does it too. At some point soon we really need to consolidate all the commonly used patterns

Stringy (approved):

.openshift-ci/slack/notify.sh
erthalion: Yep, that's what I thought as well (and originally the issue even was that I assumed there *was* a global env.sh). Will address it in a follow-up PR.

PR #819, opened 2022-09-05
Adds all-integration-tests label support

Stringy:
## Description

Adds support for all-integration-tests label. Runs rhel-7, rhel-8 and all ubuntu tests on PRs by default. If the label is provided, all the other platforms are tested. On master, all platforms are tested.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)

### DIFF
--- a/.openshift-ci/jobs/integration-tests/env.sh
+++ b/.openshift-ci/jobs/integration-tests/env.sh
@@ -39,17 +39,17 @@ for cred in /tmp/secret/**/[A-Z]*; do
     export "$(basename "$cred")"="$(cat "$cred")"
 done
 
+if pr_has_label "skip-integration-tests"; then
+    echo "Skipping integration tests for ${VM_CONFIG}"
+    exit 0
+fi
+
 # only run all integration tests on master, or when the all-integration-tests
 # label is added. This is checked in this common env script because it allows
 # us to skip pre- and post- steps as well (which source this file)
-if ! pr_has_label "all-integration-tests" && is_in_PR_context; then
+if is_in_PR_context && ! pr_has_label "all-integration-tests"; then
     if [[ ! "$IMAGE_FAMILY" =~ (rhel-(7|8)|ubuntu-) ]]; then
         echo "Not running integration tests for ${VM_CONFIG}"
         exit 0
     fi
-fi
-
-if pr_has_label "skip-integration-tests"; then
-    echo "Skipping integration tests for ${VM_CONFIG}"
-    exit 0
 fi


.openshift-ci/jobs/integration-tests/env.sh
```diff
@@ -33,3 +38,13 @@ shopt -s nullglob
 for cred in /tmp/secret/**/[A-Z]*; do
     export "$(basename "$cred")"="$(cat "$cred")"
 done
+
+# only run all integration tests on master, or when the all-integration-tests
+# label is added. This is checked in this common env script because it allows
+# us to skip pre- and post- steps as well (which source this file)
+if ! pr_has_label "all-integration-tests" && [[ ! "$(get_branch)" =~ ^master\$ ]]; then
```
Molter73, line 45: Can we not check `"$(get_branch)" != "master"` directly?

Molter73: Also, maybe we could directly use this? https://github.com/stackrox/collector/blob/dcfb1b22cd33157159eead083247603012cb4b7f/.openshift-ci/drivers/scripts/lib.sh#L63

Molter73 (commented):
Overall looks good, left a couple minor comments 🙂

.openshift-ci/jobs/integration-tests/env.sh
Stringy: that seems a lot better, I'll use that!

Stringy:
Also added support for skip-integration-tests since I realised that was super easy to add

.openshift-ci/jobs/integration-tests/env.sh
```diff
@@ -33,3 +38,18 @@ shopt -s nullglob
...
+    if [[ ! "$IMAGE_FAMILY" =~ (rhel-(7|8)|ubuntu-) ]]; then
+        echo "Not running integration tests for ${VM_CONFIG}"
+        exit 0
+    fi
+fi
+
+if pr_has_label "skip-integration-tests"; then
+    echo "Skipping integration tests for ${VM_CONFIG}"
+    exit 0
+fi
```
Molter73, line 55: [ultra-nit] It might make sense to make this the very first check, since it will skip all integration tests disregarding any other conditions.

.openshift-ci/jobs/integration-tests/env.sh
```diff
@@ -33,3 +38,18 @@ shopt -s nullglob
 for cred in /tmp/secret/**/[A-Z]*; do
     export "$(basename "$cred")"="$(cat "$cred")"
 done
+
+# only run all integration tests on master, or when the all-integration-tests
+# label is added. This is checked in this common env script because it allows
+# us to skip pre- and post- steps as well (which source this file)
+if ! pr_has_label "all-integration-tests" && is_in_PR_context; then
```
Molter73, line 45: [ultra-nit] Maybe we should check if we are in a PR before attempting to gather the labels?
```suggestion
if is_in_PR_context && ! pr_has_label "all-integration-tests"; then
```

Molter73 (approved):
/lgtm

Left 2 more nits, but otherwise this looks great!

PR #822, opened 2022-09-06
Consolidates common CI logic

Stringy:
## Description

Tries to tidy up our CI scripts and consolidate some of the functionality into a single place

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


.openshift-ci/scripts/lib.sh
```diff
@@ -140,3 +140,46 @@ get_pr_details() {
     _PR_DETAILS="$pr_details"
     echo "$pr_details"
 }
+
+import_creds() {
```
erthalion, line 144: There are few examples of using mounted secrets directly, does it make sense to generalize them as well, what do you think?

Stringy: seems reasonable! I'll add in a helper or two

.openshift-ci/scripts/lib.sh
```diff
@@ -140,3 +140,91 @@ get_pr_details() {
...
+    for cred in /tmp/secret/**/[A-Z]*; do
+        export "$(basename "$cred")"="$(cat "$cred")"
+    done
+}
+
+ensure_secret_exists() {
+    local secret_name="$1"
+
+    if [[ -z "${secret_name}" ]]; then
+        import_creds
```
erthalion, line 155: Does it mean as a side effect everything will be already exported? If so, probably makes sense to mention in the commentary, in case if in the future this function will be used in some other contexts, what do you think?

Molter73: This is fine as is right now, but maybe we want an alternative to `import_creds` that imports just the secret that we are asking for? I think it could make this a function a little easier to follow, at least it took me a second to realize that `-z "${secret_name}"` is a different check in 157 and 159, (also, lazzy initialization and all that jazz).

.openshift-ci/scripts/lib.sh
```diff
@@ -140,3 +140,94 @@ get_pr_details() {
...
+
+    echo "${secret_name}" > "${destination}"
+    chmod "${permissions}" "${destination}"
+}
+
+get_secret_content() {
+    local secret_name="$1"
+
+    ensure_secret_exists "$secret_name"
+    echo "${secret_name}"
```
Molter73, line 183: Wait... How does this work? Lets assume we call `get_secret_content SOME_CREDENTIAL`, then line 182 would expand `secret_name` and call `ensure_secret_exists "SOME_CREDENTIAL"` the you return from the call and... `secret_name` is now magically the secret itself rather than the key? What?

.openshift-ci/scripts/gcloud-init.sh
```diff
@@ -1,15 +1,19 @@
...
 which gsutil
 gsutil ver
 gsutil ver -l | grep crcmod
 gcloud version
 cat ~/.boto 2> /dev/null || true
 echo '[Credentials]' > ~/.boto
-echo 'gs_service_key_file = /tmp/secret/stackrox-collector-e2e-tests/GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT' >> ~/.boto
-gcloud auth activate-service-account --key-file /tmp/secret/stackrox-collector-e2e-tests/GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT
+echo "gs_service_key_file = $(get_secret_file GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT)" >> ~/.boto
+gcloud auth activate-service-account --key-file "$(get_secret_file GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT)"
```
Molter73, line 16: [nitpick]
```suggestion
KEY_FILE="$(get_secret_file GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT)"
echo "gs_service_key_file = ${KEY_FILE}" >> ~/.boto
gcloud auth activate-service-account --key-file "${KEY_FILE}"
```

Molter73 (commented):
Overall looks great, thanks for the effort in simplifying our CI scripts @Stringy! Left a few comments, (one of which is probably just me being confused by bash)

.openshift-ci/scripts/lib.sh
Stringy: good spot... not sure what I was thinking here

.openshift-ci/build/install-gcloud-sdk.sh
```diff
@@ -1,20 +0,0 @@
-#!/usr/bin/env bash
```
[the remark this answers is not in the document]
Stringy: it's still used in the drivers docker files I was just consolidating two different implementations of the install script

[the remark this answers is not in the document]
Molter73: I must be blind, where is the consolidated script?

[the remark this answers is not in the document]
Stringy: it's `.openshift-ci/scripts/gcloud-sdk-install.sh` which existed before this PR - IIRC I kept that version over the one in `.openshift-ci/driver/scripts` because it was slightly more robust

PR #823, opened 2022-09-09
ROX-11098 Test fixture that includes the current "known networks" list

ovalenti:
## Description

The current public IP list includes ~28K IP ranges, and we want to check that this size is not a concern.

## Testing Performed

The PR only adds a test


collector/test/NetworkStatusNotifierTest.cpp
```diff
@@ -361,6 +365,165 @@ TEST(NetworkStatusNotifier, UpdateIPnoAfterglow) {
   net_status_notifier->Stop();
 }
 
+namespace network_list {
+
+/* Created with:
```
robbycochran, line 370: Thanks for including this note.

PR #824, opened 2022-09-09
[do-not-merge] Testing changes for s390x to eBPF probe

Molter73:
## Description

This PR is simply meant to run our CI on the latest changes required for s390x to use eBPF. 

Even if the changes are meant for a different arch, they affect x86 and running CI is the easiest way we have to validate changes.


hbrueckner:
Thanks @Molter73 !

PR #826, opened 2022-09-16
Refactor collector image build

Molter73:
## Description

This refactor makes building of the collector image more streamlined.
This is achieved by removing the intermediate bundle and directly
copying the needed files into the image itself.

Some other minor changes have also been made across the repository:
- Created a single make target for running linters on CI.
- Created a dedicated CI image that replaces the old root/src images in
  OSCI.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Rehearsed on openshift/release#32351


.openshift-ci/jobs/apollo-src/dump-pr-data.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+mkdir /ci-data/
+touch /ci-data/dump.sh
+
+if [[ -n "${REPO_OWNER}" ]]; then
+    echo "export REPO_OWNER='${REPO_OWNER}'" >> /ci-data/dump.sh
+fi
```
Stringy, line 9: [ultranit] can we have newlines between the if statements?

utilities/lint.sh
```diff
@@ -0,0 +1,19 @@
...
+fi
+if ! make -C "${ROOT_DIR}" shfmt-check; then
+    ((errors = errors + 1))
+fi
+if ! make -C "${ROOT_DIR}" shellcheck-all; then
+    ((errors = errors + 1))
+fi
+
+if ((errors)); then
+    echo >&2 "${errors} linters have failed"
```
Stringy, line 18: can we also have some logging that describes which linters failed?

Makefile
```diff
@@ -196,3 +186,7 @@ $(eval $(call linter,clang-format))
 $(eval $(call linter,shellcheck))
 $(eval $(call linter,shfmt))
 $(eval $(call linter,circleci_validate))
+
+.PHONY: linters
+linters:
```
Stringy, line 191: could we use `pre-commit` to do this? i.e. `pre-commit run --all-files`

.openshift-ci/drivers/pre-build.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM pipeline:src
+FROM quay.io/stackrox-io/apollo-ci:collector-0.3.44-1-gb00ffc52af
```
Stringy, line 1: Should this use `pipeline:apollo-ci`?

.openshift-ci/jobs/apollo-src/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM replaced-by-osci:bin AS pr-data
+
+RUN .openshift-ci/jobs/apollo-src/dump-pr-data.sh
+
+FROM quay.io/stackrox-io/apollo-ci:collector-0.3.44-1-gb00ffc52af
```
Stringy, line 5: As above - should this use `pipeline:apollo-ci`?

Stringy (commented):
Looks really good - just a couple of relatively minor comments / questions

.openshift-ci/jobs/integration-tests/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM pipeline:src
+FROM quay.io/stackrox-io/apollo-ci:collector-0.3.44-1-gb00ffc52af
```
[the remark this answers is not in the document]
Molter73: It could, but after working in this PR I've come to the conclusion we made a terrible mistake by using things like `pipeline` and `replaced-by-osci` in our dockerfiles, the reason being that OSCI can properly replace them when running there without any problem and if we use ACTUAL, PROPER images in the dockerfiles we can run the same builds locally with relative ease.

utilities/lint.sh
Molter73: I mean, we could, but the output will already have the linters output above this line, I'd prefer to not complicate the logic in this script to also be appending names onto a variable to be print, WDYT?

Makefile
Molter73: We would need to install `pre-commit` in the apollo-ci image, can we do it in a followup?

.openshift-ci/drivers/pre-build.Dockerfile
Molter73: See https://github.com/stackrox/collector/pull/826#discussion_r983375354

.openshift-ci/jobs/apollo-src/Dockerfile
Molter73: See https://github.com/stackrox/collector/pull/826#discussion_r983375354

.openshift-ci/jobs/integration-tests/Dockerfile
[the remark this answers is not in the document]
Molter73: I'll replace the apollo-ci image, the images were drivers are copied from are currently not built locally and will require some extra work to get them working, so we can probably leave it for a follow up, does that make sense to you?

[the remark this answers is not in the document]
Stringy: Yep sounds reasonable to me!

Makefile
Stringy: sure - I might find some time this week to do so

Stringy (approved):

PR #831, opened 2022-09-20
Add driver build flow as plantuml

Molter73:
## Description

Graphs are uploaded as svg since github does not support rendering plantuml as part of markdown files.

For ease of edit in the future, the raw plantuml is embedded in the corresponding markdown file and hidden in a 'details' block.

## Testing Performed

The change is purely for documentation.


Stringy (approved):
LGTM!

PR #836, opened 2022-09-26
Integration tests retry

erthalion:
## Description

Add a simple retry function, apply it to the command executor and container launcher. The implementation has few quirks:

* It doesn't use routines, but in case of integration tests it doesn't matter if we would block.

* It has two return arguments with fixed type: string and error. The latter one is required for retry logic, the former seem to be rather limited choice. The reason is that without returning any data except an error the scope of the payload function would have to be much bigger (as nothing has to be returned). At the moment it looks like we mostly return strings anyway, so it feels like a good compromise between full-fledged solution and a quick hack.

## Checklist

**Automated testing**
  - [ ] Added unit tests

## Testing Performed

TODO

### DIFF
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -528,9 +528,10 @@ func (s *RepeatedNetworkFlowTestSuite) TestRepeatedNetworkFlow() {
 func (s *IntegrationTestSuiteBase) launchContainer(args ...string) (string, error) {
 	cmd := []string{"docker", "run", "-d", "--name"}
 	cmd = append(cmd, args...)
-	output, err := s.executor.Exec(cmd...)
-	outLines := strings.Split(output, "\n")
-	return outLines[len(outLines)-1], err
+
+	return retry(func() (string, error) {
+		return s.executor.Exec(cmd...)
+	})
 }
 
 func (s *IntegrationTestSuiteBase) waitForContainerToExit(containerName, containerID string, tickSeconds time.Duration) (bool, error) {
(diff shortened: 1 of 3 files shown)


integration-tests/integration_test.go
```diff
@@ -528,9 +528,10 @@ func (s *RepeatedNetworkFlowTestSuite) TestRepeatedNetworkFlow() {
 func (s *IntegrationTestSuiteBase) launchContainer(args ...string) (string, error) {
 	cmd := []string{"docker", "run", "-d", "--name"}
 	cmd = append(cmd, args...)
-	output, err := s.executor.Exec(cmd...)
-	outLines := strings.Split(output, "\n")
-	return outLines[len(outLines)-1], err
```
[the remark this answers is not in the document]
erthalion: Good point, I've missed that.

integration-tests/retry.go
```diff
@@ -0,0 +1,26 @@
...
+
+// Simple retry in loop for commands produce only one string output and error
+func retry(f retryable) (output string, err error) {
+	for i := 0; i < max_retries; i++ {
+		output, err = f()
+		if err == nil {
+			return output, nil
+		} else {
+			time.Sleep(retry_wait_time)
+		}
```
Molter73, line 22: [nitpick] The last loop will wait for 3 seconds before finally exiting, we could add a small if condition here, (though I'm sure you can come up with a nicer and cleaner way to do this).
```suggestion
		} else if i != max_retries - 1 {
			time.Sleep(retry_wait_time)
		}
```

Molter73 (approved):
/lgtm

integration-tests/retry.go
erthalion: Looks fine for me.

PR #840, opened 2022-09-26
Normalize secrets mount to /tmp/secret

Stringy:
## Description

Normalizes secrets mount and ensures consistency between helper functions. Now normalised to /tmp/secret

## Checklist

**Automated testing**
  ~- [ ] Added unit tests~
 ~ - [ ] Added regression tests~

### DIFF
--- a/.openshift-ci/jobs/integration-tests/env.sh
+++ b/.openshift-ci/jobs/integration-tests/env.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 CI_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")"/../.. && pwd)"
-# shellcheck source=SCRIPTDIR=../../scripts/lib.sh
+# shellcheck source=SCRIPTDIR/../../scripts/lib.sh
 source "${CI_ROOT}/scripts/lib.sh"
 
 # to ensure that the locale is consistent between GCP VMs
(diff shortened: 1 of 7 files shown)


.openshift-ci/jobs/integration-tests/env.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 CI_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")"/../.. && pwd)"
-# shellcheck source=SCRIPTDIR=../../scripts/lib.sh
+# shellcheck source=SCRIPTDIR/../../scripts/lib.sh
```
JoukoVirtanen, line 4: Why is the = removed here.

Stringy: it was invalid for the spellcheck directive, because it's expecting a path. `SCRIPTDIR` is a constant that shellcheck treats as the directory containing the script it's currently processing, so any source directives need to be relative to that directory

JoukoVirtanen: I didn't look closely. I did not notice that there were two equal signs or that this was attempting to set SCRIPTDIR equal to something.

JoukoVirtanen (approved):
LGTM!

PR #844, opened 2022-09-27
Jv rox 12845 integration tests for connscraper

JoukoVirtanen:
## Description

We need integration tests for ConnScraper as we are about to modify it for the "Processes listening on ports" MVP. ConnScraper is able to find open endpoints when Collector starts up (and thus knows about endpoints that were opened before Collector started up), whereas NetworkSignalHandler only knows about endpoints that open after Collector starts up. To test ConnScraper we start an nginx container before launching Collector, and test that we can find the nginx endpoint when scraping is on and that we cannot find the nginx endpoint when scraping is off.

Related to this PR is a change for the mock grpc server. See https://github.com/stackrox/stackrox/pull/3215

## Checklist
- [x] Performed integration tests locally

**Automated testing**
  - [x] Added integration tests

## Testing Performed

Go into the stackrox repo root directory and run the following commands

Checkout the branch of the stackrox PR listed above
make mock-grpc-server-build
cp bin/linux_amd64/mock-grpc-server integration-tests/mock-grpc-server/image/bin/ // You may need to adjust the path at linux_amd64
make mock-grpc-server-image

Go int the collector repo root directory
make
make integration-tests-connscraper

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,6 +139,7 @@ ci-integration-tests: integration-tests-repeat-network \
 					  integration-tests-process-network \
 					  integration-tests-missing-proc-scrape \
 					  integration-tests-image-label-json \
+					  integations-tests-connscraper \
 					  integration-tests-report
 
 .PHONY: ci-benchmarks
--- a/Makefile
+++ b/Makefile
@@ -125,6 +125,11 @@ integration-tests-image-label-json:
 	make -C integration-tests image-label-json ||\
 		( .openshift-ci/slack/notify-if-needed.sh "image-label-json" $$? )
 
+.PHONY: integration-tests-connscraper
+integration-tests-connscraper:
+	make -C integration-tests connscraper ||\
+		( .openshift-ci/slack/notify-if-needed.sh "connscraper" $$? )
+
 .PHONY: integration-tests-report
 integration-tests-report:
 	make -C integration-tests report
--- a/Makefile
+++ b/Makefile
@@ -139,7 +139,7 @@ ci-integration-tests: integration-tests-repeat-network \
 					  integration-tests-process-network \
 					  integration-tests-missing-proc-scrape \
 					  integration-tests-image-label-json \
-					  integations-tests-connscraper \
+					  integration-tests-connscraper \
 					  integration-tests-report
 
 .PHONY: ci-benchmarks
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -102,6 +102,26 @@ func TestRepeatedNetworkFlowThreeCurlsNoAfterglow(t *testing.T) {
 	suite.Run(t, repeatedNetworkFlowTestSuite)
 }
 
+// There is one test in which scraping is turned on and we expect to see
+// endpoints opened before collector is turned on. There is another test
+// in which scraping is turned off and we expect that we will not see
+// endpoint opened before collector is turned on.
+func TestConnScraper(t *testing.T) {
+	connScraperTestSuite := &ConnScraperTestSuite{
+		turnOffScrape:	false,
+		expectedResult:	true,
+	}
+	suite.Run(t, connScraperTestSuite)
+}
+
+func TestConnScraperNoScrape(t *testing.T) {
+	connScraperTestSuite := &ConnScraperTestSuite{
+		turnOffScrape:	true,
+		expectedResult:	false,
+	}
+	suite.Run(t, connScraperTestSuite)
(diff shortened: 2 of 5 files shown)


integration-tests/integration_test.go
```diff
@@ -525,6 +553,75 @@ func (s *RepeatedNetworkFlowTestSuite) TestRepeatedNetworkFlow() {
...
+}
+
+func (s *ConnScraperTestSuite) TestConnScraper() {
+	found := false
+	pattern := "^EndpointInfo:.*" + s.serverContainer
+	for _, logLine := range s.logLines {
+
+		match, _ := regexp.MatchString(pattern, logLine)
+		if match {
+			found = true
```
Stringy, line 619: might as well break here, rather than looping through the rest of the log lines

integration-tests/integration_test.go
```diff
@@ -525,6 +553,75 @@ func (s *RepeatedNetworkFlowTestSuite) TestRepeatedNetworkFlow() {
...
+	err := s.collector.Setup()
+	s.Require().NoError(err)
+
+	err = s.collector.Launch()
+	s.Require().NoError(err)
+
+	time.Sleep(10 * time.Second)
+
+
+	s.logLines = s.GetLogLines("grpc-server")
```
Stringy, line 582: Can't we get the grpc-server to store the necessary information in the BoltDB rather than looking in the logs?

Makefile
```diff
@@ -125,6 +125,11 @@ integration-tests-image-label-json:
 	make -C integration-tests image-label-json ||\
 		( .openshift-ci/slack/notify-if-needed.sh "image-label-json" $$? )
 
+.PHONY: integration-tests-connscraper
+integration-tests-connscraper:
+	make -C integration-tests connscraper ||\
+		( .openshift-ci/slack/notify-if-needed.sh "connscraper" $$? )
```
Stringy, line 131: Obviously this matches the other targets in this file but I think we should really avoid having CI-specific logic in the rest of the codebase - one of the most challenging things about the CI port was how coupled everything is

integration-tests/integration_test.go
JoukoVirtanen: I am not sure. I tried doing that for the afterglow integration tests, but apparently BoltDB only stored one of the networking events. I know almost nothing about BoltDB, so maybe this is something that is actually easy, but looking at the grpc-server logs seemed like the best way of doing this.

Makefile
JoukoVirtanen: I think this issue will be left to a different PR.

integration-tests/integration_test.go
```diff
@@ -525,6 +553,76 @@ func (s *RepeatedNetworkFlowTestSuite) TestRepeatedNetworkFlow() {
...
+}
+
+func (s *ConnScraperTestSuite) TearDownSuite() {
+	s.cleanupContainer([]string{"nginx", "collector"})
+	stats := s.GetContainerStats()
+	s.PrintContainerStats(stats)
+	s.WritePerfResults("ConnScraper", stats, s.metrics)
+}
+
+func (s *ConnScraperTestSuite) TestConnScraper() {
```
robbycochran, line 612: This test only checks for the container id. Can you also validate all of the other fields?

integration-tests/integration_test.go
JoukoVirtanen: As I have switched to BoltDB this is no longer applicable.

integration-tests/integration_test.go
JoukoVirtanen: Done. Though I had difficulty with the network address and ended up leaving that out.

integration-tests/integration_test.go
```diff
@@ -685,6 +775,41 @@ func (s *IntegrationTestSuiteBase) GetProcesses(containerID string) ([]ProcessIn
 	return processes, nil
 }
 
+func (s *IntegrationTestSuiteBase) GetEndpoints(containerID string) ([]EndpointInfo, error) {
+	if s.db == nil {
+		return nil, fmt.Errorf("Db %v is nil", s.db)
+	}
+
+	endpoints := make([]EndpointInfo, 0)
```
Stringy, line 783: Where is `EndpointInfo` defined? I think it might be missing from this PR

JoukoVirtanen: Oops. I forgot to add it. It is there now.

Stringy (approved):
LGTM!

PR #845, opened 2022-09-28
Add s390x support--allow socketcall and inserting kernel fix

kcrane:
## Description

On s390x we are relegated to socketcall wrapping various system calls until we have glibc support--[details](https://sourceware.org/pipermail/libc-alpha/2022-September/142108.html).  We will not filter syscall type so we can receive these syscalls.

Also included patch to read kernel module into memory and insert from there instead of finit_module.

## Testing Performed

Ran unit and integration tests on s390x and x86

### DIFF
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -111,6 +111,7 @@ int InsertModule(int fd, const std::unordered_map<std::string, std::string>& arg
   fstat(fd, &st);
   size_t image_size = st.st_size;
   void* image = malloc(image_size);
+  lseek(fd, 0, SEEK_SET);
   read(fd, image, image_size);
   int res = init_module(image, image_size, args_str.c_str());
   free(image);
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -61,7 +61,7 @@ extern "C" {
 #include "SysdigService.h"
 #include "Utility.h"
 
-#define finit_module(fd, opts, flags) syscall(__NR_finit_module, fd, opts, flags)
+#define init_module(module_image, len, param_values) syscall(__NR_init_module, module_image, len, param_values)
 #define delete_module(name, flags) syscall(__NR_delete_module, name, flags)
 
 extern unsigned char g_bpf_drop_syscalls[];  // defined in libscap
@@ -107,9 +107,14 @@ int InsertModule(int fd, const std::unordered_map<std::string, std::string>& arg
     args_str += entry.first + "=" + entry.second;
   }
   CLOG(DEBUG) << "Kernel module arguments: " << args_str;
-  int res = finit_module(fd, args_str.c_str(), 0);
-  if (res != 0) return res;
   struct stat st;
+  fstat(fd, &st);
+  size_t image_size = st.st_size;
+  void *image = malloc(image_size);
+  read(fd, image, image_size);
+  int res = init_module(image, image_size, args_str.c_str());
+  free(image);
+  if (res != 0) return res;
   std::string param_dir = GetHostPath(std::string("/sys/module/") + SysdigService::kModuleName + "/parameters/");
   res = stat(param_dir.c_str(), &st);
   if (res != 0) {
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -110,7 +110,7 @@ int InsertModule(int fd, const std::unordered_map<std::string, std::string>& arg
   struct stat st;
   fstat(fd, &st);
   size_t image_size = st.st_size;
-  void *image = malloc(image_size);
+  void* image = malloc(image_size);
   read(fd, image, image_size);
   int res = init_module(image, image_size, args_str.c_str());
   free(image);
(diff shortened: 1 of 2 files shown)


Molter73:
Well this is weird, 2 of the failing tests are showing this error:
```
[F 20220928 212645 collector.cpp:181] Error inserting kernel module: /module/collector.ko: Exec format error. Aborting...
```
But then the same kernel module is inserted normally on other subset of tests 🤔

kcrane:
@Molter73 O_CLOEXEC didn't help with Permission denied.

Perhaps Exec Format Error is caused by not reading from beginning.  Added lseek to test this out.

The modutils.c will fallback to init_module if finIt_module fails.  Could possibly do same in collector.

Molter73:
> The modutils.c will fallback to init_module if finIt_module fails. Could possibly do same in collector.

That sounds great, but I would not consider it necessary for this PR.

collector/collector.cpp
```diff
@@ -107,9 +107,15 @@ int InsertModule(int fd, const std::unordered_map<std::string, std::string>& arg
     args_str += entry.first + "=" + entry.second;
   }
   CLOG(DEBUG) << "Kernel module arguments: " << args_str;
-  int res = finit_module(fd, args_str.c_str(), 0);
-  if (res != 0) return res;
   struct stat st;
+  fstat(fd, &st);
```
Molter73, line 111: We should probably check the return code for `fstat`, just to be extra sure we are not trying to push bogus information into the kernel.

collector/collector.cpp
```diff
@@ -107,9 +107,15 @@ int InsertModule(int fd, const std::unordered_map<std::string, std::string>& arg
...
+  size_t image_size = st.st_size;
+  void* image = malloc(image_size);
```
Molter73, line 113: We should check that malloc doesn't return a null pointer.

collector/collector.cpp
```diff
@@ -107,9 +107,15 @@ int InsertModule(int fd, const std::unordered_map<std::string, std::string>& arg
...
+  lseek(fd, 0, SEEK_SET);
+  read(fd, image, image_size);
```
Molter73, line 115: And while we are at it, It'd also be nice to check if the `read` bytes are equal to the expected size and no error occurred here.

Molter73: We could also check the return code for `lseek` but if that failed, `read` will probably fail itself or read less bytes than expected.

Molter73 (changes requested):
Thanks for this change @kcrane, I've left a few comments on return values that are not being checked and probably should. Other than those, it looks good!

Molter73 (approved):
/lgtm

collector/collector.cpp
```diff
@@ -107,9 +107,14 @@ int InsertModule(int fd, const std::unordered_map<std::string, std::string>& arg
     args_str += entry.first + "=" + entry.second;
   }
   CLOG(DEBUG) << "Kernel module arguments: " << args_str;
-  int res = finit_module(fd, args_str.c_str(), 0);
-  if (res != 0) return res;
   struct stat st;
+  fstat(fd, &st);
+  size_t image_size = st.st_size;
+  void *image = malloc(image_size);
+  read(fd, image, image_size);
```
hbrueckner: We probably should check the return code, regarding `EINTR` to re-start reading like in:
````
static int __read(int fd, void *buf, int buflen)
{
        unsigned char *p = buf;
        int n, i = 0;

        while (i < buflen) {
                n = read(fd, p+i, buflen-i);
                if (n < 0) {
                        if (errno == EINTR)
                                continue;
                        else
                                return n;
                } else if (n == 0) {
                        return i;
                } else {
                        i += n;
                }
        }

        return i;
}
````

File: collector/collector.cpp [hunk 5]
```diff
@@ -108,12 +108,27 @@ int InsertModule(int fd, const std::unordered_map<std::string, std::string>& arg
...
+  if (!image) {
+      CLOG(ERROR) << "Could not allocate memory for kernel module: " << StrError();
+      errno = EINVAL;
+      return -1;
+  }
   lseek(fd, 0, SEEK_SET);
-  read(fd, image, image_size);
-  int res = init_module(image, image_size, args_str.c_str());
+  size_t read_image_size = read(fd, image, image_size);
+  if (read_image_size != image_size) {
```
hbrueckner, line 126: sorry.. just started reading from the beginning... see comment on first commit to handle `EINTR` and reading chunks.

collector/collector.cpp
kcrane: thanks, added code for partial reads.

collector/collector.cpp
```diff
@@ -96,6 +96,25 @@ static void AbortHandler(int signum) {
   raise(signum);
 }
 
+static int __read(int fd, void* buf, int buflen) {
```
Molter73, line 99: Use of `_` as the first character in a global function is reserved by the compiler and could potentially lead to a conflict somewhere down the line. Could we rename it to something like `read_module` instead?

@stackrox/collector-team, we might want to consider moving module loading and unloading to a separate file in a follow up, it's starting to take up quite a bit of room here.

hbrueckner (approved):
/lgtm

Issue #838, opened 2022-09-26
Segmentation Fault on all nodes in OpenShift 4.9.33


balpert89:
Hi,

Disclaimer: I have opened the same issue at https://github.com/stackrox/stackrox/issues/3195 because I am not sure on which repository this should be tracked as here we have a area/collector label. Please close the one which is at the wrong location.

we are experiencing crashes in collector containers across all nodes in one of our OpenShift clusters.

Debug Log:
```
Collector Version: 3.9.0
OS: Red Hat Enterprise Linux CoreOS 49.84.202205050701-0 (Ootpa)
Kernel Version: 4.18.0-305.45.1.el8_4.x86_64
Starting StackRox Collector...
[I 20220926 112218 HostInfo.cpp:126] Hostname: '<redacted>'
[I 20220926 112218 CollectorConfig.cpp:119] User configured logLevel=debug
[I 20220926 112218 CollectorConfig.cpp:149] User configured collection-method=kernel_module
[I 20220926 112218 CollectorConfig.cpp:206] Afterglow is enabled
[D 20220926 112218 HostInfo.cpp:200] EFI directory exist, UEFI boot mode
[D 20220926 112218 HostInfo.h:100] identified kernel release: '4.18.0-305.45.1.el8_4.x86_64'
[D 20220926 112218 HostInfo.h:101] identified kernel version: '#1 SMP Wed Apr 6 13:48:37 EDT 2022'
[D 20220926 112218 HostInfo.cpp:297] SecureBoot status is 2
[D 20220926 112218 collector.cpp:254] Core dump not enabled
[I 20220926 112218 collector.cpp:302] Module version: 2.0.1
[I 20220926 112218 collector.cpp:329] Attempting to download kernel module - Candidate kernel versions:
[I 20220926 112218 collector.cpp:331] 4.18.0-305.45.1.el8_4.x86_64
[D 20220926 112218 GetKernelObject.cpp:148] Checking for existence of /kernel-modules/collector-4.18.0-305.45.1.el8_4.x86_64.ko.gz and /kernel-modules/collector-4.18.0-305.45.1.el8_4.x86_64.ko
[D 20220926 112218 GetKernelObject.cpp:151] Found existing compressed kernel object.
[I 20220926 112218 collector.cpp:262]
[I 20220926 112218 collector.cpp:263] This product uses kernel module and ebpf subcomponents licensed under the GNU
[...]

PR #846, opened 2022-09-28
Stress test for ProtoAllocator

ovalenti:
## Description

This work is intended to stress the ProtoAllocator to try and reproduce the issue at https://github.com/stackrox/collector/issues/838

## Checklist

**Automated testing**
  - ~~[ ] Added unit tests~~

### DIFF
--- a/collector/test/ProtoAllocatorTest.cpp
+++ b/collector/test/ProtoAllocatorTest.cpp
@@ -1,0 +1,48 @@
+/** collector
+
+A full notice with attributions is provided along with this source code.
+
+This program is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License version 2 as published by the Free Software Foundation.
+
+This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with this program; if not, write to the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+* In addition, as a special exception, the copyright holders give
+* permission to link the code of portions of this program with the
+* OpenSSL library under certain conditions as described in each
+* individual source file, and distribute linked combinations
+* including the two.
+* You must obey the GNU General Public License in all respects
+* for all of the code used other than OpenSSL.  If you modify
+* file(s) with this exception, you may extend this exception to your
+* version of the file(s), but you are not obligated to do so.  If you
+* do not wish to do so, delete this exception statement from your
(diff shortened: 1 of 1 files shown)


misberner (changes requested):
Could you try printing the value of `default_block_alloc` after `ProtoAllocator.h:40`? I would like to see if it's indeed null, and how this is even possible given https://chromium.googlesource.com/external/github.com/google/protobuf/+/8d5d7cc6d03b00fcb7162525100fbd4cc0d1929d/src/google/protobuf/arena.h#103.

And maybe just print the value of `google::protobuf::ArenaOptions().block_alloc` at the beginning of the test.

ovalenti:
When running with the debugger, I could note that default_block_alloc is nullptr as well as google::protobuf::ArenaOptions().block_alloc;

It is initialized to null at initialization and never updated. It seems like the convention is to use null as a marker for the default value. The default case is later handled here https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/arena.cc#L86

I launched the same test with a null value as allocator in the options and it passed.

ovalenti:
@misberner, I did a bit of archaeology and the default value strategy changed two years ago in https://github.com/protocolbuffers/protobuf/commit/5c028d6cf42e426d47f5baa6ea3f0f5c86b97beb#diff-dabb7d9784d266c7ecd19ed77ff1574de2c87ede037d26ba907f4b371e8c9748L145

This issue is thus very likely due to an update of our protobuf dependency.

ovalenti:
The drawback of the proposed fix is that we won't have a warning when we hit the pool limit.

misberner:
@ovalenti thanks for finding this! I know I had looked at the arena option implementations several times, no idea why it was always the old protobuf code that popped up for me...

I think just using the default values is probably fine. I don't know how much value the warning message provides - apparently, this never got triggered in our testing, and if it happens at a customer, we won't be able to derive much benefit from it. This sounds like more of a use case for telemetry (e.g., `collector_proto_arena_extra_block_allocations_count`).

My understanding is that the default was changed to `nullptr` such that the actual arena allocator can choose its own allocation/deallocation functions in that regard. In any case, I think when setting a custom (non-`nullptr`) `block_alloc`, one MUST also set a `block_dealloc` - if the decision on what to use for alloc/free is delegated to the allocator, that allocator can only make a decision on what to use for freeing if it's also in charge of deciding how to allocate. Obviously you can write a small function that just calls `::operator delete`, but again, I doubt that the log message is important enough to take the freedom as to how to alloc/dealloc from the arena allocator.

ovalenti:
> I think just using the default values is probably fine. I don't know how much value the warning message provides - apparently, this never got triggered in our testing

This sounds reasonable, let me discard the last change then.

misberner (approved):

robbycochran (approved):

PR #847, opened 2022-09-29
Jv rox 12855 add pid to endpoint information in connscraper

JoukoVirtanen:
## Description

Connscraper adds the PID of a processes that opened an endpoint to the endpoint. This also adds a make target, connscrape, which runs connscrape in a docker container, which is also running netcat.

## Checklist
- [x] Ran connscrape make target

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

See checklist.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,11 @@ endif
 collector: builder
 	make -C collector container/bin/collector
 
+.PHONY: connscrape
+connscrape:
+	make -C collector connscrape ||\
+		( .openshift-ci/slack/notify-if-needed.sh "connscrape" $$? )
+
 .PHONY: unittest
 unittest:
 	make -C collector unittest ||\
--- a/collector/Makefile
+++ b/collector/Makefile
@@ -39,6 +39,7 @@ post-build:
 container/bin/collector: 
 	$(MAKE) pre-build
 	$(MAKE) cmake-build/collector 
+	$(MAKE) post-build
 
 .PHONY: build-connscrape-test
 build-connscrape-test:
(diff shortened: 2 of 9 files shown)


collector/lib/NetworkConnection.h
```diff
@@ -347,12 +347,13 @@ size_t Hash(const L4ProtoPortPair& pp);
...
-      : container_(std::move(container)), endpoint_(endpoint), l4proto_(l4proto) {}
+  ContainerEndpoint(std::string container, const Endpoint& endpoint, L4Proto l4proto, int pid = -1)
+      : container_(std::move(container)), endpoint_(endpoint), l4proto_(l4proto), pid_(pid) {}
 
   const std::string& container() const { return container_; }
   const Endpoint& endpoint() const { return endpoint_; }
   const L4Proto l4proto() const { return l4proto_; }
+  const int pid() const { return pid_; }
 
   bool operator==(const ContainerEndpoint& other) const {
```
ovalenti, line 358: Not absolutely sure about the endpoints, but for the connections, the uniqueness in the big map is determined through operator==()
If we want to be able to store several endpoints having the same port but distinct PIDs, then such endpoints should compare as being different.

Makefile
```diff
@@ -37,6 +37,11 @@ endif
 collector: builder
 	make -C collector container/bin/collector
 
+.PHONY: connscrape
+connscrape:
+	make -C collector connscrape ||\
+		( .openshift-ci/slack/notify-if-needed.sh "connscrape" $$? )
```
Molter73, line 43: Notification are specific to CI, we should probably not do this as part of the make target so we can safely run it locally.
```suggestion
	make -C collector connscrape
```

collector/Makefile
```diff
@@ -41,6 +41,17 @@ container/bin/collector:
 	$(MAKE) cmake-build/collector 
 	$(MAKE) post-build
 
+.PHONY: build-connscrape-test
+build-connscrape-test:
+	docker build -f $(CURDIR)/connscrape-test/Dockerfile -t connscrape-test $(CURDIR)/connscrape-test
+
+connscrape: build-connscrape-test
```
Molter73, line 48: We probably want to rename this as `connscrape-test`or something of the sorts.

PR #850, opened 2022-09-30
Build a collector image using the CPaaS drivers

Molter73:
## Description

Make some minor changes in order to allow for integration tests to run with CPaaS built drivers. In broad strokes the changes are:
- Create a collector image that holds the CPaaS built drivers.
- Adjust the image used for testing depending on whether we want to test upstream drivers or CPaaS drivers.

## Checklist

## Testing Performed

- [x] Green CI should be enough.
- [x] New jobs rehearsed on openshift/release#32765

### DIFF
--- a/.openshift-ci/build/Dockerfile.cpaas
+++ b/.openshift-ci/build/Dockerfile.cpaas
@@ -1,0 +1,17 @@
+FROM quay.io/stackrox-io/apollo-ci:collector-0.3.44-1-gb00ffc52af as driver-cache
+
+COPY /kernel-modules/MODULE_VERSION /MODULE_VERSION
+COPY /.openshift-ci/build/build-driver-cache.sh /scripts/build-driver-cache.sh
+
+# Pull the CPaaS built drivers from GCP
+RUN export MODULE_VERSION="$(cat /MODULE_VERSION)" && \
+    mkdir -p "/kernel-modules/${MODULE_VERSION}" && \
+    gsutil -m rsync -r \
+        "gs://collector-modules-osci-public/cpaas//$MODULE_VERSION/" \
+        "/kernel-modules/$MODULE_VERSION/" || true
+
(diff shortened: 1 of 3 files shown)


.openshift-ci/jobs/push-images/push-images.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -eo pipefail
+set -exo pipefail
```
Molter73, line 2: - [ ] Roll this back before merging.

.openshift-ci/build/Dockerfile.cpaas
```diff
@@ -0,0 +1,17 @@
...
+# Pull the CPaaS built drivers from GCP
+RUN export MODULE_VERSION="$(cat /MODULE_VERSION)" && \
+    mkdir -p "/kernel-modules/${MODULE_VERSION}" && \
+    gsutil -m rsync -r \
+        "gs://collector-modules-osci-public/cpaas//$MODULE_VERSION/" \
+        "/kernel-modules/$MODULE_VERSION/" || true
+
+RUN /scripts/build-driver-cache.sh
+
+FROM quay.io/stackrox-io/collector:3.11.0-slim
```
Stringy, line 15: Does this need to parameterised or are we expecting to update this tag for each release?

Stringy (approved):
One minor comment but otherwise LGTM!

.openshift-ci/build/Dockerfile.cpaas
Molter73: For now, it's handled by OSCI and it runs with the collector version that is currently built by the repo. If we want to add a way to run this manually, we can parametrize it to use the output of `make tag` instead of a fixed tag, but I don't know if we want to spend the extra time at this point, WDYT?

PR #855, opened 2022-10-03
disable ebpf on s390x

kcrane:
## Description

Disable ebpf probe on s390x until its ready for production use.

## Testing Performed

ran unit and integration tests

### DIFF
--- a/collector/lib/HostInfo.h
+++ b/collector/lib/HostInfo.h
@@ -112,7 +112,11 @@ struct KernelVersion {
     if (kernel < 4 || (kernel == 4 && major < 14)) {
       return false;
     }
+#ifdef __s390x__
+    return false;
+#else
     return true;
+#endif
   }
 
   // Whether or not the kernel has secure_boot option present in boot_params.
--- a/collector/lib/HostInfo.h
+++ b/collector/lib/HostInfo.h
@@ -109,12 +109,14 @@ struct KernelVersion {
   // Whether or not the kernel has built-in eBPF support
   // Anything before 4.14 doesn't have support, anything newer does.
   bool HasEBPFSupport() const {
+#ifdef __s390x__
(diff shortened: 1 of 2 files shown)


collector/lib/HostInfo.h
```diff
@@ -112,7 +112,11 @@ struct KernelVersion {
     if (kernel < 4 || (kernel == 4 && major < 14)) {
       return false;
     }
+#ifdef __s390x__
+    return false;
+#else
     return true;
+#endif
```
Molter73, line 119: This is a nitpick, but maybe we can move the kernel version check into the else section of this ifdef. It's silly, but there isn't much point in check kernel versions if we are always returning false in s390x.

Stringy: I think this is the wrong place to make this change, since now the KernelVersion is incorrectly reporting that a given kernel version does not support eBPF on s390x. 

It would be better to find and store the architecture in `HostInfo` and either enforce this logic in `HostInfo::HasEBPFSupport` (again, not ideal) or write a HostHeuristic that enforces kernel module when we're on s390x

kcrane: @Stringy I failed to mention in the description that we are actively working on ebpf for s390x and expect it to be completed either in upcoming or following release.  I considered doing a runtime check, but figured the added complexity unnecessary as the code will be temporary.  wdyt?

Stringy: Hi @kcrane - If it's expected to span one or more releases then I'd be inclined to implement the more complex solution if only to make the behaviour a bit clearer, but I totally see what you're saying and I'm happy if we go with this lighter solution, though could we have a note here (`TODO` or similar) to describe why it's there and what conditions need to be in place before it can be removed

Molter73 (approved):
/lgtm

PR #860, opened 2022-10-05
Add re2 dependency

Molter73:
## Description

Adds the re2 dependency for the collector builder. This is needed in newer versions for Falco and is required to build #849, but because of the way OSCI handles the root image build, we need to add the dependency to master before bumping the version of falco.

## Checklist

## Testing Performed

Green CI is enough.


JoukoVirtanen (approved):
/lgtm

PR #861, opened 2022-10-05
ROX-13114 add process info to network endpoints

ovalenti:
## Description

Originator process information is added to network endpoint objects sent to Sensor.
This information comes from a ProcessStore which deduplicates process information between endpoints sharing the same originator. Ultimately, the information comes from a new Process scraper.

## Checklist

**Automated testing**
  - [x] Added unit tests (some tests for the ProcessStore are planned in a subsequent PR)

### DIFF
--- a/collector/connscrape.cpp
+++ b/collector/connscrape.cpp
@@ -32,7 +32,7 @@ using namespace collector;
 
 namespace {
 
-BoolEnvVar scrape_endpoints("SCRAPE_ENDPOINTS", false);
+BoolEnvVar scrape_endpoints("SCRAPE_ENDPOINTS", true);
 
 }  // namespace
 
@@ -63,7 +63,7 @@ int main(int argc, char** argv) {
     std::cout << std::endl
               << "Endpoints:" << std::endl;
     for (const auto& ep : endpoints) {
-      std::cout << " " << ep << std::endl;
+      std::cout << " " << ep << " " << ep.pid() << std::endl;
     }
   }
 
--- a/collector/connscrape.cpp
+++ b/collector/connscrape.cpp
@@ -63,7 +63,7 @@ int main(int argc, char** argv) {
     std::cout << std::endl
               << "Endpoints:" << std::endl;
     for (const auto& ep : endpoints) {
-      std::cout << " " << ep << " " << ep.pid() << std::endl;
+      std::cout << " " << ep << " " << ep.originator()->getPid() << std::endl;
     }
   }
 
--- a/collector/connscrape.cpp
+++ b/collector/connscrape.cpp
@@ -43,7 +43,9 @@ int main(int argc, char** argv) {
     proc_dir = argv[1];
   }
 
-  ConnScraper scraper(proc_dir);
+  std::shared_ptr<ProcessScraper> process_scraper = std::make_shared<ProcessScraper>(proc_dir);
+  std::shared_ptr<ProcessStore> process_store = std::make_shared<ProcessStore>(process_scraper);
+  ConnScraper scraper(proc_dir, process_store);
   std::vector<Connection> conns;
   std::vector<ContainerEndpoint> endpoints;
 
--- a/collector/connscrape.cpp
+++ b/collector/connscrape.cpp
@@ -25,8 +25,8 @@ You should have received a copy of the GNU General Public License along with thi
 
 #include <iostream>
 
-#include "ConnScraper.h"
 #include "EnvVar.h"
+#include "ProcfsScraper.h"
 
 using namespace collector;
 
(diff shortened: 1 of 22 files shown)


collector/lib/ConnScraper.cpp
```diff
@@ -190,6 +190,60 @@ int ReadHexBytes(const char* p, const char* endp, void* buf, int chunk_size, int
...
+
+    info.comm = info.exe_path = buffer;
+
+    if (buffer[0] == '/')
+      info.comm = strrchr(buffer, '/') + 1;
+  } else {
+    CLOG(ERROR) << "Could not read 'exe' for " << process_id << ": " << StrError();
+  }
+
+  // parse xx/cmdline
```
JoukoVirtanen, line 216: This could be split into two functions.

collector/connscrape.cpp
```diff
@@ -63,7 +63,7 @@ int main(int argc, char** argv) {
     std::cout << std::endl
               << "Endpoints:" << std::endl;
     for (const auto& ep : endpoints) {
-      std::cout << " " << ep << " " << ep.pid() << std::endl;
+      std::cout << " " << ep << " " << ep.originator()->getPid() << std::endl;
```
JoukoVirtanen, line 66: This is something I intended to change in my own PR. This is not the best place to print the pid. Instead change the << operator for ContainerEndpoint and Process.

collector/lib/ConnScraper.cpp
```diff
@@ -190,6 +190,60 @@ int ReadHexBytes(const char* p, const char* endp, void* buf, int chunk_size, int
...
 }
 
+struct ProcessInfo {
+  std::string comm;
+  std::string exe;
+  std::string exe_path;
+  std::string args;
+};
+
+static void ReadProcessInfo(const char* process_id, int dirfd, struct ProcessInfo& info) {
```
JoukoVirtanen, line 200: I thought we might put functions that read the process information into a separate class from ConnScraper.

collector/test/ConnTrackerTest.cpp
```diff
@@ -1246,6 +1246,7 @@ TEST(ConnTrackerTest, TestComputeDeltaAfterglowBenchmark2) {
   std::cout << "Time taken by ComputeDeltaAfterglow= " << dur.count() << " ms\n";
 }
 
+/*
```
JoukoVirtanen, line 1249: Probably best to keep the tests, but modify them.

collector/lib/ConnScraper.cpp
```diff
@@ -190,6 +190,60 @@ int ReadHexBytes(const char* p, const char* endp, void* buf, int chunk_size, int
   return i;
 }
 
+struct ProcessInfo {
```
JoukoVirtanen, line 193: We should be more consistent with the naming of the classes and their fields. They should be consistent with the protobufs. What is the difference between exe and comm? What is the purpose of having this struct and the class Process, which has the same fields. It feels redundant.

collector/lib/ConnScraper.cpp
```diff
@@ -433,7 +487,14 @@ bool ReadContainerConnections(const char* proc_path, std::vector<Connection>* co
       auto emplace_res = conns_by_ns.emplace(netns_inode, NSNetworkData());
       if (emplace_res.second) {
         auto& ns_network_data = emplace_res.first->second;
-        if (!GetConnections(dirfd, pid, &ns_network_data.connections, listen_endpoints ? &ns_network_data.listen_endpoints : nullptr)) {
+
+        ProcessInfo info;
+        ReadProcessInfo(curr->d_name, dirfd, info);
+
+        std::shared_ptr<Process> process =
+            ProcessStore::global_instance.FetchReference(Process(container_id, info.comm, info.exe, info.exe_path, info.args, pid));
```
JoukoVirtanen, line 495: It seems strange that we are reading from proc and then looking for the process information from the process store. Wouldn't it make more sense to look in the process store and then read proc if the process isn't in the process store.

collector/lib/ConnScraper.cpp
ovalenti: I agree. This will be moved to to the Process factory soon, I will change that during the transition.

collector/connscrape.cpp
ovalenti: Please, feel free to do so; I'll handle the conflicts.

collector/lib/ConnScraper.cpp
ovalenti: Correct, this is the current plan.

collector/lib/ConnScraper.cpp
ovalenti: I don't agree on this point. Protobuf messages are loaded with the semantics of processing information in the Stackrox context, while Collector internal objects represent Linux resources. For instance, a Process is a process, not the signal event.

collector/lib/ConnScraper.cpp
ovalenti: This is going to change with the moving of the scraping functions to a dedicated factory. The process information will be fetched by providing only the PID.

JoukoVirtanen (approved):
LGTM!

PR #865, opened 2022-10-06
collector: force ebpf if SecureBoot undetermined only on x86_64

yselkowitz:
## Description

Although I wonder if this block shouldn't be removed entirely (after all, why assume Secure Boot if none of the signs thereof could be detected?), at a minimum, this logic should not be applied to other architectures.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

### DIFF
--- a/collector/lib/HostHeuristics.cpp
+++ b/collector/lib/HostHeuristics.cpp
@@ -140,11 +140,13 @@ class SecureBootHeuristic : public Heuristic {
       hconfig->SetCollectionMethod("ebpf");
     }
 
+#ifdef __x86_64__
     if (sb_status == SecureBootStatus::NOT_DETERMINED) {
       CLOG(WARNING) << "SecureBoot status could not be determined. "
                     << "Switching to eBPF based collection.";
       hconfig->SetCollectionMethod("ebpf");
     }
+#endif
   }
 };
 


robbycochran:
@erthalion What are your thoughts on removing this block entirely? On non-x86 architectures, we may not have eBPF support.

hbrueckner:
Hi @robbycochran , @yselkowitz ,

On `s390x` secure boot is supported when running natively in LPAR and when booting from FCP devices. RHOCP on zSystems/LinuxONE does not have support to run natively in LPAR and, thus, secure boot will not be available.

On the other side, the `HostInfo.cpp` considers x86 only at this point in time, I wonder whether we should add some `s390x` secure boot detection (enabled if ` /sys/firmware/ipl/secure` returns 1) at some future point in time (enhancement issue?)

cc: @kcrane for awareness

P.S. For reference, here are some more information on that topic: https://www.ibm.com/docs/en/linux-on-systems?topic=linux-current-ipl-parameters#scsiipl_work

Molter73:
Thanks for the additional information @hbrueckner, as you well noted this heuristic only accounts for x86 secure boot. Having proper support for secure boot on s390x and/or ppc64le would be amazing and I agree that we could add that in a follow up.

Molter73 (approved):

hbrueckner:
Hi @Molter73,

> Thanks for the additional information @hbrueckner, as you well noted this heuristic only accounts for x86 secure boot. Having proper support for secure boot on s390x and/or ppc64le would be amazing and I agree that we could add that in a follow u

already started to have a look into this. Stay tuned.

hbrueckner:
Hi @Molter73 ,

> > Thanks for the additional information @hbrueckner, as you well noted this heuristic only accounts for x86 secure boot. Having proper support for secure boot on s390x and/or ppc64le would be amazing and I agree that we could add that in a follow 

For s390x, the logic would like this:
````cpp
SecureBootStatus HostInfo::GetSecureBootOnS390x() {
  std::string is_secure;
  std::string ipl_secure_path = GetHostPath("/sys/firmware/ipl/secure");

  std::ifstream ipl_secure(ipl_secure_path, std::ios::in);

  /* If "secure" attribute is not available, check IPL type for the
   * "unknown" case where secure boot cannot be determined.
   */
  if (!bool(ipl_secure)) {
    std::string ipl_type_path = GetHostPath("/sys/firmware/ipl/ipl_type");
    std::ifstream ipl_type(ipl_type_path, std::ios::in);
    std::string type;
    if (ipl_type.is_open() && std::getline(ipl_type, type))
            if (type.compare("unknown") == 0)
                    return SecureBootStatus::NOT_DETERMINED;

    return SecureBootStatus::DISABLED;
  }

  if (!ipl_secure.is_open()) {
    CLOG(WARNING) << "Failed to open " << ipl_secure_path;
    return SecureBootStatus::NOT_DETERMINED;
  }

  if (!std::getline(ipl_secure, is_secure)) {
    CLOG(WARNING) << "Failed to read " << ipl_secure_path;
    return SecureBootStatus::NOT_DETERMINED;
  }

  if (is_secure.compare("1") == 0)
    return SecureBootStatus::ENABLED;

  return SecureBootStatus::DISABLED;
}
````

However, the question will be how to best integrate this, e.g., add this to the existing `x86` routines or refactor this and add a secure boot function that is implemented by each architecture. Also awaiting feedback from `ppc64`, not sure about `aarch64` (and if the use similar UEFI logic for secure boot).

yselkowitz:
aarch64 (at least, those platforms supported in RHEL) is UEFI, and afaik there is a concept of Secure Boot but I'm not sure about the implementation details.  The eBPF probes currently do not build on aarch64, and even when patched enough to compile, they crash, so my PoC has been `KernelModule`-only so far.

erthalion:
> @erthalion What are your thoughts on removing this block entirely? On non-x86 architectures, we may not have eBPF support.

The idea behind was to turn on eBPF mode in this case, since it's going to be
the default anyway, assuming it's supported. This last assumption doesn't
necessarily hold on non-x86. My suggestion would be to keep it under x86-only
flags, but if it leads to unnecessary complications it could be removed
altogether (I assume the case of undefined secure boot would be quite rare,
there are no reports about this in practice yet).

Give me a moment, I'll take a look at the proposed solution for detection for
other architectures.

erthalion (approved):

erthalion:
> Give me a moment, I'll take a look at the proposed solution for detection for
other architectures.

The moment was quite long one, sorry. The snippet looks good, I guess it could be turned into a follow up PR.

PR #867, opened 2022-10-08
Jv integration tests for networksignalhandler endpoints

JoukoVirtanen:
## Description

Makes it so that TestNetworkFlows checks for endpoints as well as connections. Right now NetworkSignalHandler does not report endpoints, and TestNetworkFlows only checks for connections. Probably someday NetworkSignalHandler will report endpoints. 

## Checklist

**Automated testing**
  - [x] Added integration tests

## Testing Performed

See above

### DIFF
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -367,6 +367,11 @@ func (s *ProcessNetworkTestSuite) TestProcessLineageInfo() {
 func (s *ProcessNetworkTestSuite) TestNetworkFlows() {
 
 	// Server side checks
+
+	// NetworkSignalHandler does not currently report endpoints
+	_, err := s.GetEndpoints(s.serverContainer)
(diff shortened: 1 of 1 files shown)


integration-tests/integration_test.go
```diff
@@ -367,6 +367,11 @@ func (s *ProcessNetworkTestSuite) TestProcessLineageInfo() {
 func (s *ProcessNetworkTestSuite) TestNetworkFlows() {
 
 	// Server side checks
+
+	// NetworkSignalHandler does not currently report endpoints
```
Stringy, line 371: Could we expand this comment (and the one below) to describe a bit more about why this is the case, and what needs to exist before we adjust this test

JoukoVirtanen: Done

Stringy (approved):
LGTM!

PR #868, opened 2022-10-08
ROX-13015 collector integration tests should check for process info in endpoints

JoukoVirtanen:
## Description

The ConnScraper integration test which checks endpoint information sent by collector will now check that the endpoint information includes the process originator.

This PR is the parent of https://github.com/stackrox/collector/pull/870

The difference between the two PRs is that this expects the process information to be nil and the other expects to find actual process information.

## Checklist

**Automated testing**
  - [x] Added integration tests

## Testing Performed

See above


ovalenti:
We will probably want to expect the originator only when the feature flag is set. This will also check that the feature flag behaves as expected as well.

Stringy (approved):

PR #869, opened 2022-10-11
ROX-11447: refactor kernel driver start up

Stringy:
## Description

Refactors SysdigService and CollectorService to centralise the kernel driver insertion logic to better inform start up diagnostics. This also decouples some components within CollectorService.

The kernel driver setup logic is now in a set of classes of its own, and then initialised by the SysdigService. 

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Some local testing, but CI should uncover any significant problems with this refactor.

### DIFF
--- a/collector/lib/KernelDriver.cpp
+++ b/collector/lib/KernelDriver.cpp
@@ -34,7 +34,7 @@ std::string constructArgsStr(std::unordered_map<std::string, std::string>& args)
   return stream.str();
 }
 
-int doInsertModule(std::vector<char>& image, std::unordered_map<std::string, std::string>& args) {
+int insertModule(std::vector<char>& image, std::unordered_map<std::string, std::string>& args) {
   std::string args_str = constructArgsStr(args);
   CLOG(DEBUG) << "Kernel module arguments: " << args_str;
 
@@ -115,7 +115,7 @@ bool KernelDriverModule::insert(const std::vector<std::string>& syscalls, std::s
 
   // Attempt to insert the module. If it is already inserted then remove it and
   // try again
-  int result = doInsertModule(image, module_args);
+  int result = insertModule(image, module_args);
   while (result != 0) {
     if (errno == EEXIST) {
       // note that we forcefully remove the kernel module whether or not it has a non-zero
@@ -128,7 +128,7 @@ bool KernelDriverModule::insert(const std::vector<std::string>& syscalls, std::s
                   << ". Aborting...";
       return false;
     }
-    result = doInsertModule(image, module_args);
+    result = insertModule(image, module_args);
   }
 
   CLOG(INFO) << "Successfully inserted kernel module " << path << ".";
--- a/collector/lib/KernelDriver.cpp
+++ b/collector/lib/KernelDriver.cpp
@@ -8,31 +8,15 @@ extern "C" {
 #include <sys/stat.h>
 }
 
+#include <fstream>
+#include <iterator>
+
 #define init_module(module_image, len, param_values) syscall(__NR_init_module, module_image, len, param_values)
 #define delete_module(name, flags) syscall(__NR_delete_module, name, flags)
 
 namespace collector {
 
 namespace {
-
-int read_module(int fd, void* buf, int buflen) {
-  unsigned char* p = static_cast<unsigned char*>(buf);
-  int n, i = 0;
-  while (i < buflen) {
-    n = read(fd, p + i, buflen - i);
-    if (n < 0) {
-      if (errno == EINTR)
-        continue;
-      else
-        return n;
-    } else if (n == 0) {
-      return i;
-    } else {
-      i += n;
-    }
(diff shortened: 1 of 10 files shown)


collector/lib/NRadix.cpp
```diff
@@ -73,8 +73,8 @@ bool NRadixTree::Insert(const IPNet& network) const {
       // Reset and move to lower part.
       bit = 0x8000000000000000ULL;
       if (network.bits() >= 64) {
-        *addr_p++;
-        *net_mask_p++;
```
Stringy, line 77: These aren't strictly part of this PR but the compiler warnings were annoying so I thought it was worth fixing

collector/lib/KernelDriver.cpp
```diff
@@ -0,0 +1,138 @@
...
+  module.seekg(0, std::ios_base::end);
+  size_t image_size = module.tellg();
+  module.seekg(0, std::ios_base::beg);
+
+  std::vector<char> image;
+  image.reserve(image_size);
+
+  std::copy(std::istreambuf_iterator<char>(module),
+            std::istreambuf_iterator<char>(),
+            std::back_inserter(image));
```
Stringy, line 110: @kcrane - I have adjusted the kernel module reading code to be a little more idiomatic C++, but I believe it behaves the same way as before. Could you have a look in case I've missed something that might affect s390x?

kcrane: all integration tests pass on s390x with your branch

collector/lib/KernelDriver.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include <iterator>
+
+#define init_module(module_image, len, param_values) syscall(__NR_init_module, module_image, len, param_values)
+#define delete_module(name, flags) syscall(__NR_delete_module, name, flags)
+
+namespace collector {
+
+namespace {
+
+std::string constructArgsStr(std::unordered_map<std::string, std::string>& args) {
```
JoukoVirtanen, line 21: [nit] It seems that the style within collector is to start each word in a function with a capital letter including the first.

collector/lib/KernelDriver.cpp
```diff
@@ -0,0 +1,138 @@
...
+    } else {
+      stream << " ";
+    }
+    stream << entry.first << "=" << entry.second;
+  }
+
+  return stream.str();
+}
+
+int doInsertModule(std::vector<char>& image, std::unordered_map<std::string, std::string>& args) {
```
JoukoVirtanen, line 37: [nit] "do" is not needed in this function name.

collector/lib/KernelDriver.cpp
Stringy: I believe that's only true for public functions/methods, and everything else starts with a lowercase letter, similar to Go

collector/lib/SysdigService.cpp
```diff
@@ -79,6 +68,38 @@ void SysdigService::Init(const CollectorConfig& config, std::shared_ptr<Connecti
...
+    useEbpf = true;
+    KernelDriverEBPF driver;
+    if (!driver.Setup(config, SysdigService::kProbePath)) {
+      CLOG(ERROR) << "Failed to setup eBPF probe";
+      return false;
+    }
+    inspector_->set_bpf_probe(kProbePath);
+  } else {
+    KernelDriverModule driver;
+    driver = KernelDriverModule();
```
ovalenti, line 93: Is this line necessary ?

ovalenti (commented):
The new layout is a much better. Here are a few things I found while going through it.

collector/lib/SysdigService.cpp
```diff
@@ -79,6 +68,38 @@ void SysdigService::Init(const CollectorConfig& config, std::shared_ptr<Connecti
...
+    if (!driver.Setup(config, SysdigService::kModulePath)) {
+      CLOG(ERROR) << "Failed to setup Kernel module";
+      return false;
+    }
+  }
```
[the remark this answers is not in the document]
Stringy: Yeah that's a fair point. I was originally trying to go for proper polymorphism but ended up with this mechanism, but I think the interface at least ensures there's a consistent API for each implementation, even if we don't take advantage of any polymorphism. What do you think?

collector/lib/SysdigService.cpp
Stringy: Not even a little. I'll fix it :)

collector/lib/SysdigService.cpp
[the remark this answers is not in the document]
ovalenti: It probably doesn't hurt. But I would not be shocked if the drivers had distinct Setup() requirements.

ovalenti (approved):
Looks good to me

PR #874, opened 2022-10-17
Make 'make linters' fail correctly

Molter73:
## Description

Make the `lint.sh` script fail when at least one of the underlying linters fail.

Also fix some small formatting errors that were not being alerted on.

## Checklist

## Testing Performed

Green CI should be enough.


ovalenti (approved):
LGTM

PR #875, opened 2022-10-18
Fix bucket used for cpaas images

Molter73:
## Description

The additional `/` in the bucket name makes it so `gsutil` fails to properly download drivers into the image.

## Checklist

## Testing Performed

- [x] rehearsed on openshift/release#32765


ovalenti (approved):
Looks like a small but necessary change.

PR #881, opened 2022-10-24
ROX-11447: write fatal log messages to termination log

Stringy:
## Description

Writes all `FATAL` log messages to [/dev/termination-log](https://kubernetes.io/docs/tasks/debug/debug-application/determine-reason-pod-failure/) if that device exists. In k8s, it is used to enrich container status information, and allows us to report better errors through k8s APIs for diagnostics of Collector startup. 

Commands for getting this enriched information:

```bash
$ kubectl describe pod -n stackrox collector-cr2mf
# ...
    Last State:     Terminated
      Reason:       Error
      Message:      No suitable kernel object downloaded
      Exit Code:    1
      Started:      Fri, 21 Oct 2022 11:50:56 +0100
      Finished:     Fri, 21 Oct 2022 11:51:25 +0100
# ...

$ kubectl get pods -n stackrox collector-cr2mf -o jsonpath='{.status.containerStatuses[].lastState}' | jq .
{
  "terminated": {
    "containerID": "docker://33abaef2f6b18e06e4c8ac8dbb7c3195355faa34b288731203fcbdde7b3a1920",
    "exitCode": 1,
    "finishedAt": "2022-10-21T10:57:09Z",
    "message": "No suitable kernel object downloaded",
    "reason": "Error",
    "startedAt": "2022-10-21T10:56:30Z"
  }
}
```

This also includes some improvements to `FATAL` log messages.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Tested locally on k8s running in Colima (which has the added advantaged of using a kernel for which we don't have a kernel driver for, so collector always hits a `FATAL` error)

### DIFF
--- a/collector/lib/Logging.h
+++ b/collector/lib/Logging.h
@@ -55,7 +55,10 @@ bool ParseLogLevelName(std::string name, LogLevel* level);
 const char* GetGlobalLogPrefix();
 void SetGlobalLogPrefix(const char* prefix);
 
+void WriteTerminationLog(std::string message);
+
 const size_t LevelPaddingWidth = 7;
+const std::string TerminationLog = "/dev/termination-log";
 
 class LogMessage {
  public:
@@ -93,6 +96,7 @@ class LogMessage {
               << std::endl;
 
     if (level_ == LogLevel::FATAL) {
+      WriteTerminationLog(buf_.str());
       exit(1);
     }
   }
--- a/collector/lib/Logging.h
+++ b/collector/lib/Logging.h
@@ -58,7 +58,6 @@ void SetGlobalLogPrefix(const char* prefix);
 void WriteTerminationLog(std::string message);
 
 const size_t LevelPaddingWidth = 7;
-const std::string TerminationLog = "/dev/termination-log";
 
 class LogMessage {
  public:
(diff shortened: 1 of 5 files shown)


collector/lib/Logging.h
```diff
@@ -55,7 +55,10 @@ bool ParseLogLevelName(std::string name, LogLevel* level);
 const char* GetGlobalLogPrefix();
 void SetGlobalLogPrefix(const char* prefix);
 
+void WriteTerminationLog(std::string message);
+
 const size_t LevelPaddingWidth = 7;
+const std::string TerminationLog = "/dev/termination-log";
```
Molter73, line 61: Do we need a full-on string? Can we use a raw `char*` instead?

Molter73 (approved):
/lgtm

Just a minor comment on a string that we could probably downgrade to a raw pointer.

collector/lib/Logging.h
Stringy: I had to move this into `Logging.cpp` to get the linker to stop complaining about redefinitions once I changed the type but it's probably a better place for it anyway

Stringy:
The two failing tests are timeout flakes, unrelated to this change and from the logs all tests are passing so I'll merge

PR #882, opened 2022-10-24
ROX-13249 plop: use falco as process info source

ovalenti:
## Description

Using /proc as the only source of process information exposes us to cmdline being modified by the process (as is the case for nginx). Here, we replace our ProcessStore::ISource by the falco library, which keeps information as seen in the fork syscall.

## Checklist

**Automated testing**
  - ~~[ ] Added unit tests~~
  - ~~[ ] Added integration tests~~ Those are already in place and show the issues

### DIFF
--- a/collector/lib/SysdigService.h
+++ b/collector/lib/SysdigService.h
@@ -93,7 +93,7 @@ class SysdigService : public Sysdig, public ProcessStore::ISource {
 
   void AddSignalHandler(std::unique_ptr<SignalHandler> signal_handler);
 
-  mutable std::mutex sysdig_mutex_;
+  mutable std::mutex libsinsp_mutex_;
   std::unique_ptr<sinsp> inspector_;
   std::unique_ptr<sinsp_chisel> chisel_;
   std::vector<SignalHandlerEntry> signal_handlers_;
--- a/collector/lib/SysdigService.h
+++ b/collector/lib/SysdigService.h
@@ -93,6 +93,7 @@ class SysdigService : public Sysdig, public ProcessStore::ISource {
 
   void AddSignalHandler(std::unique_ptr<SignalHandler> signal_handler);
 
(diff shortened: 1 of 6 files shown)


robbycochran:
`SysdigService::ByPID` will be executed for every connection event but only utilized when `data.state == TCP_LISTEN`. Did you consider the performance impact? What about lazily getting process details from Falco when a listen endpoint is encountered?

collector/lib/SysdigService.h
```diff
@@ -89,6 +93,7 @@ class SysdigService : public Sysdig {
 
   void AddSignalHandler(std::unique_ptr<SignalHandler> signal_handler);
 
+  mutable std::mutex sysdig_mutex_;
```
robbycochran, line 96: The Falco project has endeavored to remove `sysdig` from the codebase after donation to the CNCF, please rename this to remove `sysdig`, something like `libsinsp_mutex` maybe?

ovalenti:
> `SysdigService::ByPID` will be executed for every connection event but only utilized when `data.state == TCP_LISTEN`. Did you consider the performance impact? What about lazily getting process details from Falco when a listen endpoint is encountered?

@robbycochran, I blindly replaced PID by process references and missed the performance impact. Thanks for spotting this.

The current code uses PID until we know that we need process information, and does the lookup from there.

collector/lib/SysdigService.h
ovalenti: Renamed as suggested.

Do you think that we should rename the whole SysdigService as well ?

Molter73 (commented):
I'll put more general comments on #900

Stringy (commented):
Changes generally look good! wrt the performance impact, do we have any way of measuring the difference? I'll soon re-introduce the benchmarking back into CI (via ansible) and that may help, but I feel we need more targeted and isolated measurements

PR #883, opened 2022-10-27
Some dev-env enhancements

ovalenti:
## Description

These are a few little changes I gathered over time and which should make life a bit easier.

## Checklist
- ~~[ ] Updated documentation accordingly~~ There doesn't seem to be any reference to `make tests`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ cmake-build/
 .vscode/
 .devcontainer/
 .devcontainer.json
+cmake-build/
+build/
+out/
--- a/.gitignore
+++ b/.gitignore
@@ -26,5 +26,4 @@ cmake-build/
 .devcontainer/
 .devcontainer.json
 cmake-build/
-build/
 out/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,6 @@
 project(collector)
 cmake_minimum_required(VERSION 3.15)
+
+enable_testing()
+
 add_subdirectory(collector)
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 BASE_PATH = .
 include Makefile-constants.mk
 
-NPROCS ?= 6
+NPROCS ?= $(shell nproc)
 
 MOD_VER_FILE=$(CURDIR)/kernel-modules/kobuild-tmp/MODULE_VERSION.txt
 
--- a/Makefile
+++ b/Makefile
@@ -88,10 +88,6 @@ image-dev-full: image-dev build-drivers
 		--build-arg max_layer_depth=1 \
 		$(CURDIR)/kernel-modules/container
 
-.PHONY: integration-tests
-integration-tests:
-	make -C integration-tests tests
-
 .PHONY: integration-tests-benchmark
 integration-tests-benchmark:
 	make -C integration-tests benchmark ||\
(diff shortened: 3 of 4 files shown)


Makefile
```diff
@@ -1,6 +1,8 @@
 BASE_PATH = .
 include Makefile-constants.mk
 
+NPROCS ?= 6
```
Molter73, line 4: How about doing something like this instead?
```make
NPROCS ?= $(shell nproc)
```

Makefile
```diff
@@ -85,10 +88,6 @@ image-dev-full: image-dev build-drivers
 		--build-arg max_layer_depth=1 \
 		$(CURDIR)/kernel-modules/container
 
-.PHONY: integration-tests
-integration-tests:
```
Molter73, line 89: Why are we removing this target entirely? It might not be as convenient as some of the others, but if it works I don't see any harm in having it.

builder/Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN dnf -y update \
         bison \
         ca-certificates \
         clang \
+        clang-tools-extra \
```
Molter73, line 18: I don't know if we want to add this package to the builder, it is pretty heavy. We might want to start considering creating a dev image that is based on but separate from the builder image, that way we can have a slimmer builder for CI to use.

CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
 project(collector)
 cmake_minimum_required(VERSION 3.15)
+
+enable_testing()
```
Molter73, line 4: ❤️

.gitignore
```diff
@@ -25,3 +25,6 @@ cmake-build/
 .vscode/
 .devcontainer/
 .devcontainer.json
+cmake-build/
+build/
+out/
```
Molter73, line 30: Just out of curiosity, what tool creates the `out/` directory?

.gitignore
```diff
@@ -25,3 +25,6 @@ cmake-build/
 .vscode/
 .devcontainer/
 .devcontainer.json
+cmake-build/
+build/
```
Molter73, line 29: We might want to unify these two at some point, seems kinda weird we could have both in our repo.

Makefile
ovalenti: I agree, I'll give it a try !

Makefile
ovalenti: I propose to remove it because the `tests` target doesn't exist anymore in the integration-tests subdir.

.gitignore
ovalenti: It is the default build dir for VSCode when using cmake integration

ovalenti:
Thanks for the very valuable review @Molter73 . I will rework it a bit :)

Makefile
Molter73: I totally missed that, let's go ahead and get it outta here then!

.gitignore
Molter73: Weird, I thought it was `build/`, it's fine either way.

builder/Dockerfile
ovalenti: I will keep this change locally for now

.gitignore
ovalenti: I kept only cmake-build and out. I don't know where `build` may be used.

Molter73 (approved):
/lgtm

PR #884, opened 2022-10-27
New minikube candidates

Molter73:
## Description

Since minikube kernel crawling was modified by stackrox/kernel-packer#232, the way collector looks for drivers for it needs to be slightly adjusted as well. This change also makes it so the kernel 'as is' used to try and get the driver, this is to ensure that when minikube is used with the docker driver our driver is properly downloaded. This all means the complete list of candidates to download will be of the form:
- `collector-ebpf-${KERNEL_VERSION}.o.gz`
- `collector-ebpf-${KERNEL_VERSION}-minikube-${MINIKUBE_VERSION}.o.gz`

## Checklist

## Testing Performed

- [x] Manually built the driver and run on minikube.
- [x] ~Run integration tests on minikube with the images built by CI.~ Check the CI built images work on minikube (integration tests won't work properly because of the delay to pick the proper driver).


Stringy (approved):
LGTM!

PR #889, opened 2022-11-03
Add docs on how to debug collector on k8s

Molter73:
## Description

Add instructions on how to build a debug enabled collector image and deploy it to a k8s instance for debugging with GDB.

## Checklist

## Testing Performed

This change only changes documentation.

### DIFF
--- a/docs/how-to-start.md
+++ b/docs/how-to-start.md
@@ -164,5 +164,52 @@ $ make image
 ```
 
 ## Debugging
+### Debugging on a k8s deployment
+First step for debugging collector on kubernetes is to have a debug enabled collector image with `gdb` installed.
+You can build such an image by running the following command at the root of the repository:
+```sh
+CMAKE_BUILD_TYPE=Debug make image-dev
+```
+
+You will also need a full deployment of Stackrox up and running. You can get precise instructions on how to deploy it
+[here](https://github.com/stackrox/stackrox/blob/master/README.md#deploying-stackrox).
+
+With the image built and Stackrox up and running, you can make the following changes to the collector daemonset by
+running `kubectl -n stackrox edit ds/collector`
+
+```yaml
+# ...
+spec:
+  template:
+    spec:
+      containers:
+      - env:
+        # Make sure this variable is edited like this for proper JSON escaping
+        - name: COLLECTOR_CONFIG
+          value: |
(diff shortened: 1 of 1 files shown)


docs/how-to-start.md
```diff
@@ -164,5 +164,52 @@ $ make image
...
+# ...
+spec:
+  template:
+    spec:
+      containers:
+      - env:
+        # Make sure this variable is edited like this for proper JSON escaping
+        - name: COLLECTOR_CONFIG
+          value: |
+            '{"tlsConfig":{"caCertPath":"/var/run/secrets/stackrox.io/certs/ca.pem","clientCertPath":"/var/run/secrets/stackrox.io/certs/cert.pem","clientKeyPath":"/var/run/secrets/stackrox.io/certs/key.pem"}}'
```
erthalion, line 190: Does the `COLLECTOR_CONFIG` have to be configured for debugging, or is it just for the picture to be complete?

Molter73: It needs to be configured like this for escaping to work properly. This is a json string, placed inside a yaml file, that gets put in a shell environment variable and finally passed into collector as an argument by doing `--collector-config=$COLLECTOR_CONFIG`, running normally the configuration looks like this:
```yaml
        - name: COLLECTOR_CONFIG
          value: '{"tlsConfig":{"caCertPath":"/var/run/secrets/stackrox.io/certs/ca.pem","clientCertPath":"/var/run/secrets/stackrox.io/certs/cert.pem","clientKeyPath":"/var/run/secrets/stackrox.io/certs/key.pem"}}'
```
But when run with GDB, I guess it might need one extra level of escaping, otherwise it goes cuckoo and collector fails to parse the string.

erthalion: But if we do `kubectl edit ...`, wouldn't the cli handle escaping correctly for us? I recall all what you've mentioned in the docker compose context.

Molter73: It would be great if that was the case, but that is not what I found when testing it unfortunately. My guess is `kubectl` is doing the thing properly, but then we have printing of the env variable to an argument and then that also goes through `gdb` into collector.

Stringy (approved):
I've run this up and tried it locally and it's pretty much seamless, thanks for documenting this @Molter!

robbycochran:
Nice!

PR #890, opened 2022-11-04
Only append CID on PR builds

Molter73:
## Description

When running from PRs, the URL used for driver download gets a `cid=collector` parameter appended to it, this is done for some internal filtering. When we migrated to OSCI, this extra parameter has been added to all versions of collector, it has no functional impact on it, but we should still only add it when it is actually needed.

## Checklist

## Testing Performed
Testing will need to happen once we merge to master or by removing the `export COLLECTOR_APPEND_CID=true` line altogether.


Stringy (approved):
LGTM!

PR #895, opened 2022-11-08
Jv rox 13324 performance testing

JoukoVirtanen:
## Description

The following changes to the performance tests are made.

Performance testing tailored for processes listening on ports. These performance tests start large numbers of processes that open one port each. These processes are killed randomly.

More metrics are measured, including the breakdown of cpu usage by stackrox component.

The main, and scanner images can be varied not just the collector image. Patches can also be applied so that different configs can be tested in one run.

The input is a json config file, instead of a set of parameters.

## Checklist
- [ ] Ran performance testing

## Testing Performed

See above

### DIFF
--- a/utilities/PerformanceTesting/GetAverages.py
+++ b/utilities/PerformanceTesting/GetAverages.py
@@ -58,7 +58,7 @@ def initialize_cumulative(metrics, cumulative):
             cumulative[metric_group][metric] = {}
             for key in metrics[metric_group][metric]:
                 value = metrics[metric_group][metric][key]
-                if type(value) is float or value is None:
+                if isinstance(value, (float, int)) or value is None:
                     cumulative[metric_group][metric][key] = []
                 else:
                     cumulative[metric_group][metric][key] = value
--- a/utilities/PerformanceTesting/OpenClosePortsLoad/open-close-ports-load.py
+++ b/utilities/PerformanceTesting/OpenClosePortsLoad/open-close-ports-load.py
@@ -1,0 +1,48 @@
+import argparse
+from argparse import RawTextHelpFormatter
+import os, sys
+import random
+import time
+import subprocess
+
+
+def main(num_ports, num_per_second):
+    ports = [False] * (num_ports + 1)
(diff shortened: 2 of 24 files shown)


utilities/PerformanceTesting/GetAverages.py
```diff
@@ -58,7 +58,7 @@ def initialize_cumulative(metrics, cumulative):
             cumulative[metric_group][metric] = {}
             for key in metrics[metric_group][metric]:
                 value = metrics[metric_group][metric][key]
-                if type(value) is float or value is None:
+                if type(value) is float or value is None or type(value) is int:
```
JoukoVirtanen, line 61: Use isinstance

utilities/PerformanceTesting/OpenClosePortsLoad/open-close-ports-load.py
```diff
@@ -0,0 +1,49 @@
...
+            except (subprocess.CalledProcessError, BlockingIOError) as e:
+                print(e)
+                ports[rand_port] = False
+        else:
+            try:
+                ports[rand_port] = False
+                procs[rand_port].terminate()
+            except (subprocess.CalledProcessError, BlockingIOError) as e:
+                print(e)
+                ports[rand_port] = True
```
Molter73, line 29: I think we can simplify this to a single list by setting `procs[rand_port]` to `None` after we terminate it
```suggestion
    procs = [None] * (num_ports + 1)
    while True:
        time.sleep(1.0 / float(num_per_second))
        rand_port = random.randint(1, num_ports)
        print("rand_port", rand_port)
        if procs[rand_port] is None: 
            try:
                procs[rand_port] = subprocess.Popen(["socat", "TCP-LISTEN:" + str(rand_port) + ",fork", "STDOUT"])
            except (subprocess.CalledProcessError, BlockingIOError) as e:
                print(e)
        else:
            try:
                procs[rand_port].terminate()
            except (subprocess.CalledProcessError, BlockingIOError) as e:
                print(e)
            finally:
                procs[rand_port] = None
```

PR #900, opened 2022-11-14
ROX-13249 plop: use falco as process info source, using a workqueue

ovalenti:
## Description

The current implementation for plop reads the content of /proc to retrieve information about processes listening on the network. A caveat was hit when dealing with processes modifying their environment, which is reflected in the content of /proc.
In order to circumvent this issue, and to minimize the chances have divergent information from process signals, we propose to get process information from Falco. Indeed, when a process is created while Falco is active, the information is retrieved from the kernel fork/exec event, which cannot be modified by the process.

An attempt to address these issues was made in https://github.com/stackrox/collector/pull/882, but made use of a general lock. This PR uses an asynchronous approach.

## Checklist

**Automated testing**
  - [x] Updated unit tests
  - [x] Enabled integration tests


collector/lib/ProcfsScraper.cpp
```diff
@@ -383,7 +384,12 @@ void ResolveSocketInodes(const SocketsByContainer& sockets_by_container, const C
         } else if (listen_endpoints) {
           if (const auto* ep = Lookup(ns_network_data->listen_endpoints, socket_inode)) {
             if (!IsRelevantEndpoint(ep->endpoint)) continue;
-            listen_endpoints->emplace_back(container_id, ep->endpoint, ep->l4proto, ep->process);
+
+            std::shared_ptr<Process> process;
+            if (process_store)
+              process = process_store->Fetch(ep->pid);
```
Molter73, line 390: [nit] Put braces around this or move it to the same line (braces preferred).
```suggestion
            if (process_store) {
              process = process_store->Fetch(ep->pid);
            }
```

collector/lib/ProcfsScraper.cpp
```diff
@@ -407,7 +413,7 @@ bool ReadContainerConnections(const char* proc_path, ProcessStore* process_store
   // Read all the information from proc.
   while (auto curr = procdir.read()) {
     if (!std::isdigit(curr->d_name[0])) continue;  // only look for <pid> entries
-    int pid = stoi(curr->d_name);
+    int pid = std::strtoll(curr->d_name, 0, 10);
```
Molter73, line 416: You are using strtoll but only using an int to store it, since the original code used atoi I would recommend using strtol instead, if the bigger integers are indeed needed change int for auto or long long.

collector/lib/ProcfsScraper.cpp
```diff
@@ -540,32 +542,25 @@ bool ConnScraper::Scrape(std::vector<Connection>* connections, std::vector<Conta
...
   }
 
-  return Process(
-      std::move(container_id),
-      std::move(comm),
-      std::move(exe),
-      std::move(exe_path),
-      std::move(args),
-      pid);
+  return success_status;
```
Molter73, line 563: It's more code, but I don't see the point in calling `ReadProcessExe` and `ReadProcessCmdline` if `GetContainerID` fails.
```cpp
if (!dirfd.valid()) {
  return false;
}

if (!GetContainerID(dirfd, &pi.container_id)) {
  return false;
}

if (!ReadProcessExe(process_path, dirfd, pi.comm, pi.exe_path)) {
  return false;
}

if (!ReadProcessCmdline(process_path, dirfd, pi.exe, pi.args)) {
  return false;
}

return true;
```

Alternatively, I guess the compiler should be smart enough to properly optimize something like this:
```cpp
if (!dirfd.valid()) {
  return false;
}

return GetContainerID(dirfd, &pi.container_id) &&
    ReadProcessExe(process_path, dirfd, pi.comm, pi.exe_path) &&
    ReadProcessCmdline(process_path, dirfd, pi.exe, pi.args);
```

collector/lib/SysdigService.h
```diff
@@ -89,6 +93,7 @@ class SysdigService : public Sysdig {
 
   void AddSignalHandler(std::unique_ptr<SignalHandler> signal_handler);
 
+  mutable std::mutex libsinsp_mutex_;
```
Molter73, line 96: Do we still need this mutex? I thought the point of using the workqueue is that we could directly access data from the scraper without blocking the main inpector thread.

Molter73 (commented):
I went through this and #882, overall I like this implementation better, even if it is slightly more complicated. I've left a few comments with suggestions and some questions.

collector/lib/Process.cpp
```diff
@@ -23,29 +23,124 @@ You should have received a copy of the GNU General Public License along with thi
...
 
-  std::shared_ptr<CachedProcess> cached_process = std::make_shared<CachedProcess>(process_source_->ByPID(pid), *this);
+  std::shared_ptr<Process> cached_process = std::make_shared<Process>(pid, cache_, falco_instance_);
 
-  cache_.emplace(cached_process->pid(), cached_process);
+  cache_->emplace(pid, cached_process);
   return cached_process;
 }
 
+const std::string Process::NOT_AVAILABLE("N/A");
```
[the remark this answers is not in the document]
ovalenti: I moved this definition to the top of the namespace as suggested

collector/lib/Process.cpp
```diff
@@ -23,29 +23,124 @@ You should have received a copy of the GNU General Public License along with thi
...
+    std::ostringstream args;
+    for (auto it = falco_threadinfo_->m_args.begin(); it != falco_threadinfo_->m_args.end();) {
+      args << *it++;
+      if (it != falco_threadinfo_->m_args.end()) args << " ";
+    }
+    return args.str();
+  }
+
+  return NOT_AVAILABLE;
+}
```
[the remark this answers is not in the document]
ovalenti: Agreed, I used this layout

collector/lib/Process.h
```diff
@@ -24,88 +24,84 @@ You should have received a copy of the GNU General Public License along with thi
...
+
+  /* Get a Process by PID.
+     Returns a reference to the cached Process entry, which may have just been created
+     if it wasn't already known. */
+  const std::shared_ptr<Process> Fetch(uint64_t pid);
+
+  typedef std::shared_ptr<std::unordered_map<uint64_t, std::weak_ptr<Process>>> MapRef;
+
+ private:
+  SysdigService* falco_instance_;
```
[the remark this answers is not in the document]
ovalenti: falco_instance_ is hosted as a value in the CollectorService; and I was reluctant to turn it into a shared_ptr. This particular reference is used only when creating a new Process object, which occurs when scraping. In order for the reference to be used dangling, the NetworkStatusNotifier would have to be running when CollectorService has been destroyed, which should never happen.

collector/lib/SysdigService.cpp
```diff
@@ -315,4 +325,27 @@ void SysdigService::AddSignalHandler(std::unique_ptr<SignalHandler> signal_handl
...
+void SysdigService::ServePendendingProcessRequests() {
+  std::lock_guard<std::mutex> lock(process_requests_mutex_);
+
+  while (!pending_process_requests_.empty()) {
+    auto& request = pending_process_requests_.front();
+    uint64_t pid = request.first;
+    auto callback = request.second.lock();
+
+    if (callback)
+      callback->operator()(inspector_->get_thread_ref(pid, true));
```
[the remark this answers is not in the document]
ovalenti: .get() is an expression, which cannot be used to invoke an operator. However, `(*callback)(...)` is possible. I did that.

collector/lib/ProcfsScraper.cpp
ovalenti: Changed pid to `long long`

collector/lib/ProcfsScraper.cpp
ovalenti: Changed to make use of the alternative proposal :+1:

collector/lib/SysdigService.cpp
```diff
@@ -315,4 +325,27 @@ void SysdigService::AddSignalHandler(std::unique_ptr<SignalHandler> signal_handl
...
 
+void SysdigService::GetProcessInformation(
+    uint64_t pid,
+    std::weak_ptr<std::function<void(threadinfo_map_t::ptr_t)>> callback) {
+  std::lock_guard<std::mutex> lock(process_requests_mutex_);
+
+  pending_process_requests_.emplace_back(pid, callback);
+}
+
+void SysdigService::ServePendendingProcessRequests() {
```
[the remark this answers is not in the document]
ovalenti: That's correct, thanks !

Molter73 (approved):
/lgtm

collector/lib/SysdigService.h
ovalenti: The purpose of this mutex is to avoid executing some Falco concurrently; which was very likely to happen with the previous approach, and a lot less now.
It still guards against executing CleanUp() (`inspector_close()`) while the Falco thread is inside GetNext()  (`inspector_.next()`) for instance.
However, this situation will likely never happen, while the mutex is in the hot path....

ovalenti: A mutex that is not subject to a lot of contention does not cost much. So I would say 'yes'.

collector/lib/SysdigService.cpp
```diff
@@ -262,10 +270,23 @@ void SysdigService::CleanUp() {
...
+
+    if (callback) (*callback)(0);
+
+    pending_process_requests_.pop_front();
+  }
 }
 
 bool SysdigService::GetStats(SysdigStats* stats) const {
-  std::lock_guard<std::mutex> lock(running_mutex_);
+  std::lock_guard<std::mutex> sysdig_lock(libsinsp_mutex_);
```
robbycochran, line 288: Nit: rename occurances of `sysdig_lock` to `libsinsp_lock`.

collector/lib/Process.cpp
```diff
@@ -23,29 +23,127 @@ You should have received a copy of the GNU General Public License along with thi
...
+  falco_threadinfo_ = process_info;
+  process_info_resolved_ = true;
+
+  process_info_condition_.notify_all();
+}
+
+void Process::WaitForProcessInfo() const {
+  std::unique_lock<std::mutex> lock(process_info_mutex_);
+
+  if (!process_info_resolved_) {
```
robbycochran, line 138: Can you add a count metrics to this function to see `WaitForProcessInfo` hits and misses?

collector/lib/Process.cpp
```diff
@@ -23,29 +23,143 @@ You should have received a copy of the GNU General Public License along with thi
...
 #include "Process.h"
 
+#include <chrono>
+
+#include "CollectorStats.h"
+#include "SysdigService.h"
+
 namespace collector {
 
+const std::string Process::NOT_AVAILABLE("N/A");
```
JoukoVirtanen, line 33: I think something longer and more descriptive might be better. E.g "Process could not be determined by Collector"

collector/lib/ProcfsScraper.cpp
```diff
@@ -565,32 +566,22 @@ bool ConnScraper::Scrape(std::vector<Connection>* connections, std::vector<Conta
   return ReadContainerConnections(proc_path_.c_str(), process_store_, connections, listen_endpoints);
 }
 
-Process ProcessScraper::ByPID(uint64_t pid) {
+bool ProcessScraper::Scrape(uint64_t pid, ProcessInfo& pi) {
```
JoukoVirtanen, line 569: pi makes me think of 3.1415... Maybe change to processInfo.

collector/lib/SysdigService.h
```diff
@@ -101,6 +106,11 @@ class SysdigService : public Sysdig {
   mutable std::mutex running_mutex_;
   bool running_ = false;
   bool useEbpf_ = false;
+
+  void ServePendingProcessRequests();
+  mutable std::mutex process_requests_mutex_;
+  // [ ( pid, callback ), ( pid, callback ), ... ]
+  std::list<std::pair<uint64_t, std::weak_ptr<std::function<void(threadinfo_map_t::ptr_t)>>>> pending_process_requests_;
```
JoukoVirtanen, line 113: Can a typedef be used here to simplify this?

collector/test/ConnTrackerTest.cpp
```diff
@@ -1246,122 +1246,6 @@ TEST(ConnTrackerTest, TestComputeDeltaAfterglowBenchmark2) {
...
-
-  ContainerEndpointMap old_state = {{ce1, oldConnStatus}};
-  ContainerEndpointMap new_state = {{ce2, newConnStatus}};
-
-  ContainerEndpointMap expected_delta = {{ce1, ConnStatus(connection_time1, false)}, {ce2, newConnStatus}};
-
-  CT::ComputeDelta(new_state, &old_state);
-  EXPECT_THAT(old_state, expected_delta);
-}
-
```
JoukoVirtanen, line 1364: Why remove all of these tests. If they are not working anymore maybe just modify them.

integration-tests/suites/listening_ports.go
```diff
@@ -96,11 +96,7 @@ func (s *ProcessListeningOnPortTestSuite) TestProcessListeningOnPort() {
 	for _, endpoint := range endpoints {
 		assert.Equal(s.T(), "L4_PROTOCOL_TCP", endpoint.Protocol)
 
-		// TODO
-		// There is a difference in the lengths of the process name
-		// The process stream truncates to 16 characters while endpoints does not.
-		// We need to decide which way to go and make these tests consistent with that.
-		//assert.Equal(s.T(), endpoint.Originator.ProcessName, process.Name)
+		assert.Equal(s.T(), endpoint.Originator.ProcessName, process.Name)
```
JoukoVirtanen, line 99: I am happy to see these asserts being enabled.

JoukoVirtanen:
Looks good. I just had a few minor comments. I don't want to delay this from being merged.

JoukoVirtanen (approved):

collector/lib/ProcfsScraper.cpp
ovalenti: :heavy_check_mark:  Renamed to something more explicit.

collector/lib/SysdigService.h
ovalenti: It is not reused much, but 4 levels is a lot, I agree. I introduced a typedef for the callback reference.

collector/test/ConnTrackerTest.cpp
ovalenti: Agreed, I modified Process so that it can now be constructed without a reference to the SysdigService or the ProcessStore. The tests are restored almost as-is.

PR #902, opened 2022-11-15
rox-13539 collector integration tests for an executable with symbolic link

JoukoVirtanen:
## Description

Adds a test to check that when a process is launched using a symbolic link that the process information and the process information in the endpoint stream match.

## Checklist

## Testing Performed

CI and ran tests locally.

### DIFF
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -821,6 +821,27 @@ func (s *ProcessListeningOnPortTestSuite) TestProcessListeningOnPort() {
 			endpoints8081 = append(endpoints8081, endpoint)
 		} else {
 			endpoints9091 = append(endpoints9091, endpoint)
+		}
+		// other ports cannot exist at this point due to previous assertions
+	}
+
+	// This helper simplifies the assertions a fair bit, by checking that
+	// the recorded endpoints have an open event (CloseTimestamp == nil) and
+	// a close event (CloseTimestamp != nil) and not two close events or two open
+	// events.
+	//
+	// It is also agnostic to the order in which the events are reported.
+	hasOpenAndClose := func(infos []EndpointInfo) bool {
+		if !assert.Len(s.T(), infos, 2) {
+			return false
+		}
+		return infos[0].CloseTimestamp != infos[1].CloseTimestamp &&
+			(infos[0].CloseTimestamp == nilTimestamp || infos[1].CloseTimestamp == nilTimestamp)
+	}
+
+	assert.True(s.T(), hasOpenAndClose(endpoints8081), "Did not capture open and close events for port 8081")
+	assert.True(s.T(), hasOpenAndClose(endpoints9091), "Did not capture open and close events for port 9091")
+}
 
 func (s *SymbolicLinkProcessTestSuite) SetupSuite() {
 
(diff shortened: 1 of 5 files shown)


integration-tests/integration_test.go
```diff
@@ -768,24 +778,38 @@ func (s *ProcessListeningOnPortTestSuite) TestProcessListeningOnPort() {
 	endpoints, err := s.GetEndpoints(s.serverContainer)
 	s.Require().NoError(err)
 
-	if !assert.Equal(s.T(), 4, len(endpoints)) {
+	if !assert.Equal(s.T(), 5, len(endpoints)) {
```
Stringy, line 781: The changes in this test are quite extensive and hide a lot of additional test cases amongst the existing test. I think it would be a bit cleaner to separate these out into their own test specifically for the symlink case

JoukoVirtanen: Done.

integration-tests/integration_test.go
```diff
@@ -834,6 +843,90 @@ func (s *ProcessListeningOnPortTestSuite) TestProcessListeningOnPort() {
...
+	processImage := qaImage("quay.io/rhacs-eng/qa", "collector-processes-listening-on-ports")
+
+	containerID, err := s.launchContainer("process-ports", "-v", "/tmp:/tmp", processImage)
+
+	s.Require().NoError(err)
+	s.serverContainer = containerShortID(containerID)
+
+	actionFile := "/tmp/action_file_ln.txt"
+
+	_, err = s.collector.executor.Exec("sh", "-c", "rm "+actionFile+" || true")
```
Stringy, line 872: [nit] I think the `sh -c` part is redundant based on the way the executor works (in that gcloud ssh and ssh will run commands via sh -c themselves)

integration-tests/integration_test.go
```diff
@@ -834,6 +843,90 @@ func (s *ProcessListeningOnPortTestSuite) TestProcessListeningOnPort() {
...
+
+	_, err = s.execContainer("process-ports", []string{"ln", "-s", "/process-listening-on-ports", "/plop"})
```
Stringy, line 874: do we need to use an absolute path for `ln` like we do with bash on 877?

Stringy: Also, can't we perform this step during the container build? If we set the container up with the necessary symlinks we can just change the entrypoint when we start the container and remove a fair bit of additional complexity from this test setup

integration-tests/integration_test.go
Stringy: Does the action file even exist at this point?

JoukoVirtanen: I think `sh -c` is needed when running the tests locally. Also when running locally there is a chance that the action file exists. We should be defensive and not make assumptions.

Stringy: Then this indicates a shortcoming in the local executor. We have executor functions for copying files, perhaps we could add one for writing files as well? We are making an assumption here that the executor is not running the command in a shell, which is false in two thirds of the scenarios.

integration-tests/integration_test.go
JoukoVirtanen: Done

integration-tests/integration_test.go
Stringy: We shouldn't ignore this, but happy for it to be completed in a follow up

Stringy (approved):

PR #903, opened 2022-11-16
Exclude broken Debian bundles from driver builds

Molter73:
## Description

The drivers added to `BLOCKLIST` are currently failing due to a known problem with kernel-packer that breaks the bundles.

## Checklist

## Testing Performed

- [x] Green check for driver builds


Molter73:
Failing test is a timeout, should be unrelated to these changes.

Stringy (approved):

PR #904, opened 2022-11-16
Organise Integration Tests

Stringy:
## Description

To simplify maintenance and on-going addition of tests, this PR organises the tests and common functionality into specific files, cleaning up the directory and making sure that each test is separated from the rest.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


integration-tests/suites/base.go
```diff
@@ -0,0 +1,416 @@
+package suites
```
Molter73, line 1: I would argue that this file should actually live under `suites/common/`, since most functions and methods seem to be quite generic. Alternatively, it might be worth it at some point to go over individual functions and split them into more specific files, just by going through the file I see potential splits into:
- Stats container
- Performance analysis
- Container handling
- Various assertions & DB access

This can be done in follow up though, no point on stop merging this PR on a nit.

integration-tests/suites/procfs_scraper.go
```diff
@@ -0,0 +1,105 @@
...
+	err := s.executor.PullImage(image)
+	s.Require().NoError(err)
+
+	// invokes default nginx
+	containerID, err := s.launchContainer("nginx", image)
+	s.Require().NoError(err)
+	s.ServerContainer = common.ContainerShortID(containerID)
+
+	time.Sleep(10 * time.Second)
+}
```
Molter73, line 66: This can probably be turned into a more generic function, I just saw `process_network.go` also launches nginx containers.

integration-tests/suites/base.go
Stringy: Absolutely agree that a lot of this can and should be split out into their own structures, and I think the base structure could then pull some of those things together to keep it all nicely separated

integration-tests/suites/procfs_scraper.go
Stringy: Agreed, will look to implement in a follow up to consolidate a lot of the generic functionality throughout the tests

PR #910, opened 2022-11-21
Adds Ansible playbook for benchmarking

Stringy:
## Description

Adds an Ansible playbook to run the benchmarks against a pre-defined set of VMs. 

Includes a fairly significant restructure of the Ansible roles and playbooks to make them more generic and easier to reuse. In particular, VMs can now be designated for specific collection methods, and VM lifecycle has its own playbook for slightly easier use during general development / testing (i.e. outside of CI)

Duplicate of #909 due to branch name change to allow for OSCI changes

## Checklist

**Automated testing**
~- [ ] Added unit tests~

## Testing Performed

Tested extensively locally (i.e. on GCP VMs)

OSCI integration not yet included for the benchmarks, but integration test changes will be tested in CI.


.openshift-ci/jobs/integration-tests/run-benchmarks.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+BRANCH="$(jq -r '.extra_refs[0].base_ref' <(echo "$JOB_SPEC"))"
```
Molter73, line 5: Can we use the `get_branch` function from `libs.sh` instead? https://github.com/stackrox/collector/blob/42fbb50c7f03d37bfd37ff452e2d75ca9fcca79a/.openshift-ci/scripts/lib.sh#L82

.openshift-ci/jobs/integration-tests/run-benchmarks.sh
```diff
@@ -0,0 +1,44 @@
...
+function save_artifacts() {
+    cp "integration-tests/perf.json" "${ARTIFACT_DIR}/${JOB_NAME_SAFE}-perf.json"
+    cp -r "integration-tests/container-logs/" "${ARTIFACT_DIR}"
+
+    if [[ -d "integration-tests/performance-logs" ]]; then
+        cp -r "integration-tests/performance-logs" "${ARITFACT_DIR}"
+    fi
+}
+
+run_benchmarks || exit_code=$?
```
Molter73, line 20: Define `exit_code` as 0 just to be double safe it is defined if the benchmark works correctly.
```suggestion
exit_code=0
run_benchmarks || exit_code=$?
```

.openshift-ci/jobs/integration-tests/run-benchmarks.sh
```diff
@@ -0,0 +1,44 @@
...
+save_artifacts
+
+if [ "${exit_code}" -ne 0 ]; then
```
Molter73, line 23: Since exit_code will be a number, think you can do this instead:
```suggestion
if ((exit_code)); then
```

ansible/roles/run-test-target/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
+---
+- name: Log into quay.io
+  shell:
+    cmd: "docker login -u {{ quay_username }} --password-stdin quay.io"
+    stdin: "{{ quay_password }}"
```
Molter73, line 5: Out of curiosity, have we considered using the [docker_login module](https://docs.ansible.com/ansible/latest/collections/community/docker/docker_login_module.html#ansible-collections-community-docker-docker-login-module) instead of running `docker login` in a shell?

Molter73 (commented):
It'll be great to have our benchmarking tests back!

ansible/roles/run-test-target/tasks/main.yml
Stringy: That is an excellent idea, I'll see what I can do.

Stringy: As discussed offline this is painful to support due to python version differences (particularly on RHEL 7) so we'll keep it as it is for now.

I'll probably figure something out and fix in a follow up PR

.openshift-ci/scripts/baseline/main.py
```diff
@@ -0,0 +1,335 @@
...
+    baseline_keys = [items[0] for items in baseline_filtered]
+    new_filtered = [
+        data for data in new_grouped
+        if data[0] in baseline_keys
+    ]
+
+    return (baseline_filtered, new_filtered)
+
+
+def normalize_collection_method(method):
```
Stringy, line 192: I lied above. As it turns out I also needed to add this function to normalise the collection methods and some changes in `process` below to normalise VmConfig and collection method

.openshift-ci/jobs/integration-tests/Dockerfile
```diff
@@ -23,3 +23,7 @@ RUN mkdir -p "$GOCACHE" && \
 # to work around this, we have to create the directory
 # first.
 RUN mkdir /.ssh && chmod g+rwx /.ssh
+
+# Install hub-comment for commenting on PRs
+RUN wget --quiet https://github.com/joshdk/hub-comment/releases/download/0.1.0-rc6/hub-comment_linux_amd64 && \
+    install hub-comment_linux_amd64 /usr/bin/hub-comment
```
Molter73, line 29: This image is built on top of `quay.io/stackrox-io/apollo-ci:collector`, that image should already have `hub-comment` installed in it: https://github.com/stackrox/rox-ci-image/blob/60cb33a19eb42b7f425eb257dada42096dd75d5c/images/collector.Dockerfile#L57-L59

Makefile
```diff
@@ -17,6 +17,10 @@ dev-build: image integration-tests-process-network
 tag:
 	@echo "$(COLLECTOR_TAG)"
 
+.PHONY: builder-tag
+builder-tag:
+	@echo "$(COLLECTOR_BUILDER_TAG)"
```
Molter73, line 22: Do we really need a make target for this?

Molter73 (approved):
/lgtm

Approving changes, provided CI passes. The comments I left are minor and can be ignored if they are an annoyance.

.openshift-ci/jobs/integration-tests/Dockerfile
Stringy: That's what I get for just copying from scripts and not checking the base image! I'll remove this

Makefile
Stringy: the thinking here is to keep the builder tag logic in one place rather than hard coding it in the playbooks or in the OSCI scripts

PR #913, opened 2022-11-25
Simplifies Makefile targets for testing

Stringy:
## Description

Removes test-specific targets in favour of auto-generated targets based on output of `go test -list` to make the system a little more scalable to large numbers of tests.

## Checklist


ovalenti (approved):
I believe that this is a great improvement, thanks !

PR #914, opened 2022-11-25
Add latest file to list of support packages

mtodor:
## Description

This PR just adds a new file: `latest` - to the list of built support packages.
The `latest` file will contain the last build for that kernel module version.

Related to the conversation in Slack: https://srox.slack.com/archives/CFMQ5C2TT/p1669284268331969

**Open question**

- I'm not sure if this file will be properly synced to Google Storage.

## Checklist

**Automated testing**
  - ~~[ ] Added unit tests~~

## Testing Performed

I didn't test it. I just made a conclusion from the code about what should be changed. 😞

### DIFF
--- a/kernel-modules/support-packages/04-create-support-packages.sh
+++ b/kernel-modules/support-packages/04-create-support-packages.sh
@@ -60,6 +60,9 @@ for mod_ver_dir in "${MD_DIR}/module-versions"/*; do
     compress_files "$package_root" "${package_out_dir}/${filename}"
     generate_checksum "${package_out_dir}" "${filename}"
 
+    # Export the latest build filename
+    echo "${filename}" > "${package_out_dir}/latest"
+
     cp "${package_out_dir}/${filename}" "${package_out_dir}/${latest_filename}"
     generate_checksum "${package_out_dir}" "${latest_filename}"
     rm -rf "$package_root" || true


Molter73 (approved):
/lgtm

kernel-modules/support-packages/04-create-support-packages.sh
```diff
@@ -60,6 +60,9 @@ for mod_ver_dir in "${MD_DIR}/module-versions"/*; do
     compress_files "$package_root" "${package_out_dir}/${filename}"
     generate_checksum "${package_out_dir}" "${filename}"
 
+    # Export the latest build filename
+    echo "${filename}" > "${package_out_dir}/latest"
```
Molter73, line 64: I think this might actually be a little more useful, since support packages are split by mod version and it would make it easier when creating the download URL
```suggestion
    echo "${mod_ver}/${filename}" > "${package_out_dir}/latest"
```

mtodor: I think it's not needed since `package_out_dir="${OUT_DIR}/${mod_ver}"` <- already contains the mod version. We need the mod version anyway to fetch the `latest` file. And it's easy to concatenate it to generate the full URL.

PR #916, opened 2022-11-29
Driver builds on GHA

Molter73:
## Description

This PR adds a GHA workflow dedicated to building kernel drivers. It should be on par with the process of building drivers that OSCI provides now, including PR labels support and build failure checks.

Follow up work:
- Add support package builds.
- Integrate with the collector image build (either existing OSCI or a new workflow on GHA)

## Checklist

## Testing Performed

- [x] Run a full build of a collector driver version with no cache.
- [x] Run a full build of a collector driver version with cache.
- [x] Try and build a full collector image with the GHA built drivers and run integration tests with it manually.


Stringy (approved):
Couple of very minor comments but otherwise looks great!

.github/workflows/drivers.yml
```diff
@@ -0,0 +1,233 @@
...
+          name: built-drivers
+          path: /tmp/output/
+
+      - name: Choose destination bucket
+        id: gcp-bucket
+        run: |
+          if [[ "${{ github.event_name == 'pull_request' }}" == "true" ]]; then
+            echo "gcp-bucket=mauro-drivers-test/pr-builds/${HEAD_REF}/${{ github.run_id }}" >> "$GITHUB_OUTPUT"
+          else
+            echo "gcp-bucket=mauro-drivers-test/master" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
Molter73: Oh yeah, these are just for testing it out, eventually we'll want to start pushing to the buckets OSCI is now pushing to. I'll add the TODO

tommartensen, line 220: You're using the workflow from `main`, which will be triggered for `master` & `release-*` branches. Consider this for destination bucket on branch builds: 
```suggestion
            echo "gcp-bucket=mauro-drivers-test/${{ github.ref_name }}" >> "$GITHUB_OUTPUT"
```

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

tommartensen (commented):
Great to see you happy with GitHub actions! One tiny remark from me, only reviewed the workflows to the best of my understanding of collector.

.github/workflows/drivers.yml
Molter73: This is actually just for testing purposes, once we move to using only GHA the bucket for the master branch is completely different from the one used for staging. I just used a `master` subdirectory because I'm too lazy to go setup 2 separate buckets 😅

Molter73: Just realized, we are fine having tag branches push drivers to the same bucket master pushes too, but we'll probably make it so tags don't build drivers at all, since they should be using the ones built on master anyways.

Stringy (approved):
Reapproving following the mock cache changes.

Also just to note that we will need to figure out the `main.yml` workflow for the whole pipeline including the integration tests and image builds.

PR #917, opened 2022-11-29
integration tests on GHA

Stringy:
## Description

Adds an integration-tests workflow for GitHub Actions.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,45 @@
+name: Collector Integration Tests
+
+on: 
+  pull_request:
+    types: [opened, reopened, synchronize, labeled]
```
Molter73, line 5: Wouldn't `labeled` make it so this runs whenever we add or remove a label? That seems a little overkill.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+        vm_type:
+          - rhel
+          - ubuntu-os
+    with:
+      vm_type: ${{ matrix.vm_type }}
+    secrets: inherit
+
+  all-integration-tests:
+    uses: ./.github/workflows/integration-tests-vm-type.yml
+    if: contains(github.event.pull_request.labels.*.name, 'all-integration-tests')
```
Molter73, line 30: This we might want to run these on master/release too
```suggestion
    if: contains(github.event.pull_request.labels.*.name, 'all-integration-tests') || github.event_name == 'push'
```

.pre-commit-config.yaml
```diff
@@ -16,3 +16,8 @@ repos:
     rev: "6230247"
     hooks:
       - id: clang-format
+
+  - repo: git@github.com:rhysd/actionlint
```
Molter73, line 20: Nice!

ansible/ci/group_vars/all.yml
```diff
@@ -2,7 +2,7 @@
 
 gcp_project: "stackrox-ci"
 gcp_auth_kind: "serviceaccount"
-gcp_service_account_file: /tmp/secret/stackrox-collector-e2e-tests/GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT
+gcp_service_account_file: /tmp/secret/gcp-account.json
```
Molter73, line 5: Could this change interfere with OSCI?

.github/workflows/integration-tests-vm-type.yml
```diff
@@ -0,0 +1,84 @@
...
+        run: make -C "${{ github.workspace }}/ansible" destroy-vms
+
+      - name: Store artifacts
+        if: always()
+        uses: actions/upload-artifact@v3
+        with:
+          path: |
+            ${{ github.workspace }}/integration-tests/container-logs/**/*
+            ${{ github.workspace }}/integration-tests/perf.json
+            ${{ github.workspace }}/integration-tests/performance-logs/**/*
```
Molter73, line 83: I would also add a name to this artifact, so it doesn't get uploaded as the generic `artifact` name and potentially grouped with other stuff. Maybe something like this?
```yaml
name: ${{ inputs.vm_type }}-logs
```

.github/workflows/integration-tests.yml
Stringy: it would but without it there's seemingly no other way of re-running when a label is added (you could feasibly add a label and then push something to the PR to trigger it, but that doesn't feel like behaviour we want to encourage)

ansible/ci/group_vars/all.yml
Stringy: For now I'll copy the creds to the same place as OSCI would, to maintain compatibility

Molter73 (approved):
/lgtm

.github/workflows/integration-tests-vm-type.yml
```diff
@@ -0,0 +1,118 @@
...
+
+      - name: Normalize job ID
+        run: |
+          NORM_JOB_ID="${JOB_ID/-/}"
+          echo "JOB_ID=${NORM_JOB_ID}" | tee -a "$GITHUB_ENV"
+
+      - name: Construct COLLECTOR_TAG
+        run: |
+          COLLECTOR_TAG="$(make tag)-gha-slim"
+          echo "COLLECTOR_TAG=${COLLECTOR_TAG}" | tee -a "$GITHUB_ENV"
```
Molter73, line 92: We can do this in a follow up, but I was thinking we can set the collector tag as an output on a previous job, then simply use it here as an input, WDYT? (see #925 for an example on what that might look like).

.github/workflows/collector-image.yml
```diff
@@ -1,13 +1,14 @@
...
-  pull_request:
-    branches:
-      - master
+  #  push:
+  #    branches:
+  #      - master
+  #      - release-*
+  #  pull_request:
+  #    branches:
+  #      - master
```
Molter73, line 10: I think we can safely remove these, we probably only want the `main.yml` to be called on pushes and PRs.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,46 @@
...
+
+on: 
+  #  pull_request:
+  #    types: [opened, reopened, synchronize, labeled]
+  #    branches:
+  #      - master
+  #  push:
+  #    branches:
+  #      - master
+  #      - release-*
```
Molter73, line 11: Same as previous comment about removing these triggers.

.github/workflows/main.yml
```diff
@@ -0,0 +1,21 @@
...
+    types: [opened, reopened, synchronize, labeled]
+
+jobs:
+  build-collector:
+    uses: ./.github/workflows/collector-image.yml
+    secrets: inherit
+
+  integration-tests:
+    uses: ./.github/workflows/integration-tests.yml
+    needs: [build-collector]
```
Molter73, line 20: [nit] When the requirement is a single job, we don't need to use an array
```suggestion
    needs: build-collector
```

.github/workflows/integration-tests.yml
Molter73: But then you could add and remove a bunch of labels and it will trigger just as many jobs, feels like a great way to get DoS'ed.

Molter73 (approved):
/lgtm

.github/workflows/integration-tests-vm-type.yml
Stringy: yeah I think that's a good shout, it'll make it easier to unify versions/naming conventions

.github/workflows/main.yml
```diff
@@ -0,0 +1,20 @@
+name: Collector CI
+
+on:
+  push:
+    branches:
+      - master
+      - release-*
+  pull_request:
+    branches:
+      - master
```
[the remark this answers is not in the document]
Stringy: I've removed this limitation now, but we'll have to keep an eye on resource usage & parallel jobs because with the integration tests this could get reasonably large. 

With some discussion with @Molter73 we think we can reduce the integration tests down to a single job by combining VM creation in Ansible and running those over just two jobs rather than the current eight. I'll tackle this in a follow up PR if it becomes a problem

PR #919, opened 2022-11-30
Keep dockerfiles up to date

janisz:
## Description

Monitor docker images and update them with @dependabot

Refs: https://github.com/dependabot/feedback/issues/145

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -6,15 +6,7 @@
 version: 2
 updates:
   - package-ecosystem: 'docker'
-    directory: .openshift-ci/drivers/
-  - package-ecosystem: 'docker'
-    directory: .openshift-ci/jobs/apollo-src/
-  - package-ecosystem: 'docker'
-    directory: .openshift-ci/jobs/integration-tests/
-  - package-ecosystem: 'docker'
     directory: collector/connscrape-test/
-  - package-ecosystem: 'docker'
-    directory: collector/container/
   - package-ecosystem: 'docker'
     directory: collector/container/
   - package-ecosystem: 'docker'
@@ -29,10 +21,6 @@ updates:
     directory: integration-tests/container/processes-listening-on-ports/
   - package-ecosystem: 'docker'
     directory: integration-tests/container/schedule-curls/
-  - package-ecosystem: 'docker'
-    directory: kernel-modules/container/
-  - package-ecosystem: 'docker'
-    directory: kernel-modules/dockerized/
   - package-ecosystem: 'docker'
     directory: oss-public-server/container/
   - package-ecosystem: 'docker'
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -7,21 +7,41 @@ version: 2
 updates:
   - package-ecosystem: 'docker'
     directory: collector/connscrape-test/
+    schedule:
+      interval: "weekly"
   - package-ecosystem: 'docker'
     directory: collector/container/
+    schedule:
+      interval: "weekly"
   - package-ecosystem: 'docker'
     directory: integration-tests/container/benchmark/
+    schedule:
+      interval: "weekly"
   - package-ecosystem: 'docker'
     directory: integration-tests/container/containerstats/
+    schedule:
+      interval: "weekly"
   - package-ecosystem: 'docker'
     directory: integration-tests/container/jsonlabel/
+    schedule:
+      interval: "weekly"
   - package-ecosystem: 'docker'
     directory: integration-tests/container/perf/
+    schedule:
+      interval: "weekly"
   - package-ecosystem: 'docker'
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -0,0 +1,39 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: 'docker'
+    directory: .openshift-ci/drivers/
```
Molter73, line 9: We don't want the driver images to be auto-updated, that can lead to driver errors and would be pretty bad for us. Please remove these.

.github/dependabot.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - package-ecosystem: 'docker'
+    directory: .openshift-ci/jobs/apollo-src/
+  - package-ecosystem: 'docker'
+    directory: .openshift-ci/jobs/integration-tests/
```
Molter73, line 13: The apollo-ci images on these dockerfiles need to be manually mirrored into OSCI, so auto-updating these are kind of pointless on their own (and we are also moving to use GHA soon, so the .openshift-ci will cease to exist soon).

.github/dependabot.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - package-ecosystem: 'docker'
+    directory: collector/connscrape-test/
+  - package-ecosystem: 'docker'
+    directory: collector/container/
```
Molter73, line 17: We have 2 dockerfiles here:

    Dockerfile.template uses ARGs to set the image being used as base and I'm not sure dependabot will pick up on it.
    Dockefile.ubi is not being used right now, so updating it would not really have any effect.

.github/dependabot.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - package-ecosystem: 'docker'
+    directory: collector/container/
```
Molter73, line 19: These 2 are duplicates.

.github/dependabot.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - package-ecosystem: 'docker'
+    directory: integration-tests/container/jsonlabel/
+  - package-ecosystem: 'docker'
+    directory: integration-tests/container/perf/
+  - package-ecosystem: 'docker'
+    directory: integration-tests/container/processes-listening-on-ports/
+  - package-ecosystem: 'docker'
+    directory: integration-tests/container/schedule-curls/
+  - package-ecosystem: 'docker'
+    directory: kernel-modules/container/
```
Molter73, line 33: The Dockerfile in this directory suffers from the same ARGs as tag issue I mentioned before and it is also meant to be set by either CI or through a make target to a specific version of collector, so I don't think we need/want to auto update it.

.github/dependabot.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - package-ecosystem: 'docker'
+    directory: kernel-modules/dockerized/
```
Molter73, line 35: The dockerfiles on this directory have fallen out of use and are in a similar situation to the ones in `.openshift-ci/drivers/` (i.e. we want to have full control on when they are updated).

Molter73 (commented):
Thanks for this PR @janisz! I left a few comments on some directories we probably don't want to have dependabot auto-update images.

The ones that don't have comments are mostly for CI and internal tooling, so we could have them auto-updated but I don't think it is 100% needed (and might lead to some unexpected errors in our tests, or might not 🤷🏻‍♂️ ). @stackrox/collector-team what do you guys think? Do we want to add auto-updating CI images?

.github/dependabot.yaml
janisz: It should https://github.com/dependabot/feedback/issues/145

Molter73: That link talks about dependabot finding files with names other than `Dockerfile`, I'm more concerned about the actual content of the dockerfile: https://github.com/stackrox/collector/blob/3433fd8211380a99ac418c2c5616c5de2830b76a/collector/container/Dockerfile.template#L1-L5

janisz: I give it a test and you are right https://github.com/janisz/collector/pulls

janisz: If Dockefile.ubi is not being used right now, why not deleting it?

Molter73: Sorry for the delayed response. Basically, that image was used as a mirror to the downstream release, it fell out of use because OSCI is similar enough to downstream and we didn't bother to re-implement it. But now we are considering going to GHA, so it might make sense to go back to building this image.

janisz:
@Molter73 PTAL

PR #922, opened 2022-12-01
GHA workflow for linters

Stringy:
## Description

Very simple workflow for running linters across the codebase, using pre-commit. Notable change is a move from SSH to HTTPS for cloning the pre-commit actions as well as running the hooks only on the diff between a PR and the base branch (usually master, but it is configured for any base branch) 

this has all been successfully tested locally.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,3 +16,8 @@ repos:
     rev: "6230247"
     hooks:
       - id: clang-format
+
+  - repo: https://github.com/rhysd/actionlint.git
+    rev: "v1.6.22"
+    hooks:
+      - id: actionlint-docker
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 repos:
-  - repo: git@github.com:syntaqx/git-hooks.git
+  - repo: https://github.com/syntaqx/git-hooks.git
     rev: "v0.0.17"
     hooks:
       - id: shfmt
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -1,18 +1,23 @@
 repos:
-  - repo: git@github.com:syntaqx/git-hooks.git
+  - repo: https://github.com/syntaqx/git-hooks.git
```
Molter73, line 2: I assume these changes will not have any adverse effect as long as the pre-commit hooks we use are made publicly available, right? 

Also, did you test the changes locally? It'd be nice to have a check on the PR description if so.

.pre-commit-config.yaml
```diff
@@ -1,18 +1,23 @@
...
-  - repo: git@github.com:doublify/pre-commit-clang-format
+  - repo: https://github.com/doublify/pre-commit-clang-format.git
     rev: "6230247"
     hooks:
       - id: clang-format
+
+  - repo: https://github.com/rhysd/actionlint.git
+    rev: "v1.6.22"
+    hooks:
+      - id: actionlint-docker
```
Molter73, line 23: ❤️

Molter73 (approved):
/lgtm

Just a comment on whether the changes were tested locally.

.pre-commit-config.yaml
Stringy: Yes to both questions. I think we can probably do something with the GitHub bot token to let us use SSH here, but we're extremely unlikely to use private pre-commit hooks (even if we had some of our own I'd expect them to be public, or defined within this repo)

PR #932, opened 2022-12-07
Support packages on gha

Molter73:
## Description

This PR adds a workflow for creating and pushing support packages on GHA. The procedure is relatively straight forward, most of the process consists of calling the scripts under `kernel-modules/support-packages` in the right order and with the correct environment variables and parameters set.

## Checklist

## Testing Performed

- [x] Manually checked at least one support package is properly formatted.
- [x] Manually checked the integrity file works.
- [x] Reviewed the contents of the generated `index.html` file are correct.
- [x] The `latest` file used for releases is correct and properly pushed.

### DIFF
--- a/.github/support-packages/create-metadata-dir.sh
+++ b/.github/support-packages/create-metadata-dir.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+for version_dir in /tmp/support-packages/metadata/collector-versions/*; do
+    version="$(basename "$version_dir")"
+    git checkout "${version}"
+
+    driver_version="$(cat "${GITHUB_WORKSPACE}/kernel-modules/MODULE_VERSION")"
+    echo "${driver_version}" > "${version_dir}/MODULE_VERSION"
+done
--- a/.github/support-packages/create-metadata-dir.sh
+++ b/.github/support-packages/create-metadata-dir.sh
@@ -6,8 +6,7 @@ for version_dir in /tmp/support-packages/metadata/collector-versions/*; do
     version="$(basename "$version_dir")"
     git checkout "${version}"
 
-    driver_version="$(cat "${GITHUB_WORKSPACE}/kernel-modules/MODULE_VERSION")"
-    echo "${driver_version}" > "${version_dir}/MODULE_VERSION"
+    cp "${GITHUB_WORKSPACE}/kernel-modules/MODULE_VERSION" "${version_dir}/MODULE_VERSION"
 done
 
 git checkout "${BRANCH_NAME}"
--- a/.github/support-packages/create-metadata-dir.sh
+++ b/.github/support-packages/create-metadata-dir.sh
@@ -9,3 +9,5 @@ for version_dir in /tmp/support-packages/metadata/collector-versions/*; do
     driver_version="$(cat "${GITHUB_WORKSPACE}/kernel-modules/MODULE_VERSION")"
     echo "${driver_version}" > "${version_dir}/MODULE_VERSION"
 done
+
+git checkout "${BRANCH_NAME}"
--- a/.github/workflows/support-packages.yml
+++ b/.github/workflows/support-packages.yml
@@ -60,7 +60,7 @@ jobs:
         id: output-bucket
         run: |
           if [[ "${{ github.event_name == 'pull_request' }}" == "true" ]]; then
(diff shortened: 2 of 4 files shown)


.github/workflows/support-packages.yml
```diff
@@ -0,0 +1,93 @@
...
+            ${{ github.workspace }}/collector/LICENSE-kernel-modules.txt \
+            gs://mauro-drivers-test/drivers \
+            /tmp/support-packages/metadata \
+            /tmp/support-packages/output
+
+      - name: Create index file
+        run: |
+          relative_path="collector/support-packages"
+
+          if [[ "${{ github.event_name == 'pull_request' }}" == "true" ]]; then
```
ovalenti, line 55: So, those are useful once the workflow condition has been locally changed to run on a PR, right ?

.github/workflows/support-packages.yml
```diff
@@ -0,0 +1,93 @@
...
+
+          ${{ github.workspace }}/kernel-modules/support-packages/05-create-index.py \
+            /tmp/support-packages/metadata \
+            /tmp/support-packages/output
+
+      - name: Set output bucket
+        id: output-bucket
+        run: |
+          if [[ "${{ github.event_name == 'pull_request' }}" == "true" ]]; then
+            echo "output-bucket=mauro-drivers-test/support-packages/${GITHUB_HEAD_REF}/${{ github.run_id }}" >> "$GITHUB_OUTPUT"
```
ovalenti, line 70: We may want to have a TODO token here as well to not forget about it

ovalenti, line 70: Is the gs:// url scheme required only for some operations ?

.github/support-packages/create-metadata-dir.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+for version_dir in /tmp/support-packages/metadata/collector-versions/*; do
+    version="$(basename "$version_dir")"
+    git checkout "${version}"
+
+    driver_version="$(cat "${GITHUB_WORKSPACE}/kernel-modules/MODULE_VERSION")"
+    echo "${driver_version}" > "${version_dir}/MODULE_VERSION"
```
ovalenti, line 10: I am curious about these two lines. This could probably be written as a `cp` call. Or maybe the intent is to do some checking ?

Molter73: No, you are right, I think a `cp` command would suffice, I was just doing mindless copy pasting, I'll fix it.

.github/workflows/support-packages.yml
Molter73: Yes, the idea is to not push support-packages to the production bucket when running PRs, since we could testing changes that break support packages.

.github/workflows/support-packages.yml
Molter73: I'll add it, I also have a follow up for gathering all the buckets and environment variables used across multiple workflows in an init workflow (#944). It adds some boiler plating, but I think it would make managing these kinds of things way easier.

Molter73: Unfortunately, when we run a `gsutil` command, we need to specify it, as it is the protocol used to communicate with the bucket, but the `google-github-actions/upload-cloud-storage@v1` actions assumes this and breaks if you add the `gs://` portion to the URL.

erthalion (approved):

.github/workflows/support-packages.yml
ovalenti: Well, thanks for the explanation :+1:

.github/workflows/support-packages.yml
```diff
@@ -0,0 +1,94 @@
...
+
+      - name: Set output bucket
+        id: output-bucket
+        run: |
+          # TODO: Change the bucket to the production one, once we are done testing
+          if [[ "${{ github.event_name == 'pull_request' }}" == "true" ]]; then
+            echo "output-bucket=mauro-drivers-test/support-packages/${GITHUB_HEAD_REF}/${{ github.run_id }}" >> "$GITHUB_OUTPUT"
+          else
+            echo "output-bucket=mauro-drivers-test/support-packages/master" >> "$GITHUB_OUTPUT"
+          fi
```
JoukoVirtanen, line 74: Echo this to stdout as well.

Molter73: The `google-github-actions/upload-cloud-storage@v1` action prints the full path where the files are being pushed to, do you think that would be enough?

Molter73: It also shows the destination which is the same as this output variable.

PR #933, opened 2022-12-12
Follow up PR for GHA driver builds

Molter73:
## Description

This PR is meant for follow up fixes after merging the driver builds to the master branch.

## Checklist

## Testing Performed

CI run should be enough.

### DIFF
--- a/.github/drivers/mock-cache.sh
+++ b/.github/drivers/mock-cache.sh
@@ -13,6 +13,7 @@ for module_version_dir in /tmp/kobuild-tmp/versions-src/*; do
 
     # If the bucket doesn't exist, we are building a new version of drivers
     if ! gsutil ls "gs://mauro-drivers-test/drivers/${module_version}/" 1>&2 2> /dev/null; then
+        echo "Could not find cached drivers for '${module_version}'. Continuing with no cache"
         continue
     fi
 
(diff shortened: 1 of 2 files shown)


.github/drivers/mock-cache.sh
```diff
@@ -2,17 +2,24 @@
...
 # This script creates a mock cache of drivers, meaning the actual drivers
 # aren't downloaded, but the rest of the scripts will think they are.
 
 for module_version_dir in /tmp/kobuild-tmp/versions-src/*; do
     module_version="$(basename "${module_version_dir}")"
     mkdir -p "/tmp/kernel-modules/$module_version/"
 
+    # If the bucket doesn't exist, we are building a new version of drivers
+    if ! gsutil ls "gs://mauro-drivers-test/drivers/${module_version}/" 1>&2 2> /dev/null; then
+        continue
```
erthalion, line 16: Probably worth a warning message in the log, just in case?

erthalion (approved):

Molter73:
The failure on the `check-drivers-build` check is due to a flake that happens when a large number of drivers is built, it is also present in OSCI. For some reason, a driver that was built correctly leaves the failure file when it should be deleted. I will take a closer look and try to come up with a solution for it in a follow up PR.

PR #936, opened 2022-12-13
More fixes to driver builds on GHA

Molter73:
## Description

This PR makes 2 small changes:
- Properly silences a `gsutil ls` as to not flood logs.
- Ensures the same location is used for pushing drivers and pulling them back down.

## Checklist

## Testing Performed

CI is enough.

### DIFF
--- a/.github/drivers/mock-cache.sh
+++ b/.github/drivers/mock-cache.sh
@@ -12,7 +12,7 @@ for module_version_dir in /tmp/kobuild-tmp/versions-src/*; do
     mkdir -p "/tmp/kernel-modules/$module_version/"
 
     # If the bucket doesn't exist, we are building a new version of drivers
-    if ! gsutil ls "gs://mauro-drivers-test/drivers/${module_version}/" 1>&2 2> /dev/null; then
+    if ! gsutil ls "gs://mauro-drivers-test/drivers/${module_version}/" > /dev/null; then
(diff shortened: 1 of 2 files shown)


.github/drivers/mock-cache.sh
```diff
@@ -12,7 +12,7 @@ for module_version_dir in /tmp/kobuild-tmp/versions-src/*; do
     mkdir -p "/tmp/kernel-modules/$module_version/"
 
     # If the bucket doesn't exist, we are building a new version of drivers
-    if ! gsutil ls "gs://mauro-drivers-test/drivers/${module_version}/" 1>&2 2> /dev/null; then
+    if ! gsutil ls "gs://mauro-drivers-test/drivers/${module_version}/" > /dev/null; then
```
Stringy, line 15: am I correct in thinking it was the order of the old redirections that was wrong? don't we now get stderr written to the logs? does any get written at all?

Stringy (approved):
LGTM, just one minor question

.github/drivers/mock-cache.sh
Molter73: Probably, though `redirect stdout to stderr and stderr to /dev/null` looks correct to me, I need to brush up on bash redirections. Nothing get print to the logs, unless we get an error which... Sounds kinda useful?

PR #946, opened 2022-12-19
Allow up to 10 forks of ansible

Molter73:
## Description

By default ansible allows up to 5 forks, which was causing a slow down on tests run on 6 or more VMs. This patch ups the max number of forks to 10.

## Checklist

## Testing Performed
CI with `all-integration-tests` label should be enough.

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -3,4 +3,4 @@ stdout_callback=debug
 stderr_callback=debug
 host_key_checking=False
 remote_tmp = /tmp/ansible
-
+forks = 10


ovalenti (approved):
Not an expert of this, but it looks similar to the -j parameter of make.

https://docs.ansible.com/ansible/latest/reference_appendices/config.html?sc_cid=701f2000001OH7YAAW#default-forks

Molter73:
> Not an expert of this, but it looks similar to the -j parameter of make.

Basically yes, the way I see it, ansible forks a new process for every task it needs to accomplish, if you have more than 5 (by default) it waits for one to finish before running the 6th one.

PR #947, opened 2022-12-19
ROX-13628 Create an external representation of endpoints

ovalenti:
## Description

In Collector, an endpoint now represents a listening socket with a reference to the process making use of it. When an endpoint is published, however, the process information is stripped down to represent a the name and parameters of this process instead of the precise instance.

This as an impact on the life-cycle of endpoint objects "as seen from the outside". For instance, the same endpoint opened by several processes having the same attributes should appear as a single endpoint.

In order to implement this model, we propose to squash all instances of internal endpoints which "look the same" and publish the result.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

### DIFF
--- a/collector/lib/NetworkStatusNotifier.cpp
+++ b/collector/lib/NetworkStatusNotifier.cpp
@@ -229,11 +229,9 @@ void NetworkStatusNotifier::ConvertEndpointsToAdvertisedRepresentation(const Con
   for (auto cep : from) {
     auto insert_result = to.insert(cep);
 
-    if (!insert_result.second && cep.second.IsActive()) {
-      /* collision case
-         Replace the existing endpoint by the "active" one.
-         Since they are equal, replacing an active endpoint with an active one doesn't make any difference. */
-      (*insert_result.first).second = cep.second;
+    if (!insert_result.second) {
+      // collision case: keep the latest activity
+      (*insert_result.first).second.MergeFrom(cep.second);
     }
   }
 }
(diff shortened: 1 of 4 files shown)


collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -204,11 +204,38 @@ bool NetworkStatusNotifier::UpdateAllConnsAndEndpoints() {
...
 }
 
+bool NetworkStatusNotifier::ExternalContainerEndpointEquality::operator()(const ContainerEndpoint& lhs, const ContainerEndpoint& rhs) const {
+  if (bool(lhs.originator()) != bool(lhs.originator())) {
+    return false;
+  }
+
+  if (lhs.originator()) {
+    auto& lhs_oritinator = *lhs.originator();
+    auto& rhs_oritinator = *rhs.originator();
```
JoukoVirtanen, line 214:
```suggestion
    auto& rhs_originator = *rhs.originator();
```
Fix other typo as well.

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -204,11 +204,38 @@ bool NetworkStatusNotifier::UpdateAllConnsAndEndpoints() {
   return true;
 }
 
+bool NetworkStatusNotifier::ExternalContainerEndpointEquality::operator()(const ContainerEndpoint& lhs, const ContainerEndpoint& rhs) const {
+  if (bool(lhs.originator()) != bool(lhs.originator())) {
```
JoukoVirtanen, line 208:
```suggestion
  if (bool(lhs.originator()) != bool(rhs.originator())) {
```

JoukoVirtanen:
Add some comments to the code to explain why you are doing these things.

PR #948, opened 2022-12-20
Split builder and dev images

Molter73:
## Description

The purpose of this change is more organizational than anything else. We currently build the builder image as a development environment, which means we install quite a few packages in the image that don't actually serve the purpose of building collector.

Splitting the image provides us with:
- More granularity and visibility into what packages are required for building collector and which for development.
- Reduced image size for the builder.

This first commit does not add the development image the our CI, instead relies on the builder image built by CI as cache for building the development image locally.

I've also take this opportunity to move the 'third_party' directory inside the 'builder' directory, since it is only used when building the builder image and allows for a smaller context to be sent over.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Ensure downstream build works properly with these changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,15 +31,30 @@ container-dockerfile-dev:
 builder:
 ifdef BUILD_BUILDER_IMAGE
 	docker build \
+		--target builder \
 		--build-arg NPROCS=$(NPROCS) \
 		--cache-from quay.io/stackrox-io/collector-builder:cache \
 		--cache-from quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG) \
 		-t quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG) \
 		-f "$(CURDIR)/builder/Dockerfile" \
-		.
+		$(CURDIR)/builder/
 else
 	docker pull quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG)
 endif
+
+# For now we use the 'cache-gha' image, since it provides split layers which
+# allows for caching.
+.PHONY: dev-image
+dev-image:
+	docker build \
+		--target dev \
+		--build-arg NPROCS=$(NPROCS) \
+		--cache-from quay.io/stackrox-io/collector-builder:cache-gha \
+		--cache-from quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG) \
+		--cache-from quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG)-dev \
+		-t quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG)-dev \
+		-f "$(CURDIR)/builder/Dockerfile" \
+		$(CURDIR)/builder/
 
 collector: builder
 	make -C collector collector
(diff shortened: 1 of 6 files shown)


builder/Dockerfile
```diff
@@ -33,31 +26,42 @@ RUN dnf -y update \
...
-RUN rm -rf ${BUILD_DIR}
+FROM builder as dev
 
-RUN echo '/usr/local/lib' > /etc/ld.so.conf.d/usrlocallib.conf && ldconfig
+RUN dnf -y update \
+    && dnf -y install --nobest \
+        bison \
+        ca-certificates \
+        clang \
+        cracklib-dicts \
```
ovalenti, line 53:
```suggestion
        clang \
        clang-tools-extra \
        cracklib-dicts \
```
:wink:

Molter73: I was gonna add this alongside a few changes to improve development experience with `clangd`, but alright, we can add it.

ovalenti: I am fine with this dependency being added later; I was just eager

robbycochran, line 53: cracklib-dicts was a dependency needed for running sshd -- but that is an outdated development model. I currently use the [docker toolchain](https://www.jetbrains.com/help/clion/clion-toolchains-in-docker.html) in Clion which no longer needs sshd. If sshd isn't needed by any other development workflows I suggest that we remove it.

Molter73: @JoukoVirtanen I believe you are also using CLion, could you confirm you also don't need `sshd` anymore so I can remove it?

Molter73: Just chatted with Jouko and he is no longer using this, so I'll go ahead and drop the entire ssh workflow

Makefile
```diff
@@ -110,18 +120,11 @@ ci-benchmarks:
...
 .PHONY: start-dev
-start-dev: builder teardown-dev $(DEV_SSH_SERVER_KEY)
-	docker run -d \
+start-dev: dev-image teardown-dev
+	docker run -id \
 		--name collector_remote_dev \
-		--cap-add sys_ptrace -p127.0.0.1:$(LOCAL_SSH_PORT):22 \
-		-v $(DEV_SSH_SERVER_KEY):/etc/sshkeys/ssh_host_ed25519_key:ro \
-		quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG)
+		quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG)-dev
```
Molter73, line 127: We might want to mount the source directory here directly, now that we don't have ssh for mapping it, what do folks think?
```make
start-dev: dev-image teardown-dev
	docker run -id \
		--name collector_remote_dev \
		-v $(CURDIR):$(CURDIR) \
		quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG)-dev

ovalenti: Absolutely, this will better match the VSCode "way", in particular.

Molter73:
Regarding the `Ensure downstream build works properly with these changes.` check, looks like all we'll need is to symlink the `builder/third_party` directory to `third_party`, once we merge this PR I'll create a MR downstream with this small fix.

Molter73:
This PR has become quite outdated, if we still want to split the image we can create a separate PR.

PR #953, opened 2022-12-22
Ignore empty driver build failure files

Molter73:
## Description

When compiling a large number of drivers, sometimes some BPF drivers for kernels that don't support it are reported as failed to build, even though the actual build shows a message stating that it is being skipped. Since we redirect stderr to this file, an empty file would be the same as no error, so we should be ok to ignore it.

I suspect when the driver is skipped there might be some timing issue with the subshell being spawned to run tee, making it so we attempt to delete the file before it is actually created and then the subshell comes up and creates the empty file.

## Checklist

## Testing Performed

- [x] Run with `no-cache` label should result in a green check for the driver build check step


Molter73:
Holding off on this, since it works on GHA but fails on OSCI. We can merge it once we drop OSCI.

Stringy (approved):

PR #954, opened 2023-01-04
Move CI to GHA

Molter73:
## Description

This is the final step for making GHA the official CI solution, OSCI needs to be taken offline first. The PR will remain in draft in order to prevent OSCI from running (GHA jobs trigger regardless of this), once OSCI is disabled we can open and merge this PR directly.

## Checklist

## Testing Performed

Testing as much of the affected CI jobs as possible through the `build-builder-image` and `no-cache` labels, as well as testing the support-packages. final test will need to happen on the master branch though.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -99,11 +99,7 @@ jobs:
       drivers-bucket: ${{ needs.init.outputs.drivers-bucket }}
       support-packages-bucket: ${{ needs.init.outputs.support-packages-bucket }}
       public-support-packages-bucket: ${{ needs.init.outputs.public-support-packages-bucket }}
-    # If you want to test this workflow in a PR, you will need to replace
-    # this condition with `${{ always() }}`, since skipping the upload of
-    # bundles will cause this workflow to skip as well.
-    # if: ${{ github.event_name != 'pull_request' }}
-    if: always()
+    if: always() && (github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'test-support-packages'))
     needs:
     - init
     - build-drivers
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -98,10 +98,12 @@ jobs:
       branch-name: ${{ needs.build-drivers.outputs.branch-name }}
       drivers-bucket: ${{ needs.init.outputs.drivers-bucket }}
       support-packages-bucket: ${{ needs.init.outputs.support-packages-bucket }}
+      public-support-packages-bucket: ${{ needs.init.outputs.public-support-packages-bucket }}
     # If you want to test this workflow in a PR, you will need to replace
     # this condition with `${{ always() }}`, since skipping the upload of
     # bundles will cause this workflow to skip as well.
-    if: ${{ github.event_name != 'pull_request' }}
+    # if: ${{ github.event_name != 'pull_request' }}
+    if: always()
```
Stringy, line 106: just a note to revert this before merging. 

Also would it be worth adding a label to trigger this behaviour?

Stringy (approved):

.github/workflows/main.yml
Molter73: Maybe we could achieve this with a condition similar to this:
```yml
if: always() && (github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'test-support-package')
```
I think it is a little hard to read, WDYT?

Stringy: yeah I agree it's pretty complex as a condition but I think it'll make testing a lot easier and won't risk someone merging `if: always()`

Molter73: I'll add it and get it tested 🙂

Molter73: Looks like it is working properly.

Stringy (approved):

PR #898, opened 2022-11-12
Jv rox 13507 collector integration test that opens multiple ports in the same command

JoukoVirtanen:
## Description

Adds a test that executes the following command "socat TCP-LISTEN:80,fork STDOUT & socat  TCP-LISTEN:8080,fork STDOUT" and ensure that the container endpoints are reported correctly.

## Checklist
- [ ] Investigated and inspected CI test results

## Testing Performed

Ran integration tests locally with make integration-tests-socat. Ran CI.

PR #957, opened 2023-01-04
Fixes shell escaping problem in socat test

Stringy:
## Description

Fixes a shell escaping problem in the socat test, by simplifying the command into two executions in the socat container.  

## Checklist

**Automated testing**
 ~- [ ] Added unit tests~

## Testing Performed

Tested locally against RHEL8

### DIFF
--- a/integration-tests/suites/socat.go
+++ b/integration-tests/suites/socat.go
@@ -34,9 +34,17 @@ func (s *SocatTestSuite) SetupSuite() {
 
 	processImage := common.QaImage("quay.io/rhacs-eng/qa", "socat")
 
-	containerID, err := s.launchContainer("socat", "-e", "PORT1=80", "-e", "PORT2=8080", processImage, "/bin/sh", "-c", "socat TCP-LISTEN:$PORT1,fork STDOUT & socat TCP-LISTEN:$PORT2,fork STDOUT")
-
+	// the socat container only needs to exist long enough for use to run both
+	// socat commands. 300 seconds should be more than long enough.
+	containerID, err := s.launchContainer("socat", processImage, "/bin/sh", "-c", "/bin/sleep 300")
 	s.Require().NoError(err)
+
+	_, err = s.execContainer("socat", []string{"/bin/sh", "-c", "socat TCP-LISTEN:80,fork STDOUT &"})
+	s.Require().NoError(err)
+
+	_, err = s.execContainer("socat", []string{"/bin/sh", "-c", "socat TCP-LISTEN:8080,fork STDOUT &"})
+	s.Require().NoError(err)
+
 	s.serverContainer = common.ContainerShortID(containerID)
 
 	time.Sleep(6 * time.Second)
@@ -68,7 +76,7 @@ func (s *SocatTestSuite) TestSocat() {
 		assert.FailNowf(s.T(), "", "only retrieved %d endpoints (expect 2)", len(endpoints))
 	}
 
-	assert.Equal(s.T(), 3, len(processes))
+	assert.Equal(s.T(), 6, len(processes))
 
 	endpoint80, err := getEndpointByPort(endpoints, 80)
 	s.Require().NoError(err)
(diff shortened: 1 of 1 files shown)


integration-tests/suites/socat.go
```diff
@@ -34,9 +34,17 @@ func (s *SocatTestSuite) SetupSuite() {
...
+	// socat commands. 300 seconds should be more than long enough.
+	containerID, err := s.launchContainer("socat", processImage, "/bin/sh", "-c", "/bin/sleep 300")
+	s.Require().NoError(err)
+
+	_, err = s.execContainer("socat", []string{"/bin/sh", "-c", "socat TCP-LISTEN:80,fork STDOUT &"})
+	s.Require().NoError(err)
 
+	_, err = s.execContainer("socat", []string{"/bin/sh", "-c", "socat TCP-LISTEN:8080,fork STDOUT &"})
 	s.Require().NoError(err)
+
```
JoukoVirtanen, line 47: I think that this changes what this test checks for. Refer to https://github.com/stackrox/collector/pull/898 and https://issues.redhat.com/browse/ROX-13507. Also take a look at lines 44 and 45 in qa-tests-backend/src/test/groovy/ProcessesListeningOnPortsTest.groovy here https://github.com/stackrox/stackrox/pull/3546/files

That test was failing due to errors with running two instances of socat in the same command line. This test was a way of catching that error within collector and making it easier to solve. I named this test poorly. It should probably have been named something like "OpeningTwoPlopsInTheSameCommand".

Stringy: Ok thanks for the heads up, I'll rework this to better match the intended test case

Stringy: I am very confused about what running two processes from the same 'command line' actually tests. From the kernel's perspective and from Collector's they are just two processes, so that suggests that it might be the timing of those processes executing and possibly related to the scrape interval? I'm afraid those links do not shine much light on the overall context of this test.

In the future could we _please_ ensure that tests are well named and well documented with a much context as possible about what they're trying to test and any assumptions they are making.

JoukoVirtanen: There is more information here https://srox.slack.com/archives/C02PQ7YUNSV/p1666921527207769

Basically the error was that in collector the two endpoints were incorrectly associated with the same PID, while process messages in collector had two different PIDs, which is correct.

I will also add more comments to the test to explain the error that it tests for.

integration-tests/suites/socat.go
```diff
@@ -34,9 +34,17 @@ func (s *SocatTestSuite) SetupSuite() {
 
 	processImage := common.QaImage("quay.io/rhacs-eng/qa", "socat")
 
-	containerID, err := s.launchContainer("socat", "-e", "PORT1=80", "-e", "PORT2=8080", processImage, "/bin/sh", "-c", "socat TCP-LISTEN:$PORT1,fork STDOUT & socat TCP-LISTEN:$PORT2,fork STDOUT")
+	// the socat container only needs to exist long enough for use to run both
```
JoukoVirtanen, line 37:
```suggestion
	// In a previous version of collector the two child socat processes had the same PID in the Endpoints. They needed to have different PIDs. This led to incorrect arguments being reported for one of the child socat processes. This test is meant to ensure that this bug is fixed.
```

integration-tests/suites/socat.go
Stringy: As discussed offline - I've tested this new approach against `3.12.x-3-g1cf8af5c3f` from [1cf8af5c](1cf8af5c3f06f625e6f50520f6e0b427d8676db3) and `3.12.x-10-g74e322fd28` from [74e322fd](https://github.com/stackrox/collector/commit/74e322fd) (the commit just prior to the fix being merged) as well as a new build from master. The test fails correctly on the two older versions, and passes on the new version. I think this gives us a decent amount of confidence that the test case is still covered by this approach.

JoukoVirtanen (approved):

PR #958, opened 2023-01-04
Unifies integration test notifications

Stringy:
## Description

Puts notifications for integration tests one level higher so a notification is sent once instead of for each individual integration test platform.

## Checklist

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -47,7 +47,7 @@ jobs:
 
   notify:
     runs-on: ubuntu-latest
-    if: always() && contains(join(needs.*.result, ','), 'failure') # && github.event_name == 'push'
+    if: always() && contains(join(needs.*.result, ','), 'failure') && github.event_name == 'push'
     needs:
       - required-integration-tests
       - all-integration-tests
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -44,3 +44,22 @@ jobs:
...
+    if: always() && contains(join(needs.*.result, ','), 'failure') # && github.event_name == 'push'
+    needs:
+      - required-integration-tests
+      - all-integration-tests
+    steps:
+      - name: Slack notification
+        uses: rtCamp/action-slack-notify@v2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_ONCALL }}
+          SLACK_CHANNEL: oncall
```
Molter73, line 59: Could we also move them to the `collector-notifications` channel? I don't really see the value in sending these to an oncall channel.
```suggestion
          SLACK_WEBHOOK: ${{ secrets.SLACK_COLLECTOR_NOTIFICATIONS_WEBHOOK }}
          SLACK_CHANNEL: collector-notifications
```

Stringy: Sure, sounds good

Molter73 (approved):
LGTM!

PR #959, opened 2023-01-04
rox-14174 collector should report udp endpoints

JoukoVirtanen:
## Description

Currently ProcfsScraper only scrapes TCP endpoints. With these changes it will also scrape UDP endpoints. This does not change anything in regards to how connections are reported.

## Checklist
- [x] Run in stackrox/stackrox CI for master branch https://github.com/stackrox/stackrox/pull/4371
       Upgrade test failed, because of processes listening on ports.
- [x] Run manually and see UDP endpoints in network graph

**Automated testing**
  - [x] Added integration tests

## Testing Performed

CI is sufficient

### DIFF
--- a/collector/connscrape.cpp
+++ b/collector/connscrape.cpp
@@ -38,12 +38,13 @@ BoolEnvVar scrape_endpoints("SCRAPE_ENDPOINTS", true);
 
 int main(int argc, char** argv) {
   const char* proc_dir = "/proc";
+  bool collect_udp = true;
 
   if (argc > 1) {
     proc_dir = argv[1];
   }
 
-  ConnScraper scraper(proc_dir);
+  ConnScraper scraper(proc_dir, collect_udp);
   std::vector<Connection> conns;
   std::vector<ContainerEndpoint> endpoints;
 
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -56,6 +56,8 @@ BoolEnvVar set_enable_core_dump("ENABLE_CORE_DUMP", false);
 BoolEnvVar set_processes_listening_on_ports("ROX_PROCESSES_LISTENING_ON_PORT", CollectorConfig::kEnableProcessesListeningOnPorts);
 
 BoolEnvVar core_bpf_hardfail("ROX_COLLECTOR_CORE_BPF_HARDFAIL", true);
+// If true, UDP listening endpoints will be reported
+BoolEnvVar set_are_udp_listening_endpoints_collected("ROX_COLLECT_UDP_LISTENING_ENDPOINTS", false);
 
 BoolEnvVar set_import_users("ROX_COLLECTOR_SET_IMPORT_USERS", false);
 
@@ -189,6 +191,13 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
 
   if (set_enable_core_dump) {
     enable_core_dump_ = true;
+  }
+
+  if (set_are_udp_listening_endpoints_collected) {
+    are_udp_listening_endpoints_collected_ = true;
+    CLOG(INFO) << "Collecting UDP listening endpoints";
+  } else {
+    CLOG(INFO) << "Not collecting UDP listening endpoints";
   }
 
   HandleAfterglowEnvVars();
--- a/collector/lib/CollectorConfig.h
+++ b/collector/lib/CollectorConfig.h
@@ -132,11 +132,8 @@ end
   bool enable_core_dump_ = false;
   bool enable_processes_listening_on_ports_;
   bool core_bpf_hardfail_;
-<<<<<<< HEAD
   bool import_users_;
-=======
   bool are_udp_listening_endpoints_collected_ = false;
->>>>>>> a4040548 (X-Smart-Squash: Squashed 32 commits:)
 
   Json::Value tls_config_;
 };
--- a/collector/lib/CollectorConfig.h
+++ b/collector/lib/CollectorConfig.h
@@ -108,6 +108,7 @@ end
   bool IsProcessesListeningOnPortsEnabled() const { return enable_processes_listening_on_ports_; }
   bool CoReBPFHardfail() const { return core_bpf_hardfail_; }
   bool ImportUsers() const { return import_users_; }
+  bool AreUdpListeningEndpointsCollected() const { return are_udp_listening_endpoints_collected_; }
 
   std::shared_ptr<grpc::Channel> grpc_channel;
 
@@ -131,7 +132,11 @@ end
   bool enable_core_dump_ = false;
   bool enable_processes_listening_on_ports_;
   bool core_bpf_hardfail_;
+<<<<<<< HEAD
   bool import_users_;
+=======
+  bool are_udp_listening_endpoints_collected_ = false;
+>>>>>>> a4040548 (X-Smart-Squash: Squashed 32 commits:)
 
   Json::Value tls_config_;
 };
--- a/collector/lib/NetworkConnection.cpp
+++ b/collector/lib/NetworkConnection.cpp
@@ -55,9 +55,9 @@ size_t Hash(const L4ProtoPortPair& pp) {
 
 std::ostream& operator<<(std::ostream& os, const ContainerEndpoint& container_endpoint) {
   if (container_endpoint.originator()) {
-    return os << container_endpoint.container() << ": " << container_endpoint.endpoint() << ": " << *container_endpoint.originator();
+    return os << container_endpoint.container() << ": " << container_endpoint.endpoint() << ": " << container_endpoint.l4proto() << ": " << *container_endpoint.originator();
   } else {
-    return os << container_endpoint.container() << ": " << container_endpoint.endpoint();
+    return os << container_endpoint.container() << ": " << container_endpoint.endpoint() << ": " << container_endpoint.l4proto();
   }
 }
 
--- a/collector/lib/ProcfsScraper.cpp
+++ b/collector/lib/ProcfsScraper.cpp
@@ -313,24 +313,34 @@ bool LocalIsServer(const Endpoint& local, const Endpoint& remote, const Unordere
   return IsEphemeralPort(remote.port()) > IsEphemeralPort(local.port());
 }
 
-// ReadConnectionsFromFile reads all connections from a `net/tcp[6]` file and stores them by inode in the given map.
-bool ReadConnectionsFromFile(Address::Family family, L4Proto l4proto, std::FILE* f,
-                             UnorderedMap<ino_t, ConnInfo>* connections, UnorderedMap<ino_t, EndpointInfo>* listen_endpoints) {
+// Reads a `net/(tcp|udp)[6]` file and adds the connections to the conn_lines vector
+// Returns a boolean indicating success or failure
+bool ReadConnLinesFromFile(std::FILE* f, Address::Family family, std::vector<ConnLineData>* conn_lines) {
   char line[512];
 
   if (!std::fgets(line, sizeof(line), f)) return false;  // ignore the first *header) line.
 
-  UnorderedSet<Endpoint> all_listen_endpoints;
-
   while (std::fgets(line, sizeof(line), f)) {
     ConnLineData data;
     if (!ParseConnLine(line, line + sizeof(line), family, &data)) continue;
+    conn_lines->push_back(data);
+  }
+
+  return true;
+}
+
+// ReadTcpConnectionsAndEndpointsFromFile reads all connections from a `net/tcp[6]` file and stores them by inode in the given map.
+bool ProcessTcpConnectionsAndEndpoints(const std::vector<ConnLineData>& conn_lines,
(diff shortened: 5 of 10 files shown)


collector/lib/ProcfsScraper.cpp
```diff
@@ -349,9 +349,35 @@ bool ReadConnectionsFromFile(Address::Family family, L4Proto l4proto, std::FILE*
   return true;
 }
 
-// GetConnections reads all active connections (inode -> connection info mapping) for a given network NS, addressed by
+// ReadUdpEndpointsFromFile reads all udp endpoints from a `net/udp[6]` file and stores them by inode in the given map.
+bool ReadUdpEndpointsFromFile(Address::Family family, std::FILE* f,
```
JoukoVirtanen, line 353: Perhaps I should not have separate functions for reading UDP and TCP endpoints. For now this seemed like the safest solution.

collector/lib/ProcfsScraper.cpp
```diff
@@ -349,9 +349,35 @@ bool ReadConnectionsFromFile(Address::Family family, L4Proto l4proto, std::FILE*
...
+  char line[512];
+
+  if (!std::fgets(line, sizeof(line), f)) return false;  // ignore the first *header) line.
+
+  UnorderedSet<Endpoint> all_listen_endpoints;
+
+  while (std::fgets(line, sizeof(line), f)) {
+    ConnLineData data;
+    if (!ParseConnLine(line, line + sizeof(line), family, &data)) continue;
+    // I am not sure what the condition of this if statement should be or if there should be one
```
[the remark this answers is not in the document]
Stringy: This also begs the question of whether we also want to report UDP connections as well (I think we probably do)

integration-tests/suites/udp.go
```diff
@@ -0,0 +1,96 @@
+package suites
+
+import (
+	"time"
+
+	"github.com/stackrox/collector/integration-tests/suites/common"
+	"github.com/stretchr/testify/assert"
+)
+
+type UdpTestSuite struct {
```
Stringy, line 10: Can we have some docs here describing what this test is for and how it works

collector/lib/ProcfsScraper.cpp
JoukoVirtanen: Good idea. I took this even a step further and I put that for loop you proposed in its own function. That way it is easier to write unit tests for it.

collector/lib/ProcfsScraper.cpp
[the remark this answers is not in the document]
JoukoVirtanen: I have made the change to the if statement that you proposed, but I am not 100 % certain about it. Perhaps we should discuss these two issues during the huddle.

collector/lib/ProcfsScraper.cpp
```diff
@@ -349,9 +349,35 @@ bool ReadConnectionsFromFile(Address::Family family, L4Proto l4proto, std::FILE*
...
+    // }
+  }
+
+  return true;
+}
+
+// GetConnectionAndEndpoints reads all active connections (inode -> connection info mapping) for a given network NS, addressed by
 // the dir FD for a proc entry of a process in that network namespace.
-bool GetConnections(int dirfd, UnorderedMap<ino_t, ConnInfo>* connections, UnorderedMap<ino_t, EndpointInfo>* listen_endpoints) {
+bool GetConnectionAndEndpoints(int dirfd, UnorderedMap<ino_t, ConnInfo>* connections, UnorderedMap<ino_t, EndpointInfo>* listen_endpoints) {
```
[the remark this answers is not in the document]
JoukoVirtanen: Good catch. Done.

collector/lib/ProcfsScraper.cpp
```diff
@@ -373,6 +399,26 @@ bool GetConnections(int dirfd, UnorderedMap<ino_t, ConnInfo>* connections, Unord
     }
   }
 
+  {
+    FDHandle net_udp_fd = openat(dirfd, "net/udp", O_RDONLY);
+    if (net_udp_fd.valid()) {
+      FileHandle net_udp(std::move(net_udp_fd), "r");
+      success = ReadUdpEndpointsFromFile(Address::Family::IPV4, net_udp, listen_endpoints) && success;
+    } else {
+      success = false;  // there should always be a net/tcp file
```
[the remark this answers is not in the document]
JoukoVirtanen: Yes. Done.

integration-tests/suites/udp.go
```diff
@@ -0,0 +1,96 @@
...
+	containerSendtoID, err := s.launchContainer("socat-sendto", processImage, "/bin/sh", "-c", "/bin/sleep 300")
+	s.Require().NoError(err)
+
+	_, err = s.execContainer("socat-listen", []string{"/bin/sh", "-c", "socat UDP-LISTEN:80,fork STDOUT &"})
+	s.Require().NoError(err)
+	listenIP, err := s.getIPAddress("socat-listen")
+	s.Require().NoError(err)
+
+	_, err = s.execContainer("socat-sendto", []string{"/bin/sh", "-c", "socat UDP-SENDTO:" + listenIP + ":80,fork STDIN &"})
+	s.Require().NoError(err)
```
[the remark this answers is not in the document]
JoukoVirtanen: I have removed the forks.

integration-tests/suites/udp.go
JoukoVirtanen: Done

collector/lib/ProcfsScraper.cpp
```diff
@@ -367,7 +412,37 @@ bool GetConnections(int dirfd, UnorderedMap<ino_t, ConnInfo>* connections, Unord
...
+
+  {
+    FDHandle net_udp6_fd = openat(dirfd, "net/udp6", O_RDONLY);
+    if (net_udp6_fd.valid()) {
+      FileHandle net_udp6(std::move(net_udp6_fd), "r");
+      success = ReadConnectionsAndEndpointsFromFile(Address::Family::IPV6, net_udp6, L4Proto::UDP, connections, listen_endpoints) && success;
+    } else {
+      success = false;
+    }
+  }
```
JoukoVirtanen, line 439: Seems like this can be refactored. We could loop through the files instead of repeating all of this code. I would do the refactor, but I am not sure about the correctness of the code as is. I am not sure that it is correct to set success to false in all of these cases, if valid() is false. That was done for both net/tcp and net/tcp6. Should it only have been done for net/tcp?

Stringy: I think you're right - a loop might be a fair bit neater (and I think we can use a list of `std::tuple` to store the various bits of data that each invocation needs (filename, family, protocol))

On the point about `fd.valid()`, I think that condition holds for all the tcp and udp files since they should always exist (unless the kernel is configured without procfs, or without the necessary networking stacks but that feels very unlikely for our target kernels) There may still be some benefit to adding more granular error reporting between the two protocols, but that will mean splitting this function in two (tcp & udp) and letting the caller call both, and decide how to handle the errors. Not necessarily something we need to do now.

Stringy (approved):

collector/lib/ProcfsScraper.cpp
```diff
@@ -349,27 +359,66 @@ bool ReadConnectionsFromFile(Address::Family family, L4Proto l4proto, std::FILE*
...
+    L4Proto l4proto = std::get<1>(net_file_with_l4protocol);
+    for (Address::Family address_family : address_families) {
+      std::string net_file_ipv = (address_family == Address::Family::IPV4) ? net_file : net_file + "6"; 
+      FDHandle net_file_ipv_fd = openat(dirfd, net_file_ipv.c_str(), O_RDONLY);
+      if (net_file_ipv_fd.valid()) {
+        FileHandle net(std::move(net_file_ipv_fd), "r");
+        success = ReadConnectionsAndEndpointsFromFile(address_family, net, l4proto, connections, listen_endpoints) && success;
+      } else {
+        success = false;
+      }
```
JoukoVirtanen, line 421: @Stringy suggested using list of tuples with (filename, family, protocol). I decided to go with a list of tuples of just (filename, protocol) and loop over family. The protocol can be obtained from the filename so it is not necessary to have the tuple and we could just have a vector of filenames. I still decided to go with the tuple to avoid calling a function that determines the protocol from the filename. I am happy make that change or any other improvement.

Note also that this will return false when any of the files are not there, which will be almost always.

JoukoVirtanen: I don't think it is harmful that this function almost always returns false, but it is a bit odd. It looks if it returns false it checks if it can get the inode again, and if it can it does nothing. If it can't it erases some data.
```
     if (!GetConnectionsAndEndpoints(dirfd, &ns_network_data.connections, listen_endpoints ? &ns_network_data.listen_endpoints : nullptr)) {
          // If there was an error reading connections, that could be due to a number of reasons.
          // We need to differentiate persistent errors (e.g., expected net/tcp6 file not found)
          // from spurious/race condition errors caused by the process disappearing while reading
          // the directory. To determine if the latter is the root cause, we reattempt to read the
          // network namespace inode; if that succeeds, we assume that the process is still alive
          // and any errors encountered are persistent.
          uint64_t netns_inode2;
          if (!GetNetworkNamespace(dirfd, &netns_inode2) || netns_inode2 != netns_inode) {
            conns_by_ns.erase(emplace_res.first);
            continue;
          }
        }
```

collector/connscrape.cpp
```diff
@@ -38,12 +38,13 @@ BoolEnvVar scrape_endpoints("SCRAPE_ENDPOINTS", true);
...
 int main(int argc, char** argv) {
   const char* proc_dir = "/proc";
+  bool collect_udp;
 
   if (argc > 1) {
     proc_dir = argv[1];
   }
 
-  ConnScraper scraper(proc_dir);
+  ConnScraper scraper(proc_dir, collect_udp);
```
ovalenti, line 47: Should we initialize `collect_udp` to something ?

collector/lib/ProcfsScraper.cpp
JoukoVirtanen: I ended up looping through the files and using a tuple though the tuple just holds filename and protocol.

robbycochran (changes requested):

collector/lib/CollectorConfig.cpp
```diff
@@ -55,6 +55,9 @@ BoolEnvVar set_enable_core_dump("ENABLE_CORE_DUMP", false);
 // If true, add originator process information in NetworkEndpoint
 BoolEnvVar set_processes_listening_on_ports("ROX_PROCESSES_LISTENING_ON_PORT", CollectorConfig::kEnableProcessesListeningOnPorts);
 
+// If true, UDP listening endpoints will be reported
+BoolEnvVar set_is_udp_collected("ROX_COLLECT_UDP", false);
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

collector/lib/CollectorConfig.h
```diff
@@ -131,6 +132,7 @@ end
   bool enable_afterglow_ = true;
   bool enable_core_dump_ = false;
   bool enable_processes_listening_on_ports_;
+  bool are_udp_listening_endpoints_collected_ = false;
```
ovalenti, line 135: I find it simpler when the env-var is the only place where we set the default value.

collector/lib/NetworkConnection.cpp
```diff
@@ -55,9 +55,9 @@ size_t Hash(const L4ProtoPortPair& pp) {
 
 std::ostream& operator<<(std::ostream& os, const ContainerEndpoint& container_endpoint) {
   if (container_endpoint.originator()) {
-    return os << container_endpoint.container() << ": " << container_endpoint.endpoint() << ": " << *container_endpoint.originator();
+    return os << container_endpoint.container() << ": " << container_endpoint.endpoint() << ": " << container_endpoint.l4proto() << ": " << *container_endpoint.originator();
   } else {
-    return os << container_endpoint.container() << ": " << container_endpoint.endpoint();
+    return os << container_endpoint.container() << ": " << container_endpoint.endpoint() << ": " << container_endpoint.l4proto();
```
ovalenti, line 60: Maybe add the l4proto to the endpoint() string representation ?

collector/lib/CollectorConfig.cpp
```diff
@@ -195,6 +199,13 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
     enable_core_dump_ = true;
   }
 
+  if (set_are_udp_listening_endpoints_collected) {
+    are_udp_listening_endpoints_collected_ = true;
```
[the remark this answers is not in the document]
JoukoVirtanen: I removed

are_udp_listening_endpoints_collected_ = set_are_udp_listening_endpoints_collected.value();

collector/lib/NetworkConnection.cpp
JoukoVirtanen: Done

collector/lib/CollectorConfig.h
JoukoVirtanen: I am not 100 % sure what you mean. If the environment variable is not set, we need a default value in the code.

PR #964, opened 2023-01-10
PyTest integration tests prototype

Stringy:
## Description

This contains some experiments with pytest conversion of the existing integration tests.


integration-tests/python/conftest.py
```diff
@@ -0,0 +1,123 @@
...
+    collector = runtime.run(
+        collector_image,
+        name="collector",
+        privileged=True,
+        network="host",
+        mounts=list(mounts.values()),
+        env=env,
+    )
+
+    time.sleep(30)
```
Molter73, line 104: 30 seconds?! We really need to just add some testing probe that we can query using http

integration-tests/python/conftest.py
```diff
@@ -0,0 +1,123 @@
...
+
+    yield collector
+
+    try:
+        collector.kill()
+    except Exception:
+        pass
```
Molter73, line 111: What if we fail to kill collector before the next test tries to create it again?

Molter73 (commented):
This is awesome ❤️

integration-tests/python/common/events.py
```diff
@@ -0,0 +1,65 @@
...
+                if process_bucket:
+                    container_bucket = process_bucket.bucket(container.encode())
+                    if container_bucket:
+                        return
+                    
+            time.sleep(1)
+
+        raise RuntimeError(f'Timed out waiting for processes for {container}')
+
+    def processes(self, container=None):
```
JoukoVirtanen, line 38: Use a more descriptive function name. Also add a comment explaining what the function does.
```suggestion
    # Gets processes from BoltDB. If container is not set all processes will be returned.
    # If container is set, only processes for that container will be returned.
    def processes(self, container=None):
```

integration-tests/python/common/events.py
```diff
@@ -0,0 +1,65 @@
...
+        pass
+
+class BoltDBEvents(Events):
+    def __init__(self):
+        # lazy load the DB on demand, to avoid
+        # weird behavior when opening the DB really 
+        # early in the test
+        self.db = None
+
+    def await_processes(self, container, timeout=30):
```
JoukoVirtanen, line 25: Could await_processes return the processes as well.

PR #965, opened 2023-01-12
Migrate CPaaS related CI to GHA

Molter73:
## Description

This PR adds a new set of GHA workflows that allow for testing the downstream built drivers, as well as uploading those drivers and corresponding support-packages to GCP.

The new `cpaas-sync-drivers` workflow pulls the downstream built drivers image for all supported architectures (x86, ppc64le and s390x), extracts the drivers and support-packages from it and pushes them to GCP.

The `cpaas` workflow is based on existing workflows that are used in the `main` workflow, but taking parameters specific for the new use cases. It creates collector images embedding the downstream drivers into them and tags them with a `-cpaas` suffix. Those images are then used to run integration tests in offline mode in the pertinent VMs (`rhel` and `rhel-sap`).

There are currently 2 ways of triggering these new workflows:
- Adding the `run-cpaas-steps` label on PRs will trigger the workflows for that PR.
- Once a day, this workflow will run on the master branch, ensuring the downstream built drivers still work OK with the latest upstream changes and uploading any newly built downstream driver.

There are a few things to figure out before we can have these workflows push things to the upstream production buckets:
- Prevent drivers from different architectures from being embedded into the wrong image (i.e embed s390x drivers in a x86 image). This could be solved by pushing the drivers into subdirectories split by arch for the GCP bucket. This might create a separate problem when a running collector tries to download the driver directly, but some messing about with proxies might fix it easily.
- Support-packages don't have the arch for which they are built as part of their name, which means different architectures could collide of make it harder for users to find the correct support-package. Adding the arch to the download URL would fix this, but it would mean we also need to correct the index file generated for [cdn.stackrox.io](https://cdn.stackrox.io/collector/support-packages/index.html).

While we discuss these remaining questions with the team, we could start running these new workflows, since they are pushing to separate buckets and should not interfere with day to day operations.

## Checklist

## Testing Performed

Green CI should be enough.

### DIFF
--- a/.github/workflows/init.yml
+++ b/.github/workflows/init.yml
@@ -71,7 +71,7 @@ jobs:
       - name: Set GCP buckets
         id: gcp-buckets
         run: |
-          STAGING_RELATIVE_PATH="${GITHUB_HEAD_REF}/${{ github.run_id}}"
+          STAGING_RELATIVE_PATH="${GITHUB_HEAD_REF}/${{ github.run_id }}"
 
           MAIN_DRIVER_BUCKET="collector-modules-osci"
           STAGING_DRIVER_BUCKET="stackrox-collector-modules-staging/pr-builds/${STAGING_RELATIVE_PATH}"
--- a/.github/workflows/init.yml
+++ b/.github/workflows/init.yml
@@ -79,19 +79,27 @@ jobs:
           PUBLIC_SUPPORT_PACKAGES_BUCKET="collector-support-public/offline/v1/support-packages"
           CPAAS_DRIVERS_BUCKET="mauro-drivers-test/cpaas/drivers"
           CPAAS_SUPPORT_PACKAGES_BUCKET="mauro-drivers-test/cpaas/support-packages"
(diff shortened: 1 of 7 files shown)


Molter73:
Closing and reopening for CI trigger

.github/workflows/init.yml
```diff
@@ -62,23 +71,38 @@ jobs:
       - name: Set GCP buckets
         id: gcp-buckets
         run: |
+          STAGING_RELATIVE_PATH="${GITHUB_HEAD_REF}/${{ github.run_id}}"
```
ovalenti, line 74:
```suggestion
          STAGING_RELATIVE_PATH="${GITHUB_HEAD_REF}/${{ github.run_id }}"
```

ovalenti (approved):
I may not be authoritative with GHA (yet), but I believe everything I read here makes sense.

Molter73:
For the concerns mentioned in the last part of the PR description, #981 will address them.

PR #969, opened 2023-01-13
Initial support for RHCOS and podman

Stringy:
## Description

Adds RHCOS to ansible playbooks, including ignition configuration to add an SSH key to the host.

Most changes are in the integration tests themselves to allow differences in the container runtime
(only tested on docker and podman currently, and cli compatibility is required)


integration-tests/suites/base.go
```diff
@@ -406,7 +406,7 @@ func (s *IntegrationTestSuiteBase) RunImageWithJSONLabels() {
 func (s *IntegrationTestSuiteBase) StartContainerStats() {
 	name := "container-stats"
 	image := common.QaImage("quay.io/rhacs-eng/collector-performance", "stats")
-	args := []string{name, "-v", "/var/run/docker.sock:/var/run/docker.sock", image}
+	args := []string{name, "-v", common.RuntimeSocket + ":/var/run/docker.sock", image}
```
Molter73, line 409: How does this work? We mount the podman socket to `/var/run/docker.sock` in the container and it just works?

integration-tests/suites/common/executor.go
```diff
@@ -150,6 +159,11 @@ func (e *executor) Exec(args ...string) (string, error) {
 func (e *executor) ExecRetry(args ...string) (res string, err error) {
 	maxAttempts := 3
 	attempt := 0
+
+	if args[0] == RuntimeCommand && RuntimeAsRoot {
+		args = append([]string{"sudo"}, args...)
+	}
```
Molter73, line 165: Could we move this logic inside the `RunCommand` function? Think it would make sense to have it as close to actually running the command as possible. It's also duplicated in the `Exec` function, which is not ideal IMO.

Stringy: Unfortunately we can't easily put it there, because by that point the command is fully constructed and is prepended with various other parts of the command to run on the right VM (gcloud, ssh, etc) so we need to prepend sudo to the start of the command that will be executed on the remote host and unfortunately that does come with a little duplication currently

integration-tests/suites/base.go
Molter73: image

integration-tests/suites/common/executor.go
Molter73: That's unfortunate but understandable.

Molter73 (approved):
LGTM!

ansible/group_vars/all.yml
```diff
@@ -23,6 +28,20 @@ virtual_machines:
       - rhel-8-4-sap-ha
       - rhel-8-6-sap-ha
 
+  rhcos:
+    project: rhcos-cloud
+    username: core
+    images:
+      - rhcos-412-86-202302211547-0-gcp-x86-64
```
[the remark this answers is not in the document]
robbycochran: Agree. Let's prioritize getting automation for using the latest images from https://releases-rhcos-art.apps.ocp-virt.prod.psi.redhat.com/?stream=prod/streams/4.12&arch=x86_64. That would help us catch these issues: https://redhat-internal.slack.com/archives/CFMQ5C2TT/p1679308158011459

PR #970, opened 2023-01-16
ROX-14201 New timer to track time waiting for process-info

ovalenti:
## Description

Falco is requested by Process objects to resolve details of the process they represent. This procedure is asynchronous and Process may wait for it to complete.

"process_info_wait" is the time spent waiting (in us)

## Checklist

### DIFF
--- a/collector/lib/Process.cpp
+++ b/collector/lib/Process.cpp
@@ -155,7 +155,9 @@ void Process::WaitForProcessInfo() const {
   if (process_info_pending_resolution_) {
     std::cv_status status;
 
-    status = process_info_condition_.wait_for(lock, std::chrono::seconds(30));
+    WITH_TIMER(CollectorStats::process_info_wait) {
+      status = process_info_condition_.wait_for(lock, std::chrono::seconds(30));
+    }
 
     CLOG_IF(std::cv_status::timeout == status, ERROR) << "Timed-out waiting for process-info. PID: " << pid();
   }
--- a/collector/lib/CollectorStats.h
+++ b/collector/lib/CollectorStats.h
@@ -12,7 +12,8 @@
   X(net_scrape_update)  \
   X(net_fetch_state)    \
   X(net_create_message) \
-  X(net_write_message)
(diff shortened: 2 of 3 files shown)


collector/lib/Process.cpp
```diff
@@ -155,7 +155,9 @@ void Process::WaitForProcessInfo() const {
   if (process_info_pending_resolution_) {
     std::cv_status status;
 
-    status = process_info_condition_.wait_for(lock, std::chrono::seconds(30));
+    WITH_TIMER(CollectorStats::process_info_wait) {
+      status = process_info_condition_.wait_for(lock, std::chrono::seconds(30));
+    }
```
Molter73, line 160: This looks harmless at first sight, except it is creating and destroying a scoped timer Falco method we invoke, which means it is in a pretty hot path. Are we certain this won't cause any performance degradation? Could we somehow measure time to process the entire resolution of an event instead of every call to Falco?

ovalenti: We are already measuring the time it takes to convert all objects to a message.
The goal here is to know specifically the time involved in the asynchronous communication between Falco and Process.

I agree that this method is on the hot path; however, the specific case when this timer is created may, by construction, only be called once per Process (next call will have the info available). Also, at some point, the Falco thread is actively resolving all the requests; up to now, I never saw more than one call to wait_lock() during a scrape.

ovalenti: The if() statement surrounding the TIMER block is guarded by `process_info_pending_resolution_`, which is set to _false_ once Falco resolved the process-info at line 143.

Thus, if one call blocks, then it should be the last for that process. Most Process objects won't even block at all, because `Process::ProcessInfoResolved()` will have been called before any getter is used.

Molter73: Aaaaaaah I see! I was missing the context, then it should be fine, thanks for explaining it and sorry I missed that!

Molter73 (approved):
LGTM!

PR #974, opened 2023-01-17
Drop dockerized BLOCKLIST

Molter73:
## Description

Some kernels were being blocked from being compiled because GHA applied an additional BLOCKLIST file meant for the now deprecated dockerized build of drivers. This PR makes it so the only BLOCKLIST applied is the one located under `kernel-modules/BLOCKLIST`.

I will create a follow up PR removing the dockerized directory and cleaning up a bit, since we haven't used much of it in the past months.

I've also made it so the list of blocked kernels are output to GHA logs, that way it will be easier for us to spot a missing driver due to the blocklist.

## Checklist

## Testing Performed

- [x] Green CI with `build-full-images` label.

### DIFF
--- a/kernel-modules/BLOCKLIST
+++ b/kernel-modules/BLOCKLIST
@@ -63,7 +63,6 @@
 5.15.0-2-cloud-amd64
 5.15.0-3-amd64
 5.15.0-3-cloud-amd64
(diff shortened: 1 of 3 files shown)


Stringy (approved):
LGTM!

PR #977, opened 2023-01-18
Cleanup driver builds

Molter73:
## Description

This PR removes a bunch of stuff from the driver builds that is no longer needed, including but not limited to:
- Dockerized build of drivers.
- Legacy patching system for sysdig based modules.
- Unused driver builder Dockerfiles.

Documentation for the driver builds has been updated with the new GHA based workflows.

## Checklist

## Testing Performed
- [x] Run CI with `no-cache` label.
- [x] Run CI with `build-legacy-probes` label.
- [x] Remove dockerized specific labels from GH.


docs/driver-builds.md
```diff
@@ -1,48 +1,42 @@
...
+
 ```
 </details>
 
+In order to build the required drivers, the `Drivers build` step will pull
+kernel bundles from GCP. The bundles are tarballs containing the required
+headers for each kernel version supported, these are created by our
+[crawler](https://github.com/stackrox/kernel-packer).
+
+Once building the drivers is done, 3 follow up steps will be triggered:
```
erthalion, line 36: It's a bit weird that the previous step was "download bundles", and the following is "once building is done". Maybe for completeness rearrange sections to be more consistent?

Molter73:
Integration tests in the last CI run failed for an unrelated known failure regarding ssh keys, we are working to solve this but the PR should be g2g

docs/driver-builds.md
Molter73: I've moved this paragraph to its own `Post-build jobs` section, hope this makes it a bit easier to follow.

PR #978, opened 2023-01-19
ROX-13628 Create an external representation of endpoints (conntracker)

ovalenti:
## Description

In Collector, an endpoint now represents a listening socket with a reference to the process making use of it. When an endpoint is published, however, the process information is stripped down to represent a the name and parameters of this process instead of the precise instance.

This as an impact on the life-cycle of endpoint objects "as seen from the outside". For instance, the same endpoint opened by several processes having the same attributes should appear as a single endpoint.

In order to implement this model, we propose to change the comparison operator used by the ConnTracker when building the normalized list of endpoints.

## Checklist

**Automated testing**
  - [x] Added unit tests
  - [x] Added integration tests

### DIFF
--- a/collector/lib/ConnTracker.cpp
+++ b/collector/lib/ConnTracker.cpp
@@ -40,6 +40,24 @@ static const NRadixTree private_networks_tree(PrivateNetworks());
 
 }  // namespace
 
+bool AdvertisedEndpointEquality::operator()(const ContainerEndpoint& lhs, const ContainerEndpoint& rhs) const {
+  if (bool(lhs.originator()) != bool(rhs.originator())) {
+    return false;
+  }
+
+  if (lhs.originator()) {
+    auto& lhs_originator = *lhs.originator();
+    auto& rhs_originator = *rhs.originator();
+
+    // Here is the real difference with the comparator in ContainerEndpointMap, which makes a deep compare of originators. We only focus on process-unique-key.
+    if ((lhs_originator.comm() != rhs_originator.comm()) || (lhs_originator.exe_path() != rhs_originator.exe_path()) || (lhs_originator.args() != rhs_originator.args())) {
+      return false;
+    }
+  }
+
+  return lhs.container() == rhs.container() && lhs.endpoint() == rhs.endpoint() && lhs.l4proto() == rhs.l4proto();
+}
+
 bool ContainsPrivateNetwork(Address::Family family, NRadixTree tree) {
   return tree.IsAnyIPNetSubset(family, private_networks_tree) || private_networks_tree.IsAnyIPNetSubset(family, tree);
 }
@@ -166,17 +184,13 @@ struct dont_filter {
   }
 };
 
-template <typename T, typename ProcessFn, typename FilterFn>
-UnorderedMap<T, ConnStatus> FetchState(UnorderedMap<T, ConnStatus>* state, bool clear_inactive,
-                                       const ProcessFn& process_fn, const FilterFn& filter_fn) {
+template <typename T, typename ProcessFn, typename FilterFn, typename E = std::equal_to<T>>
+UnorderedMap<T, ConnStatus, E> FetchState(UnorderedMap<T, ConnStatus>* state, bool clear_inactive,
+                                          const ProcessFn& process_fn, const FilterFn& filter_fn) {
   constexpr bool normalize = !std::is_same<ProcessFn, dont_normalize>::value;
   constexpr bool filter = !std::is_same<FilterFn, dont_filter>::value;
 
-  UnorderedMap<T, ConnStatus> fetched_state;
-
-  if (!clear_inactive && !normalize && !filter) {
-    return *state;
-  }
+  UnorderedMap<T, ConnStatus, E> fetched_state;
 
   for (auto it = state->begin(); it != state->end();) {
     const auto& entry = *it;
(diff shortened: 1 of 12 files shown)


collector/test/ConnTrackerTest.cpp
```diff
@@ -1246,91 +1246,149 @@ TEST(ConnTrackerTest, TestComputeDeltaAfterglowBenchmark2) {
...
 }
+/* Same endpoint, but opened by two processes having the same process-unique-key.
+   We expect that the endpoint is reported only once */
+TEST(ConnTrackerTest, TestSameEndpointSimilarProcess) {
+  int64_t activity_time1 = 1000;
+  int64_t activity_time2 = 2000;
+  Endpoint a(Address(192, 168, 0, 1), 80);
+  ConnectionTracker connectionTracker;
+  std::shared_ptr<IProcess> process1 = std::make_shared<FakeProcess>(2, "container", "comm", "exe1", "exe_path", "args");
+  std::shared_ptr<IProcess> process2 = std::make_shared<FakeProcess>(3, "container", "comm", "exe2", "exe_path", "args");
```
JoukoVirtanen, line 1311: Write a unit test where there are two processes with the same pid. This would help to show that what is being compared is the addresses of the shared pointers rather than their contents.

collector/lib/ConnTracker.cpp
```diff
@@ -40,6 +40,24 @@ static const NRadixTree private_networks_tree(PrivateNetworks());
 
 }  // namespace
 
+bool AdvertisedEndpointEquality::operator()(const ContainerEndpoint& lhs, const ContainerEndpoint& rhs) const {
```
JoukoVirtanen, line 43: Perhaps there should be unit tests that tests this operator directly.

collector/lib/ConnTracker.h
```diff
@@ -89,8 +89,16 @@ class ConnStatus {
   uint64_t data_;
 };
 
+/* Adverstised Endpoint representation comparison operator. It matches when two endpoints are undistinguishable
```
JoukoVirtanen, line 92:
```suggestion
/* Advertised Endpoint representation comparison operator. It matches when two endpoints are indistinguishable
```

collector/lib/ConnTracker.cpp
```diff
@@ -40,6 +40,24 @@ static const NRadixTree private_networks_tree(PrivateNetworks());
...
+  if (bool(lhs.originator()) != bool(rhs.originator())) {
+    return false;
+  }
+
+  if (lhs.originator()) {
+    auto& lhs_originator = *lhs.originator();
+    auto& rhs_originator = *rhs.originator();
+
+    // Here is the real difference with the comparator in ContainerEndpointMap, which makes a deep compare of originators. We only focus on process-unique-key.
```
JoukoVirtanen, line 52: Process unique key is not defined in collector, so maybe mention where it is defined. However this does present a problem as these comments couple us more strongly to sensor and central, and we do want to be able to run collector as a stand alone and give people the option to have something other than sensor ingest messages from collector.

JoukoVirtanen: Also maybe explain this at the top of the operator rather than here.

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -208,7 +208,7 @@ void NetworkStatusNotifier::RunSingle(IDuplexClientWriter<sensor::NetworkConnect
   WaitUntilWriterStarted(writer, 10);
 
   ConnMap old_conn_state;
-  ContainerEndpointMap old_cep_state;
+  AdvertisedEndpointMap old_cep_state;
```
JoukoVirtanen, line 211: Perhaps put a comment here explaining why an AdvertisedEndpointMap is used here.

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -220,7 +220,7 @@ void NetworkStatusNotifier::RunSingle(IDuplexClientWriter<sensor::NetworkConnect
 
     const sensor::NetworkConnectionInfoMessage* msg;
     ConnMap new_conn_state;
-    ContainerEndpointMap new_cep_state;
+    AdvertisedEndpointMap new_cep_state;
```
JoukoVirtanen, line 223: Same as above.

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -252,7 +252,7 @@ void NetworkStatusNotifier::RunSingleAfterglow(IDuplexClientWriter<sensor::Netwo
   WaitUntilWriterStarted(writer, 10);
 
   ConnMap old_conn_state;
-  ContainerEndpointMap old_cep_state;
+  AdvertisedEndpointMap old_cep_state;
```
JoukoVirtanen, line 255: Same as above.

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -265,7 +265,7 @@ void NetworkStatusNotifier::RunSingleAfterglow(IDuplexClientWriter<sensor::Netwo
 
     int64_t time_micros = NowMicros();
     const sensor::NetworkConnectionInfoMessage* msg;
-    ContainerEndpointMap new_cep_state;
+    AdvertisedEndpointMap new_cep_state;
```
JoukoVirtanen, line 268: Same as above.

collector/lib/ConnTracker.cpp
JoukoVirtanen: Write a comment in the current comparison operator for container endpoints pointing out that it is only comparing the addresses.

collector/lib/NetworkStatusNotifier.cpp
ovalenti: This is part of the API of ConnectionTracker API (the return type of the method used to retrieve the sets). I don't think that the API user has to know about internals.

collector/lib/ConnTracker.cpp
ovalenti: I added a unit test targeting specifically the comparator, good idea.
Added a comment in the operator==() of endpoints to explain that using pointer comparison is a perf trick; and is currently possible thanks to guaranties provided by ProcessStore.

collector/test/ConnTrackerTest.cpp
ovalenti: Unit tests are useful for their expression of functional expectations; I don't think that they should fail when some internal implementation details are changed. Although, we might want to make sure that the ProcessStore always returns the same pointer for a given Process object.

collector/lib/ConnTracker.cpp
ovalenti: This is a good point ! I updated the explanations to refer to storage::NetworkProcessUniqueKey instead of the blurry Sensor concept.

collector/lib/NetworkConnection.h
```diff
@@ -348,16 +348,19 @@ size_t Hash(const L4ProtoPortPair& pp);
...
   const Endpoint& endpoint() const { return endpoint_; }
   const L4Proto l4proto() const { return l4proto_; }
-  const std::shared_ptr<Process> originator() const { return originator_; }
+  const std::shared_ptr<IProcess> originator() const { return originator_; }
 
   bool operator==(const ContainerEndpoint& other) const {
     return container_ == other.container_ && endpoint_ == other.endpoint_ && l4proto_ == other.l4proto_ &&
+           /* Warning, the following equality uses the assumption that there can be only one Process
+              object per process in the system, and compares pointers directly for performance reasons.
+              This unicity is guaranteed by the ProcessStore. */
```
JoukoVirtanen, line 363: I think you mean uniqueness and not unicity.

collector/lib/ConnTracker.cpp
JoukoVirtanen: Great. I did add a correction to the comment.

collector/test/ConnTrackerTest.cpp
JoukoVirtanen: Okay

JoukoVirtanen (approved):
LGTM, but probably get another person to approve the changes. Also I might want to improve the integration test.

PR #981, opened 2023-01-20
Multi-arch support packages

Molter73:
## Description

Change the generated `index.html` file to contain links for `s390x` and `ppc64le` support-packages. Some additional changes were needed because of how this file used to be generated. 

Previously the only built support-package (for x86 architecture) was built and the index file would be immediately generated. Since we now have support packages being downloaded from downstream once a day, the index file needs to be updated with both the x86 support packages built in this repo and the downstream built s390x and ppc64le packages. In order to achieve this, we now push the support packages directly to GCP alongside a metadata file containing the name for the latest package being pushed and it's "last modified date", we then proceed to check which files exist in GCP and create the index.html file based on that information. This means that if we are building the x86 support packages, the previously uploaded s390x and ppc64le packages will be used for the new index and it will work the other way around when running the downstream steps.

## Checklist

## Testing Performed
- [x] Test both CPaaS and regular support-packages are properly added to the index.html file


Molter73:
This is what the support-package page will look like once multi-arch is added
image

Molter73:
The full collector build in the CPaaS workflow is failing because of missing downstream drivers, this unrelated to these changes.

Other than that, we still need to figure whether the buckets that are being used to push the drivers and support packages are OK, but since that can be adjusted under `.github/workflows/init.yml`, we can probably go ahead and start reviewing the PR.

.github/actions/support-package-index/action.yml
```diff
@@ -0,0 +1,45 @@
...
+    description: |
+      Path to generated artifacts
+
+runs:
+  using: composite
+  steps:
+  - name: Get latest support-packages
+    shell: bash
+    run: |
+      ${{ github.action_path }}/latest-support-packages.sh \
```
JoukoVirtanen, line 26: Maybe change the name of the script to get-latest-support-packages.sh

.github/actions/support-package-index/latest-support-packages.sh
```diff
@@ -0,0 +1,34 @@
...
+GCP_BUCKET=$3
+
+for arch in x86_64 s390x ppc64le; do
+    for version_path in "${METADATA_PATH}/module-versions"/*; do
+        version="$(basename "${version_path}")"
+        output_dir="${OUTPUT_PATH}/${arch}/${version}"
+
+        mkdir -p "${output_dir}"
+        if ! gsutil cp "gs://${GCP_BUCKET}/${arch}/${version}/metadata.json" "${output_dir}/metadata.json"; then
+            continue
```
JoukoVirtanen, line 16: Maybe put a logging statement here that metadata.json is missing.

.github/actions/support-package-metadata-json/metadata.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+
+import json
+import os
+from datetime import datetime
+
+
+def main():
+    support_pkg_dir = os.environ['SUPPORT_PACKAGE_TMP_DIR']
```
JoukoVirtanen, line 9: Check if the environment variable exists.

.github/actions/support-package-metadata-json/metadata.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
```
JoukoVirtanen, line 1: It might be a bit confusing to have scripts having to do with support packages in .github/actions, .github/support-packages, and kernel-modules/support-packages. Perhaps they should be put into one place or there should be README.md files that explain the different locations for scripts.

.github/actions/support-package-metadata-json/metadata.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def main():
+    support_pkg_dir = os.environ['SUPPORT_PACKAGE_TMP_DIR']
+
+    for mod_ver in os.listdir(support_pkg_dir):
+        version_dir = os.path.join(support_pkg_dir, mod_ver)
+
+        with open(os.path.join(version_dir, 'latest'), 'r') as f:
+            latest_file = f.read().splitlines()[0]
```
JoukoVirtanen, line 15: [nit] 
```suggestion
            latest_file = f.read().readline()
```

.github/actions/support-package-metadata/action.yml
```diff
@@ -0,0 +1,39 @@
...
+      ${{ github.workspace }}/kernel-modules/support-packages/01-collector-to-rox-version-map.py \
+      ${{ inputs.released-versions-file }} \
+      ${{ inputs.metadata-directory }}
+
+  - name: Create driver version metadata
+    shell: bash
+    env:
+      BRANCH_NAME: ${{ inputs.branch-name }}
+    run: |
+      ${{ github.action_path }}/create-metadata-dir.sh
```
JoukoVirtanen, line 33: If 02-fetch-collectors-metadata.sh is no longer being used, it should probably be deleted and other associated changes should be made. I.e update README and run-all.sh in kernel-modules/support-packages.

.github/workflows/cpaas-sync-drivers.yml
```diff
@@ -73,14 +100,89 @@ jobs:
               gsutil -m cp -n -I "gs://${{ inputs.drivers-bucket }}/${{ matrix.platform }}/$(basename "${driver_version_dir}")/"
           done
 
+      # x86 support packages will not be uploaded to GCP.
+      # This is to prevent mixing it up with the upstream built packages.
```
JoukoVirtanen, line 104: Consider finding a way of skipping steps if matrix.platform != 'x86_64'

.github/workflows/cpaas-sync-drivers.yml
```diff
@@ -73,14 +100,89 @@ jobs:
...
+      - name: Generate 'latest' files
```
JoukoVirtanen, line 105:
```suggestion
      - name: Generate 'latest' and sha files
```

.github/workflows/cpaas-sync-drivers.yml
```diff
@@ -73,14 +100,89 @@ jobs:
...
+        if: matrix.platform != 'x86_64'
```
JoukoVirtanen, line 106: Move to a script

.github/actions/support-package-metadata-json/metadata.py
Molter73: With these changes, there are no more scripts under `.github/support-packages` and the scripts in the actions directory are there because they are specifically used with the action they are part of. I agree we need to make a better job of organizing our scripts, but IDK if this is the right PR to address this.

.github/actions/support-package-metadata-json/metadata.py
Molter73: Strings don't have a `readline()` method, I had this in a [previous commit](https://github.com/stackrox/collector/pull/981/commits/1aa3b1f9f1b1ee1e399b218cdc7566ad535410ae):
```py
latest_file = f.readline()
```

Problem is `readline()` from a file keeps the line end character, so I would still need to remove that. Splitting the line and keeping the first part seemed the cleanest and most reliable method, WDYT?

.github/workflows/cpaas-sync-drivers.yml
Molter73: I've put some thought into this one and I could try to move all the steps to a separate workflow, then make only that workflow run on the non x86 case, but I would also need to either:
- Upload the support-packages as an artifact for the follow up workflow to download it.
- Directly use the images from a separate workflow for extracting the support-packages directly.

Both seem to me as extra complexity just to avoid putting `if matrix.platform != 'x86_64'` in 3 steps. I think I'd prefer to keep it as is, but if you feel strongly about it, I can change it.

.github/workflows/cpaas-sync-drivers.yml
Molter73: Went with `Generate 'latest' and checksum files`, hope that's fine 🙂

Molter73:
@robbycochran this is how I got CI configured to push to GCP once the PR is merged:
```
drivers
 |-> upstream       -> collector-modules-osci/
 |-> downstream 
      |->   x86_64  -> collector-modules-osci/cpaas/x86_64/
      |->    s390x  -> collector-modules-osci/cpaas/s390x/
      |->  ppc64le  -> collector-modules-osci/cpaas/ppc64le/

support-packages
 |-> upstream       -> sr-roxc/collector/support-packages/x86_64/
 |-> downstream
 |    |->   x86_64  -> *Not pushed*
 |    |->    s390x  -> sr-roxc/collector/support-packages/s390x/
 |    |->  ppc64le  -> sr-roxc/collector/support-packages/ppc64le/
 |-> index.html     -> sr-roxc/collector/support-packages/index.html

support-packages Public
 |-> upstream       -> collector-support-public/offline/v1/support-packages/x86_64/
 |-> downstream
 |    |->   x86_64  -> *Not pushed*
 |    |->    s390x  -> collector-support-public/offline/v1/support-packages/s390x/
 |    |->  ppc64le  -> collector-support-public/offline/v1/support-packages/ppc64le/
 |-> index.html     -> collector-support-public/offline/v1/support-packages/index.html
```

We would need to mirror the drivers in `collector-modules-osci/`, `collector-modules-osci/cpaas/s390x/` and `collector-modules-osci/cpaas/ppc64le/` to the bucket collector downloads directly at runtime for them to be made accessible with no further changes to the collector code. I realize mirroring just the files in `collector-modules-osci/` and not the ones in `collector-modules-osci/cpaas/x86_64/` might be tricky, so if you'd like me to set them up in separate buckets, please let me know.

In addition to the drivers, we need to make sure the proxy that server support-packages from https://cdn.stackrox.io/collector/support-packages/index.html allows us to reach files in subdirectories to the one we are currently using, since the support-packages are uploaded that way based on architecture.

robbycochran (approved):

.github/actions/support-package-metadata-json/metadata.py
JoukoVirtanen: Okay

.github/actions/support-package-metadata-json/metadata.py
JoukoVirtanen: I guess it will report the error either way.

.github/actions/support-package-metadata-json/metadata.py
JoukoVirtanen: It doesn't very important. I meant to write `latest_file = f.readline()`. You could try `latest_file = f.readline().strip()`. It does avoid reading the entire file. In this case it is not important, but in some other case it might be important if the file is large.

.github/actions/support-package-metadata/action.yml
JoukoVirtanen: I created `run-all.sh` when working with OSCI, but the intention was that it could be run outside of OSCI. The intention was that it could be run manually or in another CI. I am not requiring that GHA use it. I think we did discuss that possibility earlier and decided not to do that. This issue can be dealt with in another PR.

.github/workflows/cpaas-sync-drivers.yml
JoukoVirtanen: I don't feel strongly about it.

.github/workflows/cpaas-sync-drivers.yml
JoukoVirtanen: That's fine.

JoukoVirtanen (approved):
LGTM! There are one or two minor things that you could do. I resolved some conversations, but it might be worth it for you to take a look at those still.

.github/actions/support-package-metadata-json/metadata.py
Molter73: Alright, `f.readline().strip()` is easier to read, I'll give it a try.

PR #982, opened 2023-01-24
Fix setting of tag names when running from a schedule event

Molter73:
## Description

When running from a `schedule` event, such as a the jobs related to CPaaS, the tags for the collector builder are not being set. This PR makes it so the same rules as push events are applied to schedule events.

## Checklist

## Testing Performed

Not much to test, schedule events can only be triggered once merged into the default branch.

### DIFF
--- a/.github/workflows/collector-slim.yml
+++ b/.github/workflows/collector-slim.yml
@@ -30,14 +30,14 @@ jobs:
           submodules: true
 
       - name: Checks for main and release branches
-        if: ${{ github.event_name == 'push' }}
+        if: github.event_name != 'pull_request'
         run: |
           echo 'BUILD_BUILDER_IMAGE=true' >> "$GITHUB_ENV"
           echo "COLLECTOR_BUILDER_TAG=${DEFAULT_BUILDER_TAG}" >> "$GITHUB_ENV"
 
       - name: PR checks
         id: pr-checks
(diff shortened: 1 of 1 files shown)


.github/workflows/cpaas.yml
```diff
@@ -57,3 +57,25 @@ jobs:
...
+    - build-collector-slim
+    - build-collector-full
+    - integration-tests
+    steps:
+      - name: Slack notification
+        uses: rtCamp/action-slack-notify@v2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_COLLECTOR_NOTIFICATIONS_WEBHOOK }}
+          SLACK_CHANNEL: collector-notifications
+          SLACK_COLOR: ${{ job.status }}
```
Stringy, line 76: I think this will be the wrong colour, because this job (`notify`) has not failed. I think in this case it's better to hard-code it to 'failed' (IIRC) since we know we're only triggering when the `needs` have failed

Stringy (approved):
One minor point about notification colours but otherwise LGTM!

.github/workflows/cpaas.yml
Molter73: Changed it to `failure` which is the one I found for the integration tests notification.

PR #983, opened 2023-01-24
Add gofmt to the pre-commit actions

ovalenti:
## Description

As a side-effect, this will add this step to the lint GHA workflow.

## Checklist
- [x] Introduce a change which breaks the formatting
- [x] Revert the change and verify that everything is green

### DIFF
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -1,6 +1,5 @@
 package integrationtests
 
-
 import (
 	"testing"
 


Stringy:
As a bonus, it might be worth running gofmt over the integration tests and fix the many formatting problems we already carry there

ovalenti:
> As a bonus, it might be worth running gofmt over the integration tests and fix the many formatting problems we already carry there

@Stringy I have been running `gofmt` on master, and it didn't return any problem under `./integration-tests/`. Would it be possible that those have been taken care of already ?

JoukoVirtanen (approved):
LGTM!

PR #985, opened 2023-01-25
Jv throttle logging in procfs scraper

JoukoVirtanen:
## Description

Makes it so that the logs in ProcfsScraper are throttled. This is part of https://issues.redhat.com/browse/ROX-13962.

Future work will include adding metrics associated with the current logging statements. Determining when a process is defunct, and adding metrics for them. Reporting defunct processes in an API and the UI.

## Checklist
- [ ] Saw logs throttled in local testing

## Testing Performed

TODO(replace-me)

### DIFF
--- a/collector/lib/ProcfsScraper.cpp
+++ b/collector/lib/ProcfsScraper.cpp
@@ -129,7 +129,7 @@ class SocketInfo {
 bool GetSocketINodes(int dirfd, uint64_t pid, UnorderedSet<SocketInfo>* sock_inodes) {
   DirHandle fd_dir = FDHandle(openat(dirfd, "fd", O_RDONLY));
   if (!fd_dir.valid()) {
-    CLOG(ERROR) << "could not open fd directory";
+    CLOG_THROTTLED(ERROR, std::chrono::seconds(1)) << "could not open fd directory";
     return false;
   }
 
@@ -427,7 +427,7 @@ bool ReadContainerConnections(const char* proc_path, std::shared_ptr<ProcessStor
                               std::vector<Connection>* connections, std::vector<ContainerEndpoint>* listen_endpoints) {
   DirHandle procdir = opendir(proc_path);
   if (!procdir.valid()) {
-    CLOG(ERROR) << "Could not open " << proc_path << ": " << StrError();
+    CLOG_THROTTLED(ERROR, std::chrono::seconds(1)) << "Could not open " << proc_path << ": " << StrError();
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


JoukoVirtanen:
The CLOG(DEBUG) on line 444 was left as is. If someone is doing debugging they probably want to see everything.

ovalenti (dismissed):
At first, I wondered about a performance hit, but since all the CLOG_THROTTLED calls are in ERROR code paths, there is no impact in the nominal case.

I think that this patch would solve the customer issue we had in the first place, so approving it. But I would like to get a consensus before merging. @erthalion

collector/lib/ProcfsScraper.cpp
```diff
@@ -129,7 +129,7 @@ class SocketInfo {
 bool GetSocketINodes(int dirfd, uint64_t pid, UnorderedSet<SocketInfo>* sock_inodes) {
   DirHandle fd_dir = FDHandle(openat(dirfd, "fd", O_RDONLY));
   if (!fd_dir.valid()) {
-    CLOG(ERROR) << "could not open fd directory";
+    CLOG_THROTTLED(ERROR, std::chrono::seconds(10)) << "could not open fd directory";
```
JoukoVirtanen, line 132: Set to 10 seconds as 1 second is too noisy.

collector/lib/Logging.h
```diff
@@ -127,11 +127,14 @@ class LogMessage {
...
 
-#define CLOG_THROTTLED_IF(cond, lvl, interval)                                          \
-  static std::chrono::steady_clock::time_point _clog_lastlog_##__LINE__;                \
-  if (collector::logging::CheckLogLevel(collector::logging::LogLevel::lvl) && (cond) && \
-      (std::chrono::steady_clock::now() - _clog_lastlog_##__LINE__ >= interval))        \
-  _clog_lastlog_##__LINE__ = std::chrono::steady_clock::now(),                          \
+#define CLOG_THROTTLED_IF(cond, lvl, interval)                                           \
+  static bool _clog_firstlog_##__LINE__ = true;                                          \
+  static std::chrono::steady_clock::time_point _clog_lastlog_##__LINE__;                 \
+  if (_clog_firstlog_##__LINE__ ||                                                       \
```
ovalenti, line 133: The `firstlog` boolean seems superfluous, indeed, `std::chrono::steady_clock::time_point` with default constructor represents the clocks epoch.

ref: 
https://en.cppreference.com/w/cpp/chrono/time_point/time_point

ovalenti (approved):
Looks good to me !

PR #989, opened 2023-01-30
Adds image-upload ansible playbook 

Stringy:
## Description

Adds and image-upload ansible playbook to synchronize development Collector image builds with the ansible inventory. Intended primarily as a development tool for those who do not wish to use an intermediary quay repository.

### DIFF
--- a/ansible/image-upload.yml
+++ b/ansible/image-upload.yml
@@ -1,0 +1,45 @@
+---
+
+- name: Push Collector Image to Hosts
+  vars_files:
+    - group_vars/all.yml
+  hosts: "platform_*:&job_id_{{ job_id }}"
+  strategy: free
+
+  tasks:
+    - name: Get Collector Image
+      shell: make -C "{{ collector_root }}" tag
+      register: collector_tag_result
+      delegate_to: localhost
+
+    - set_fact:
+        collector_tag: "{{ collector_tag_result.stdout | replace('\n', '') }}"
+
+    - name: Create tmp directory
+      tempfile:
+        state: directory
+      register: image_directory
+      delegate_to: localhost
+
+    - name: Save Collector Image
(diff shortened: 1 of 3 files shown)


ansible/image-upload.yml
```diff
@@ -0,0 +1,45 @@
...
+  vars_files:
+    - group_vars/all.yml
+  hosts: "platform_*:&job_id_{{ job_id }}"
+  strategy: free
+
+  tasks:
+    - name: Get Collector Image
+      shell: make -C "{{ collector_root }}" tag
+      register: collector_tag_result
+      delegate_to: localhost
```
Molter73, line 13: Could we instead control the image tag being used with an env variable? Sometimes you want to use an existing image you built a couple days back, but might not have the branch in the same state anymore.

Molter73 (commented):
Awesome! I ended up creating a registry under my quay.io account for this exact situation, having it automated will be nice.

ansible/image-upload.yml
Stringy: yeah that's a good idea. I'll make it default to the current tag if the env var is not provided

ansible/image-upload.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    - set_fact:
+        collector_tag: "{{ collector_tag_result.stdout | replace('\n', '') }}"
+
+    - set_fact:
+        collector_tag: "{{ lookup('env', 'COLLECTOR_TAG') | default(collector_tag) }}"
```
Molter73, line 19: Looks kinda weird that we are setting `collector_tag` twice, maybe we could save the filtered one to a different fact?

Molter73 (approved):
Just a minor comment on the same fact being set twice, but shouldn't be a blocker for this, so LGTM!

PR #994, opened 2023-02-01
Add a script for building drivers and sinsp-example

Molter73:
## Description

The script is meant to be run from inside a `collector-builder` container, so it should be portable enough to be run in multiple systems for testing. We also add 'cmake' to the driver builder containers, since this is a requirement to run the dev/build-drivers script, which is not great, but shouldn't make the image significantly bigger.

We also remove the 'scripts' directory, since it had a single unused script in it, I'd be happy to move this to a separate commit if needed.

## Checklist

## Testing Performed

- [x] Run CI with the `no-cache` label to make sure the builders are still working properly.


Stringy (approved):
Couple of minor comments but otherwise looks like an extremely useful tool! LGTM

utilities/sinsp-builder.sh
```diff
@@ -0,0 +1,90 @@
...
+# binary, which is useful for quickly testing driver changes.
+#
+# Run the collector-builder image with the following command:
+#   docker run --rm -it --entrypoint bash \
+#       --privileged \
+#       -v /var/run/docker.sock:/var/run/docker.sock \
+#       -v /sys:/sys:ro \
+#       -v /dev:/dev \
+#       -v /tmp:/tmp \
+#       quay.io/stackrox-io/collector-builder:cache
```
JoukoVirtanen, line 16: It is a bit annoying to copy and paste this and remove the #s. Perhaps put this in a script and mention the script here. Perhaps create a new directory within the utilities directory called sinsp-builder, put this script and the docker run script there. You could also move a lot of this into a readme file.

I am also thinking that the utilities directory could be reorganized. That doesn't have to be done in this PR. I think the other scripts in the utilities directory could be in their own directories with readme files, instead of having a docs directory.

utilities/sinsp-builder.sh
```diff
@@ -0,0 +1,90 @@
...
+#
+# If you are running on an immutable system, create a docker volume then add
+# the following arguments:
+#   -v <your-volume>:/tmp/collector -e DEV_SHARED_VOLUME=<your-volume>
+#
+# If you are lazy like me, you don't need to clone the collector repo before
+# executing, you can simply run this inside the container:
+#   curl https://raw.githubusercontent.com/stackrox/collector/master/utilities/sinsp-builder.sh | sh
```
JoukoVirtanen, line 24: If this is run before this PR is merged the correct command is `curl https://raw.githubusercontent.com/stackrox/collector/mauro-easy-debug-drivers/utilities/sinsp-builder.sh | sh`

JoukoVirtanen: This did build the sinsp-example binary, but it also exited with a failure 

```[ 97%] Building CXX object libsinsp/examples/CMakeFiles/sinsp-example.dir/test.cpp.o
[ 98%] Building CXX object libsinsp/examples/CMakeFiles/sinsp-example.dir/util.cpp.o
[100%] Linking CXX executable sinsp-example
make[3]: Leaving directory '/tmp/tmp.8hp0tSs4Yb'
[100%] Built target sinsp-example
make[2]: Leaving directory '/tmp/tmp.8hp0tSs4Yb'
make[1]: Leaving directory '/tmp/tmp.8hp0tSs4Yb'
make: Leaving directory '/tmp/tmp.8hp0tSs4Yb'
gzip: /tmp/collector/kernel-modules/container/kernel-modules/*.gz: No such file or directory
[root@999f0c8093a0 install-tmp]# echo $?
1
```

Molter73: It errored out when trying to decompress the kernel drivers, could you double check if those were properly built?

Stringy: I had this problem - it's because the images used do not contain the changes in this PR (addition of cmake, and adjustment to the `drivers` make target) This won't be a problem once this PR is merged, and the workaround for testing now is to rebuild the images locally with the changes

utilities/sinsp-builder.sh
Molter73: > It is a bit annoying to copy and paste this and remove the #s. Perhaps put this in a script and mention the script here. Perhaps create a new directory within the utilities directory called sinsp-builder, put this script and the docker run script there. You could also move a lot of this into a readme file.

I really don't want to move this to a script, depending on whether you are running on an immutable system or not the command is different and we already have a way to run the builder image using `make start-dev` for local purposes. This script is meant to slightly simplify the process of building this binary and the driver when testing in remote systems were you might not have the repo cloned, which means you will need to put some elbow grease to get it working anyways, it's just not as much elbow grease as it would be if you need to do all of this on your own.

I also don't mean for this to become an entire framework for us to run the `sinsp-example` binary, so splitting it into its own directory, adding a readme, etc. it all feels like overkill to me.

> I am also thinking that the utilities directory could be reorganized. That doesn't have to be done in this PR. I think the other scripts in the utilities directory could be in their own directories with readme files, instead of having a docs directory.

We can discuss this with the broader team, I agree the state of the utilities directory is not great, but we also kinda use it as a place to drop scripts that might be useful on their own. Keeping separate README files for each script we might come up with seems like a lot of work to put together and maintain. I wrote every possible nuisance I could come up with into the script because that way people don't need to go looking for it on a separate file and, if someone makes a change to the script, it will be pretty much obvious that you'll need to make some adjustments to the "docs" that is that comment.

utilities/sinsp-builder.sh
```diff
@@ -0,0 +1,90 @@
...
+# `COLLECTOR_BRANCH`, otherwise, master will be used.
+
+# Install docker-cli, needed for DinD
+dnf config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
+dnf install -y docker-ce-cli
+
+# Buildkit misbehaves when running DinD sometimes
+export DOCKER_BUILDKIT=0
+
+if [[ "${COLLECTOR_DIR:-}" == "" ]]; then
```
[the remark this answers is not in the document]
Molter73: I'll add a warning message here and at the end, because cmake and make generate so much output it could be easily missed.

PR #996, opened 2023-02-03
Adds zone looping to VM creation

Stringy:
## Description

To avoid the case where a zone is unable to accept our VM creation request, this PR changes the ansible logic to loop over a set of available zones, using the first one that succeeds. 

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed
Tested locally, but unable to truly test when a zone fails. I have verified that if the previous execution fails it will continue to try the next zone.


Molter73 (approved):
Gotta say I'm a little disappointed the google module for GCP VMs only allows the zone to be a string, I can't believe we would be the only people interested in creating a VM regardless of which zone it gets created in.

That said, LGTM!

PR #999, opened 2023-02-07
update apollo-ci image to 0.3.56

RTann:
## Description

I believe I updated all references to `apollo-ci`.

This bump Go version to 1.19.4. This seems to only affect the integrations tests. I also updated the `go.mod` to 1.19 and updated two dependencies since I was already touching the `go.mod`.

## Checklist
- ~[ ] Updated documentation accordingly~

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

The existing tests suffice

## Testing Performed

CI


RTann:
I'm not 100% sure how to test this in the `openshift/release` repo because there only seem to be configurations to the tagged versions of collector instead of for the master/main branch

Molter73:
Hey @RTann, thanks for the PR. 

We are no longer using OSCI, we will be removing most of that directory soon (if we get the time to do so). And now that you opened the PR, I realize we are also not using the apollo-ci images, since the GHA builders have everything we need directly installed in them.

We are also planning to replace our integration tests to use pytest, so I'm not sure if we will bother updating the go version used for those. If it was a production thing, sure, but for testing, as long as it works, I don't really see a difference.

RTann:
@Molter73 sounds good to me!

PR #1000, opened 2023-02-07
Remove the .openshift-ci directory

Molter73:
## Description

Pretty self-explanatory, we've moved on from OSCI and we are no longer using most of the contents of the dedicated `.openshift-ci` directory, so we might as well do some spring cleaning in advance.

## Checklist

## Testing Performed

- [x] Run with `run-cpaas-steps` label.
- [x] Run with `no-cache` label.


JoukoVirtanen:
There are still references to openshift-ci in the Makefile

Molter73:
You are right! Good catch!

Removed those as well.

JoukoVirtanen (approved):
LGTM!

PR #1002, opened 2023-02-08
Fix checksum file URLs for support-packages

Molter73:
## Description

Checksum URLs are missing a `/`, so we add it.

## Checklist

## Testing Performed

- [x] Run with `test-support-packages` label and check the URL for checksums is correct in the generated index.html file.

### DIFF
--- a/kernel-modules/support-packages/05-create-index.py
+++ b/kernel-modules/support-packages/05-create-index.py
@@ -42,11 +42,11 @@ class SupportPackage(object):
 
     @property
     def checksum_url(self):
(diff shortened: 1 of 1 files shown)


Stringy (approved):
These changes look good - but I have to wonder how it ever worked? I assume the metadata has changed in some other PR?

Molter73:
@Stringy is changed in #981 and I only caught it after I merged the PR, sorry I didn't mention it as part of the description.

Molter73:
Merging without the final test done, since this shouldn't affect it.

PR #1005, opened 2023-02-10
Remove the dynamic configuration of base image for collector

Molter73:
## Description

This PR removes the possibility to dynamically change the base image used for the collector image. Originally this allowed us to have a UBI and Ubuntu based images, we then dropped Ubuntu images and started using this to build a development image based on centos stream. Because the devel and prod images are pretty close to each other, I believe it is more beneficial for us to have the image set to the production one so that dependabot can bump it on its own.

The devel image will now be based on ubi-minimal, which means the available packages are somewhat limited, but it is still possible to directly install gdb on it, which is the most important thing we want in it IMO.

## Checklist

## Testing Performed

- [x] Manually built the `-dev` collector image successfully.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,9 +73,6 @@ image: collector unittest container-dockerfile
 image-dev: collector unittest container-dockerfile
 	make -C collector txt-files
 	docker build --build-arg collector_version="$(COLLECTOR_TAG)" \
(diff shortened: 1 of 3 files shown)


JoukoVirtanen (approved):
LGTM!

PR #1010, opened 2023-02-14
Fixes build caching in collector build

Stringy:
## Description

The current make target uses individual new calls to make for each step of the build, which means the artifact `container/bin/collector` is incorrectly assumed to have built even when the source files change. Making each step a dependency of the build target means that their dependencies are used as dependencies of `container/bin/collector` and any source changes will trigger a rebuild of those files.

### DIFF
--- a/collector/Makefile
+++ b/collector/Makefile
@@ -13,12 +13,14 @@ HDRS := $(wildcard lib/*.h) $(shell find falcosecurity-libs/userspace -name '*.h
 
 SRCS := $(wildcard lib/*.cpp) collector.cpp
 
+COLLECTOR_BUILD_DEPS := $(HDRS) $(SRCS) $(shell find falcosecurity-libs -name '*.h' -o -name '*.cpp' -o -name '*.c')
+
 .SUFFIXES:
 
 .PHONY: pre-build
 pre-build: txt-files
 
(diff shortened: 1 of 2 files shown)


collector/Makefile
```diff
@@ -39,10 +39,7 @@ post-build:
 	cp -r "$(CMAKE_DIR)"/THIRD_PARTY_NOTICES/* container/THIRD_PARTY_NOTICES/
 	cp "$(LIBSINSP_BIN_DIR)/libsinsp-wrapper.so" container/libs/libsinsp-wrapper.so
 
-container/bin/collector:
-	$(MAKE) pre-build
-	$(MAKE) cmake-build/collector
-	$(MAKE) post-build
+container/bin/collector: pre-build cmake-build/collector post-build
```
ovalenti, line 42: If we depend on the sequential build of the dependencies, we may want to add a `.NOTPARALLEL` statement to make sure that the behavior is correct even with `-j X`

https://www.gnu.org/software/make/manual/html_node/Parallel-Disable.html

Stringy: hmm excellent point, I'll add that

ovalenti: It's even better this way. Out of curiosity, I wonder whether we could directly have
```
container/bin/collector:  cmake-build/collector
...
```

Stringy:
Github bot is very impressively displaying benchmarking data before the benchmarks have finished running... I'll take a look.

erthalion (approved):
Thanks, looks good. I have only one small suggestion about making "collector" target a dependency of "unittest".

PR #1021, opened 2023-02-20
Testing Falco Update 2023-02-15

Stringy:
## Description

Testing Falco updates


.github/workflows/init.yml
```diff
@@ -76,8 +80,17 @@ jobs:
...
+          COLLECTOR_TEST_TAG="$COLLECTOR_TAG"
+          if [[ '${{ contains(github.event.pull_request.labels.*.name, 'build-full-images') }}' == 'true' ]]; then
+            COLLECTOR_TEST_TAG="$COLLECTOR_TAG-latest"
+          fi
+
+          {
+            echo "collector-tag=${COLLECTOR_TAG}"
+            echo "collector-image=quay.io/stackrox-io/collector:${COLLECTOR_TAG}"
+            echo "collector-test-tag=${COLLECTOR_TEST_TAG}"
+          } >> "$GITHUB_OUTPUT"
```
Molter73, line 92: Why do we need this tag? If the `build-full-images` label is set CI builds the full `quay.io/stackrox-io/collector:${COLLECTOR_TAG}` and if it is not set the `-slim` variation gets re-tagged to this same image, that makes it so the integration tests can always use the same image to run in either case.

See:
- https://github.com/stackrox/collector/blob/a61c6aed1ff3178b5d55d282ff4e139a94851180/.github/workflows/collector-full.yml#L117-L160
- https://github.com/stackrox/collector/blob/a61c6aed1ff3178b5d55d282ff4e139a94851180/.github/workflows/collector-full.yml#L34-L36

builder/Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN dnf -y update \
         bison \
         ca-certificates \
         clang \
-        cmake \
+        cmake-3.18.2-9.el8 \
```
Molter73, line 18: Looks like the default version for cmake on centos stream8 is `3.20.2`, why do we need to downgrade it? Maybe we could add a comment on why this is needed and whether there's something we could do to go back to using the latest version?

Also, we need to double check the downstream build.

Molter73 (commented):
Just a couple minor comments, otherwise LGTM!

.github/workflows/init.yml
Stringy: In my testing this was not working and the `quay.io/stackrox-io/collector:${COLLECTOR_TAG}` image did not contain the drivers. I'll investigate further.

builder/Dockerfile
Stringy: This is an awkward one because that version of cmake was segfaulting when configuring the project, but pinning to the previous available version (as in this change) prevented the segfault.

Molter73: That is... Not good.

We should probably not sleep on this and take a closer look at what's going on, but for now we can probably go ahead with the merge.

PR #1025, opened 2023-02-22
Cleanup remaining OSCI code from kernel-splitter.py

Molter73:
## Description

Minor cleanup, removes some code that was left for compatibility with OSCI.

## Checklist

## Testing Performed

- [x] Run with `no-cache` label.


ovalenti (approved):
lgtm !

PR #1036, opened 2023-03-01
Add symlink to gcc10-gcc

Molter73:
## Description

Symlink `gcc` to `gcc10-gcc` so that amazon linux kernel modules can be compiled.

## Checklist

## Testing Performed

CI should be enough.


Molter73:
After adding the initial `gcc10-gcc` symlink, I had to add `ld.bfd` and `objdump` links as well. I think this should be fine, kernel modules are now building but we don't really have an easy way to test them in CI. I'll try to see if I can create a VM that uses one of these kernels and run at least basic tests to see nothing major breaks.

Molter73:
I've manually tested the built kernel modules on a VM in my local machine.

Stringy (approved):

PR #1038, opened 2023-03-03
Bump standard to C++17 and drop StringView

Molter73:
## Description

Bumping C++ standard to C++17, since it has been widely available for some time now. This should allow us to modernize and cleanup our code base a bit (even if we don't do it all at once), first step in this is to drop the custom `StringView` class in favor of `std::string_view`

## Checklist

## Testing Performed

CI should be enough and collector version should be bump in the main repo ASAP after merge in order to get more testing on it.


Stringy (approved):
LGTM!

PR #1039, opened 2023-03-03
Add a BTF probes ADR

erthalion:
## Description

A new proposal to capture most of the discussions about BTF support for Collector. There will be another one for the probe delivery mechanism.


Stringy (approved):
Looks good to me!

PR #1047, opened 2023-03-08
Pipe scripts in integration tests

ovalenti:
## Description

Up to now, integration tests have been relying on customized images containing shell scripts dedicated to a particular test. This PR provides with a way to "inject" a shell script into a generic image at runtime.

This brings the following advantages:
- no need for maintaining customized images
- updating the test script can be achieved in the test source directly

This mechanism could be extended fairly easily to Python scripts.

## Checklist


integration-tests/suites/common/executor.go
```diff
@@ -20,6 +21,7 @@ type Executor interface {
 	IsContainerRunning(image string) (bool, error)
 	ExitCode(container string) (int, error)
 	Exec(args ...string) (string, error)
+	ExecWithStdin(pippedContent string, args ...string) (string, error)
```
Stringy, line 24: [nit] typo - should be `pipedContent`

integration-tests/suites/common/executor.go
```diff
@@ -182,6 +184,22 @@ func (e *executor) RunCommand(cmd *exec.Cmd) (string, error) {
...
+func (e *executor) ExecWithStdin(pippedContent string, args ...string) (res string, err error) {
+	cmd := e.builder.ExecCommand(args...)
+
+	stdin, err := cmd.StdinPipe()
+	if err != nil {
+		return "", err
+	}
+
+	go func() {
+		defer stdin.Close()
```
Stringy, line 196: How does ownership work in this case? Is `stdin` copied into this goroutine and is therefore safe if the "outer" function returns?

integration-tests/suites/base.go
```diff
@@ -186,6 +186,12 @@ func (s *IntegrationTestSuiteBase) execContainer(containerName string, command [
 	return s.executor.Exec(cmd...)
 }
 
+func (s *IntegrationTestSuiteBase) execContainerShellScript(containerName string, script string, args ...string) (string, error) {
+	cmd := []string{"docker", "exec", "-i", containerName, "/bin/sh", "-s"}
```
Stringy, line 190: With the new podman support on master this will need to be:

```suggestion
	cmd := []string{common.RuntimeCommand, "exec", "-i", containerName, "/bin/sh", "-s"}
```

integration-tests/suites/common/executor.go
```diff
@@ -182,6 +184,22 @@ func (e *executor) RunCommand(cmd *exec.Cmd) (string, error) {
 	return trimmed, err
 }
 
+func (e *executor) ExecWithStdin(pippedContent string, args ...string) (res string, err error) {
+	cmd := e.builder.ExecCommand(args...)
```
Stringy, line 188: we may need to prepend `sudo` to `args` this if `common.RuntimeAsRoot` is true

Stringy (changes requested):
Looks like a cool feature! It's worth rebasing though, since there are some changes to the integration tests that affect this PR (the podman support, primarily)

integration-tests/suites/base.go
JoukoVirtanen, line 190: Do we want to enforce usage of /bin/sh here? Would be nice to have the option to use /bin/bash or others.

integration-tests/suites/repeated_network_flow.go
```diff
@@ -88,7 +88,28 @@ func (s *RepeatedNetworkFlowTestSuite) SetupSuite() {
 	numIter := strconv.Itoa(s.NumIter)
 	sleepBetweenCurlTime := strconv.Itoa(s.SleepBetweenCurlTime)
 	sleepBetweenIterations := strconv.Itoa(s.SleepBetweenIterations)
-	_, err = s.execContainer("nginx-curl", []string{"/usr/bin/schedule-curls.sh", numMetaIter, numIter, sleepBetweenCurlTime, sleepBetweenIterations, serverAddress})
+	_, err = s.execContainerShellScript("nginx-curl", `
+		#!/usr/bin/env sh
```
JoukoVirtanen, line 92: Maybe save this to a variable. Also I don't know if this is the best place for it. Maybe it is.

integration-tests/suites/repeated_network_flow.go
```diff
@@ -88,7 +88,28 @@ func (s *RepeatedNetworkFlowTestSuite) SetupSuite() {
...
+
+		while [ "$i" -lt "$num_meta_iter" ]; do
+		    while [ "$j" -lt "$num_iter" ]; do
+		        curl "$url"
+		        sleep "$sleep_between_curl_time"
+		        j=$((j + 1))
+		    done
+		    sleep "$sleep_between_iterations"
+		    i=$((i + 1))
+		done`,
```
JoukoVirtanen, line 111: A blank line here would be good so that it is easier to see where the script ends.

integration-tests/suites/repeated_network_flow.go
```diff
@@ -88,7 +88,28 @@ func (s *RepeatedNetworkFlowTestSuite) SetupSuite() {
...
+	_, err = s.execContainerShellScript("nginx-curl", `
+		#!/usr/bin/env sh
+
+		num_meta_iter=$1
+		num_iter=$2
+		sleep_between_curl_time=$3
+		sleep_between_iterations=$4
+		url=$5
+
+		i=0
```
JoukoVirtanen, line 100: Are there other scripts that could be moved into the integration tests in the same way. If so create tickets to do that and add links to them in this PR.

integration-tests/suites/common/executor.go
ovalenti: My understanding is that the anonymous function is a usual closure, and shares stdin with the outer scope. Things are even simpler here since both the lambda and the defer call happen before the outer function returns.

integration-tests/suites/base.go
ovalenti: You are right, however, I have no idea how to implement that option. Would you have suggestions ?

integration-tests/suites/repeated_network_flow.go
ovalenti: I don't see other tests to convert, except ProcessesListening on ports maybe, by piping the c code into the --run option of https://bellard.org/tcc/tcc-doc.html#Option-summary :wink:

integration-tests/suites/repeated_network_flow.go
ovalenti: Inserted a newline

integration-tests/suites/base.go
JoukoVirtanen: You could try to make it an optional argument. You might not be able to do that, because of the variadic parameter. You might do something like

func (s *IntegrationTestSuiteBase) execContainerShellScriptWithShell(containerName string, script string, shell string, args ...string) (string, error) {
	cmd := []string{common.RuntimeCommand, "exec", "-i", containerName, shell, "-s"}
	cmd = append(cmd, args...)
	return s.executor.ExecWithStdin(script, cmd...)
}

func (s *IntegrationTestSuiteBase) execContainerShellScript(containerName string, script string, args ...string) (string, error) {
	return s.execContainerShellScriptWithShell(containerName, script, "/bin/sh", args...)
}

ovalenti: This suggestion has been implemented. My concern is that all the parameter are strings, including the variadic part; this might be quite error prone if the caller misses one of the initial parameters, since the compiler will be just happy with it :-/

Stringy (approved):
LGTM! It might be worth looking at putting the scripts in separate files for the benefit of linting and clarity, but I think that can be done in a follow up

PR #1049, opened 2023-03-10
Refactors COLLECTOR_SD_LOGGING to use TRACE log level

Stringy:
## Description

Adds configuration to the integration tests to enable it on demand, and removes other references around that point. 

## Checklist

**Automated testing**
  ~- [ ] Added unit tests~

## Testing Performed

Tested locally, with and without the variable.

### DIFF
--- a/.github/workflows/integration-tests-vm-type.yml
+++ b/.github/workflows/integration-tests-vm-type.yml
@@ -94,10 +94,14 @@ jobs:
           GCP_SSH_KEY: ${{ secrets.GCP_SSH_KEY }}
           GCP_SSH_KEY_PUB: ${{ secrets.GCP_SSH_KEY_PUB }}
 
-      - name: Normalize job ID
+      - name: Set environment
         run: |
           NORM_JOB_ID="${JOB_ID/-/}"
           echo "JOB_ID=${NORM_JOB_ID}" | tee -a "$GITHUB_ENV"
+
+          if [[ '${{ contains(github.event.pull_request.labels.*.name, 'integration-tests-trace-logging') }}' == 'true' ]]; then
+            echo "COLLECTOR_LOG_LEVEL=trace" | tee -a "$GITHUB_ENV"
+          fi
 
       - name: Create VMs
         if: ${{ ! inputs.run-benchmarks }}
--- a/integration-tests/suites/common/collector_manager.go
+++ b/integration-tests/suites/common/collector_manager.go
@@ -44,6 +44,7 @@ func NewCollectorManager(e Executor, name string) *CollectorManager {
 		"COLLECTION_METHOD":       collectionMethod,
 		"COLLECTOR_PRE_ARGUMENTS": collectorPreArguments,
 		"ENABLE_CORE_DUMP":        "false",
+		"COLLECTOR_EXTRA_LOGGING": ReadEnvVarWithDefault("COLLECTOR_EXTRA_LOGGING", "false"),
 	}
 	if !offlineMode {
 		env["MODULE_DOWNLOAD_BASE_URL"] = "https://collector-modules.stackrox.io/612dd2ee06b660e728292de9393e18c81a88f347ec52a39207c5166b5302b656"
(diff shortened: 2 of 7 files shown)


Molter73 (commented):
Maybe instead of `extra` we could use something like `inpector`?

Also, we don't need to do it in this PR, but a label that could be added on PRs to get this flag set would be a good to have IMO.

Stringy:
> Maybe instead of `extra` we could use something like `inpector`?

The main reason I chose `extra` was to make it agnostic of component, the thinking being that if we wanted to add some more conditional logging elsewhere or in a different component, we wouldn't need to manage multiple different config options to do so. 

> Also, we don't need to do it in this PR, but a label that could be added on PRs to get this flag set would be a good to have IMO.

Good idea. Should be simple enough to add. I'll do it in this PR.

Molter73:
> The main reason I chose extra was to make it agnostic of component, the thinking being that if we wanted to add some more conditional logging elsewhere or in a different component, we wouldn't need to manage multiple different config options to do so.

Yeah, I get this point, I just think `extra` sounds a little too agnostic or even a little vague. I'd prefer `inpector` since it doesn't mention a specific provider, yet it implies it is enabling the logs for the inspector.

We could also consider adding another level of log, something like `DebugVerbose` could just set the inspector to also log out messages.

Stringy:
Perhaps then we can attach this to the `TRACE` log level and drop the environment variable altogether?

Molter73:
Yeah, I think tying it to the `TRACE` log level could work, I don't see it being used in our codebase at all, so it should be ok.

.github/workflows/integration-tests-vm-type.yml
```diff
@@ -94,11 +94,15 @@ jobs:
...
           GCP_SSH_KEY_PUB: ${{ secrets.GCP_SSH_KEY_PUB }}
 
-      - name: Normalize job ID
+      - name: Set environment
         run: |
           NORM_JOB_ID="${JOB_ID/-/}"
           echo "JOB_ID=${NORM_JOB_ID}" | tee -a "$GITHUB_ENV"
 
+          if [[ '${{ contains(github.event.pull_request.labels.*.name, 'integration-tests-trace-logging') }}' == 'true' ]]; then
+            echo "COLLECTOR_LOG_LEVEL=trace" | tee -a "$GITHUB_ENV"
```
Molter73, line 103: [ultra-nit] Don't know if it's necessary to `tee` the environment variable, the following steps will print it out if you expand them

integration-tests/suites/common/collector_manager.go
```diff
@@ -36,15 +36,15 @@ func NewCollectorManager(e Executor, name string) *CollectorManager {
 		collectionMethod = "kernel_module"
 	}
 
+	logLevel := ReadEnvVarWithDefault("COLLECTOR_LOG_LEVEL", "debug")
 	offlineMode := ReadBoolEnvVar("COLLECTOR_OFFLINE_MODE")
 
 	env := map[string]string{
 		"GRPC_SERVER":             "localhost:9999",
-		"COLLECTOR_CONFIG":        `{"logLevel":"debug","turnOffScrape":true,"scrapeInterval":2}`,
+		"COLLECTOR_CONFIG":        fmt.Sprintf(`{"logLevel":"%s","turnOffScrape":true,"scrapeInterval":2}`, logLevel),
```
Molter73, line 44: Ah yes! Constructing a json string with a print statement! This is what peak performance looks like /s

When can we get rid of the json string being passed as an environment variable that later gets turned to an argument to the collector binary?

Molter73 (approved):
LGTM!

integration-tests/suites/common/collector_manager.go
Stringy: I did consider making a struct and populating it with defaults and then json.Marshal into a string to at least make it a little less hard-coded, but with the upcoming changes to the integration tests, I didn't think it was yet worth it.

PR #1053, opened 2023-03-15
Add CO.RE eBPF candidate

Molter73:
## Description

These changes should make it relatively easy to wire in the CO.RE eBPF probe for loading. They also put in place a fallback mechanism, allowing to try regular eBPF probes in case the CO.RE probe fails.

## Checklist

## Testing Performed

- [x] Run locally setting `COLLECTION_METHOD` to `CORE_BPF` and it fallsback to regular eBPF.

### DIFF
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -230,7 +230,7 @@ int main(int argc, char** argv) {
 
   if (!collector.InitKernel(args->GRPCServer(), candidates)) {
     startup_diagnostics.Log();
-    CLOG(FATAL) << "Failed to initialize collector service.";
+    CLOG(FATAL) << "Failed to initialize collector kernel components.";
   }
 
   // output the GPL notice only once the kernel object has been found or downloaded
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -206,18 +206,6 @@ int main(int argc, char** argv) {
 
   CLOG(INFO) << "Module version: " << GetModuleVersion();
 
-  std::vector<DriverCandidate> candidates = GetKernelCandidates(config.GetCollectionMethod());
-  if (candidates.empty()) {
-    startup_diagnostics.Log();
-    CLOG(FATAL) << "No kernel candidates available";
-  }
-
-  const char* type = config.UseEbpf() ? "eBPF probe" : "kernel module";
-  CLOG(INFO) << "Attempting to find " << type << " - Candidate versions: ";
-  for (const auto& candidate : candidates) {
-    CLOG(INFO) << candidate.GetName();
-  }
-
   // Register signal handlers
   signal(SIGABRT, AbortHandler);
   signal(SIGSEGV, AbortHandler);
@@ -228,7 +216,7 @@ int main(int argc, char** argv) {
 
   CollectorService collector(config, &g_control, &g_signum);
 
-  if (!collector.InitKernel(args->GRPCServer(), candidates)) {
+  if (!SetupKernelDriver(collector, args->GRPCServer(), config)) {
     startup_diagnostics.Log();
     CLOG(FATAL) << "Failed to initialize collector kernel components.";
   }
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -206,7 +206,7 @@ int main(int argc, char** argv) {
 
   CLOG(INFO) << "Module version: " << GetModuleVersion();
 
-  std::vector<DriverCandidate> candidates = GetKernelCandidates(config.UseEbpf());
+  std::vector<DriverCandidate> candidates = GetKernelCandidates(config.CollectionMethod());
   if (candidates.empty()) {
     startup_diagnostics.Log();
     CLOG(FATAL) << "No kernel candidates available";
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -206,7 +206,7 @@ int main(int argc, char** argv) {
 
   CLOG(INFO) << "Module version: " << GetModuleVersion();
 
-  std::vector<DriverCandidate> candidates = GetKernelCandidates(config.CollectionMethod());
+  std::vector<DriverCandidate> candidates = GetKernelCandidates(config.GetCollectionMethod());
   if (candidates.empty()) {
     startup_diagnostics.Log();
     CLOG(FATAL) << "No kernel candidates available";
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -218,8 +218,6 @@ int main(int argc, char** argv) {
     CLOG(INFO) << candidate.GetName();
   }
 
-  // startup_diagnostics.KernelDriverDownloaded();
-
   // Register signal handlers
   signal(SIGABRT, AbortHandler);
   signal(SIGSEGV, AbortHandler);
@@ -238,7 +236,6 @@ int main(int argc, char** argv) {
   // output the GPL notice only once the kernel object has been found or downloaded
   gplNotice();
 
-  startup_diagnostics.KernelDriverLoaded();
   startup_diagnostics.Log();
 
   collector.RunForever();
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -72,8 +72,6 @@ using namespace collector;
 static std::atomic<ControlValue> g_control(ControlValue::RUN);
 static std::atomic<int> g_signum(0);
 
-static StartupDiagnostics g_startup_diagnostics;
-
 static void
 ShutdownHandler(int signum) {
   // Only set the control variable; the collector service will take care of the rest.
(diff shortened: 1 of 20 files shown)


collector/lib/CollectionMethod.h
```diff
@@ -0,0 +1,35 @@
...
+* version of the file(s), but you are not obligated to do so.  If you
+* do not wish to do so, delete this exception statement from your
+* version.
+*/
+
+#ifndef COLLECTION_METHOD_H
+#define COLLECTION_METHOD_H
+
+namespace collector {
+enum collectionMethod {
```
Stringy, line 28: [nit] should probably be TitleCase

collector/lib/CollectorConfig.cpp
```diff
@@ -151,11 +149,18 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
...
     if (args->CollectionMethod().length() > 0) {
-      collection_method_ = args->CollectionMethod();
-      CLOG(INFO) << "User configured collection-method=" << collection_method_;
+      const auto& cm = args->CollectionMethod();
+
+      if (cm == "ebpf") {
+        collection_method_ = EBPF;
+      } else if (cm == "core_bpf") {
+        collection_method_ = CORE_BPF;
+      }
```
Stringy, line 158: `else if (cm == "kernel-module")`?

collector/lib/CollectorConfig.cpp
```diff
@@ -229,10 +234,8 @@ bool CollectorConfig::UseChiselCache() const {
 }
 
 bool CollectorConfig::UseEbpf() const {
```
Stringy, line 236: Given the expansion of the meaning of this method it might be better renamed to `UseBPF` WDYT?

collector/lib/CollectorConfig.h
```diff
@@ -41,10 +42,9 @@ class CollectorArgs;
 class CollectorConfig {
  public:
   static constexpr bool kUseChiselCache = true;
-  static constexpr bool kSnapLen = 0;
   static constexpr bool kTurnOffScrape = false;
   static constexpr int kScrapeInterval = 30;
-  static constexpr char kCollectionMethod[] = "kernel-module";
+  static constexpr collectionMethod kCollectionMethod = KERNEL_MODULE;
```
Stringy, line 47: Since we're defaulting to EBPF these days, we should probably set this to follow that default

collector/collector.cpp
```diff
@@ -251,13 +228,15 @@ int main(int argc, char** argv) {
 
   CollectorService collector(config, &g_control, &g_signum);
 
-  if (!collector.InitKernel()) {
-    g_startup_diagnostics.Log();
+  if (!collector.InitKernel(args->GRPCServer(), candidates)) {
```
Stringy, line 231: this is maaaaybe a nit, but I find it very odd at this level that we're passing the address of the GRPC server to a function that is meant to initialize kernel components - from an encapsulation perspective it feels a bit off. I'm not opposed to the collector service doing the download, but can it get this information in a different way that doesn't require the caller to know about it? Specifically, can we store the GRPC address in the `CollectorConfig` which is accessible from the `CollectorService`?

collector/lib/CollectorService.cpp
```diff
@@ -148,8 +150,31 @@ void CollectorService::RunForever() {
...
+      continue;
+    }
+
+    startup_diagnostics.DriverAvailable(candidate.GetName());
+
+    if (sysdig_.InitKernel(config_, candidate)) {
+      startup_diagnostics.DriverSuccess(candidate.GetName());
+      return true;
+    } else if (candidate.GetCollectionMethod() == KERNEL_MODULE) {
+      // Kernel module drops capabilities, so subsequent attempts could fail
```
Stringy, line 169: Could we ease this a little by enforcing that kernel module is always the last candidate? Probs wouldn't affect it if we were trying multiple kernel modules, and also we'll be removing kernel modules soon anyway... this is probably fine.

Stringy (commented):
Couple of questions/comments but mostly looks excellent, and the various additional improvements and tidy up are greatly appreciated 🎉

collector/lib/CollectionMethod.h
Molter73: I kinda have a collision problem, since some classes have a `CollectionMethod` method that acts as a getter, there are 2 possible solutions:
- We prefix the type were needed as `enum CollectionMethod`
- We rename the getters to be `GetCollectionMethod`

I think the second option is nicer and should make it easier when grepping for it.

collector/lib/CollectorConfig.cpp
Molter73: Well BPF and eBPF are kind of interchangeable at this point, I know we used `core_bpf` as the collection method but I've also used `CO.RE eBPF` in a bunch of logging messages, I think it's fine to keep this as `UseEbpf`. I think it's more important to understand that this function returns true if ANY kind of eBPF/BPF probe is used, which I think it does.

collector/lib/CollectorConfig.h
Molter73: Yeah, we probably should, but this PR is already quite tainted by changes that are not specific to adding the `CO.RE eBPF` candidate, mind if I do it in a follow up?

collector/collector.cpp
Molter73: Yeah I don't like this approach either. My initial thought was to have an independent function do the download but I turned out to be a bit awkward too. I'll take another look, maybe we just want some sort of `MakeCollectorService` function that takes whatever we need (I also want to cut down on the amount of stuff that is needed to be copied).

collector/lib/CollectionMethod.h
Stringy: I agree, I think the second option is best in this case

collector/collector.cpp
Molter73: Ended up adding a `SetupKernelDriver` that takes in a `CollectorService` object alongside the grpc server address and the configuration, and in turn handles everything from finding/downloading the drivers to calling `collector.InitKernel`, hopefully this is a bit better than the previous approach.

Stringy: Yeah that approach looks good!

Stringy (approved):
LGTM!

erthalion (commented):
Look great, thanks! Few commentaries/questions.

collector/lib/Diagnostics.h
```diff
@@ -15,29 +15,54 @@ class IDiagnostics {
...
+    static StartupDiagnostics instance;
+
+    return instance;
   }
 
   void ConnectedToSensor() { connectedToSensor_ = true; }
-  void KernelDriverDownloaded() { kernelDriverDownloaded_ = true; }
-  void KernelDriverLoaded() { kernelDriverLoaded_ = true; }
+
+  void DriverUnavailable(const std::string& candidate) {
```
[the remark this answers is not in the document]
Molter73: Well, unavailable means unavailable, as in, it's nowhere to be found (in the image, online, etc..). Adding the distinction between local or downloadable can be tricky, we would need to add some logic to the `GetKernelObject` function and I don't see it as necessarily useful.

The search path is already print in previous logs iirc and, excluding the candidate from `COLLECTOR_DRIVER`, it will always be `/kernel-modules/<candidate name>`.

collector/lib/CollectorConfig.cpp
```diff
@@ -150,12 +148,19 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
     }
 
     // Collection Method
-    if (args->CollectionMethod().length() > 0) {
-      collection_method_ = args->CollectionMethod();
-      CLOG(INFO) << "User configured collection-method=" << collection_method_;
+    if (args->GetCollectionMethod().length() > 0) {
+      const auto& cm = args->GetCollectionMethod();
+
+      if (cm == "ebpf") {
```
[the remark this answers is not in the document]
Molter73: It's still possible, but doing `export COLLECTION_METHOD=incorrect` will also default to `kernel_module`, this behavior is unchanged by this PR, master does the same.

collector/lib/GetKernelObject.cpp
```diff
@@ -141,9 +141,15 @@ bool DownloadKernelObject(const std::string& hostname, const Json::Value& tls_co
 }
 
 bool GetKernelObject(const std::string& hostname, const Json::Value& tls_config, const DriverCandidate& candidate, bool verbose) {
+  if (candidate.GetCollectionMethod() == CORE_BPF) {
+    // for now CO.RE bpf probes are embedded in the collector binary, nothing
```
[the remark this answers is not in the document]
Molter73: I don't get how that would work if the probe is embedded in the collector binary, do we just look for the collector binary? It's the file that is running this code, it MUST exist.

Note here that I said binary, not image.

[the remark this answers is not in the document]
erthalion: Oh, yeah, good point -- in my mind I was already a few steps ahead, thinking about a separated probe.

collector/lib/Diagnostics.h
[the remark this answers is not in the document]
erthalion: What I mean here is that we operate with an object "candidate", but diagnostics get only the name. Wouldn't it be beneficial to provide the full candidate there? There is no need to do any distinction, just dump all the info we got into the output.

collector/lib/CollectorService.cpp
erthalion: I disagree. The whole logic may change multiple time, and it's just nice to have some invariants verified.

collector/lib/Diagnostics.h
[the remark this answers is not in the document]
Molter73: My understanding is that the message output by this diagnostics is a meant to be a short, meaningful message that users/support should be able to quickly interpret, if we start dumping things in it, it will lose this.

@Stringy, you put together the feature, what do you think?

erthalion (approved):

PR #1059, opened 2023-03-20
Stacktrace improvements

erthalion:
## Description

Collector already uses a signal handler to report the stacktrace in case of failures, but doesn't report all the symbols because:

* Collector is build without `-rdynamic`, missing some set of symbols in the dynamic symbols table [[1]]. Adding it means the binary will become bigger on the amount of added symbols (5.9 Mb vs 4.7 Mb without it). Worth noting that those are not debugging symbols.

* Collector is build without `-fno-omit-frame-pointer`, potentially missing some part of the stacktrace [[2]]. Adding it means a slight performance overhead, but it's well known that it's negligible. E.g. Fedora project recently switched to use it by default [[3]].

Adding those two options will give more readable stacktrace, that contains reasonable symbolic names for functions. But the C++ parts are going to be mangled and less readable. To address this, use `Dl_info` and `abi::__cxa_demangle` [[4]] if possible.

Here is how the new stacktrace would look like:

```
    /lib64/libc.so.6 (null) 0x7f8b3df4bb80 + 0
    /usr/local/lib/libsinsp-wrapper.so check_api_compatibility 0x7f8b40c36373 + 12
    /usr/local/lib/libsinsp-wrapper.so (null) 0x7f8b40c344e9 + 0
    /usr/local/lib/libsinsp-wrapper.so scap_open_live_int 0x7f8b40c1be01 + 413
    /usr/local/lib/libsinsp-wrapper.so scap_open 0x7f8b40c1c969 + 279
    /usr/local/lib/libsinsp-wrapper.so sinsp::open_common(scap_open_args*) 0x7f8b40931a34 + 236
    /usr/local/lib/libsinsp-wrapper.so sinsp::open_bpf(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, unsigned long, std::unordered_set<unsigned int, std::hash<unsigned int>, std::equal_to<unsigned int>, std::allocator
    collector collector::SysdigService::Start() 0x75467b + 1155
    collector collector::CollectorService::RunForever() 0x6ccc86 + 2396
    collector main 0x6bb0d8 + 1593
    /lib64/libc.so.6 __libc_start_main 0x7f8b3df37d85 + 229
    collector _start 0x6b916e + 46
```

[1]: https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html
[2]: https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/Optimize-Options.html
[3]: https://fedoraproject.org/wiki/Changes/fno-omit-frame-pointer
[4]: https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-html-USERS-4.3/a01696.html

## Checklist

**Automated testing**
  - [x] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Local testing on the various types of traces, forced to truncate the stack, the message buffer, or failed to translate the address.

### DIFF
--- a/collector/CMakeLists.txt
+++ b/collector/CMakeLists.txt
@@ -3,7 +3,7 @@ project(collector-bin)
 find_package(Threads)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR})
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -Wall --std=c++17 -pthread -Wno-deprecated-declarations")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -Wall --std=c++17 -pthread -Wno-deprecated-declarations -fno-omit-frame-pointer -rdynamic")
 
 set(CMAKE_CXX_FLAGS_DEBUG "-g -ggdb")
 set(CMAKE_CXX_FLAGS_RELEASE "-O3 -fno-strict-aliasing -DNDEBUG")
@@ -13,7 +13,7 @@ if(NOT DEFINED ENV{DISABLE_PROFILING})
 endif()
 
 if(ADDRESS_SANITIZER)
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-omit-frame-pointer -fsanitize=address")
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
 endif()
 
 if(COLLECTOR_APPEND_CID)
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -117,7 +117,7 @@ static void AbortHandler(int signum) {
   size_t max_frames = sizeof(buffer) / sizeof(buffer[0]);
   int message_buffer_size = sizeof(message_buffer);
 
-  for (size_t i = 1; i < n_frames; i++) {
+  for (size_t i = 0; i < n_frames; i++) {
     Dl_info info;
 
     if (dladdr(buffer[i], &info)) {
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -37,6 +37,8 @@ You should have received a copy of the GNU General Public License along with thi
 extern "C" {
 #include <assert.h>
 #include <cap-ng.h>
+#include <cxxabi.h>
+#include <dlfcn.h>
 #include <execinfo.h>
 #include <fcntl.h>
 #include <stdio.h>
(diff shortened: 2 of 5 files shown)


collector/collector.cpp
```diff
@@ -81,19 +83,85 @@ ShutdownHandler(int signum) {
...
+  char message_buffer[256];  // Actual text buffer for a single stack trace
+                             // address. Some templated signatures could be
+                             // particularly long.
+  int num_bytes;
+
+  size_t n_frames = backtrace(buffer, 32);
+  size_t max_frames = sizeof(buffer) / sizeof(buffer[0]);
+  int message_buffer_size = sizeof(message_buffer);
+
+  for (size_t i = 1; i < n_frames; i++) {
```
Molter73, line 120: Why do we start from 1 and not 0?

collector/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ project(collector-bin)
 find_package(Threads)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR})
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -Wall --std=c++17 -pthread -Wno-deprecated-declarations")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -Wall --std=c++17 -pthread -Wno-deprecated-declarations -fno-omit-frame-pointer -rdynamic")
```
Molter73, line 6: What are the implications of always compiling with these two flags? Is the final binary bigger? Are symbols preserved in it? Does it affect performance in any meaningful way? I'll do some research on my own, but would love to hear your thoughts.

collector/collector.cpp
```diff
@@ -81,19 +83,85 @@ ShutdownHandler(int signum) {
...
-  void* buffer[32];
-  size_t size = backtrace(buffer, 32);
-  backtrace_symbols_fd(buffer, size, STDERR_FILENO);
+  // Write a stacktrace to stderr. Since the abort handler could be called
+  // after SIGSEGV or some other emergency situations:
+  // * we have to be modest and use little memory (in case if the original
+  //   reason was OOM)
+  // * as the original implementation noted, use reentrant functions
+  //
+  // The latter one was refering to snprintf, although it's not cleat if it is
```
Molter73, line 103: [nit] Typo
```suggestion
  // The latter one was refering to snprintf, although it's not clear if it is
```

Molter73 (commented):
Overall looks good, maybe we could move `WriteBuffer` and the core of the logic for backtrace dumping to a dedicated file to keep a cleaner main file. I'll have to look into the new flags used a little more to better understand them before approving.

collector/collector.cpp
erthalion: Somehow I was under the impression, that the first address is not what we're looking for, but let me check.

collector/CMakeLists.txt
Molter73: Yes, me reading the PR description 🤦🏻‍♂️

collector/collector.cpp
erthalion: Now I remember, the first function would be the signal handler, which is not necessarily what we're looking for. But at the same time doesn't hurt to report it as well.

Molter73 (approved):
LGTM!

collector/lib/AbortHandler.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+static void WriteBuffer(char* buffer, int num_bytes, int buffer_size) {
+  if (num_bytes > buffer_size) {
+    // the output was truncated, to keep it readable add a new line
+    buffer[buffer_size - 1] = '\n';
+    write(STDERR_FILENO, buffer, buffer_size);
+  } else {
+    write(STDERR_FILENO, buffer, num_bytes);
+  }
+}
```
Molter73, line 34: When I first joined StackRox I was told to use unnamed namespaces instead of `static`, after doing some research it looks like there's no difference between both approaches, but just because we've used this in other files, I have to mention it.
```suggestion
namespace {
void WriteBuffer(char* buffer, int num_bytes, int buffer_size) {
  if (num_bytes > buffer_size) {
    // the output was truncated, to keep it readable add a new line
    buffer[buffer_size - 1] = '\n';
    write(STDERR_FILENO, buffer, buffer_size);
  } else {
    write(STDERR_FILENO, buffer, num_bytes);
  }
}
}
```

Molter73 (approved):
One minor comment on using unnamed namespace over `static`, but the PR is honestly ok to merge as is.

PR #1060, opened 2023-03-20
Remove Dockerfile.template and use build arguments instead

Molter73:
## Description

Replace `Dockerfile.template` with a regular `Dockerfile`. Instead of using `envsubst` to replace content in it, we pass what we need as build arguments. The end result should be the same with less intermediate operations.

## Checklist

## Testing Performed

CI should be enough.


Stringy (approved):
LGTM! this seems a lot clearer and simpler

PR #1062, opened 2023-03-23
Block kernels with missing cc_mkenc (5.14.0 RHEL9)

ovalenti:
## Description

Some newly crawled RHEL9 kernels (5.14.0) contain backported material that triggers an issue with cc_mkenc not being exported. 

## Checklist


Molter73 (commented):
I've added the `build-legacy-probes` label, this should make it so the next commit builds drivers for all version, including `2.0.1` which is the one failing.

Molter73 (approved):
LGTM!

PR #1071, opened 2023-03-27
Automate Garden Linux VM version bump

Molter73:
## Description

This PR adds some logic to our CI in order to automate creation of a PR that bumps the Garden Linux VM used for testing when a new release for that distribution is done.

The `bumper.py` script will make a request to the [GitHub API asking for the latest release of Garden Linux](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release), then compare the version in its assets to the one configured in the repository and update the repo if the major or minor versions are bigger in the latest release. The Garden Linux version was moved to a separate file in order to prevent automation from changing anything other than that file.

The new workflow is set to run daily, this shouldn't be a big deal since Garden Linux releases at most every couple of weeks, the action used for creating the PR will not do anything if the latest version still matches the one in the repo (see https://github.com/marketplace/actions/create-pull-request for more details on how the action works).

## Checklist

## Testing Performed

- [x] A new PR should be created from this one, since the Garden Linux VM has not been updated on master yet.
- [x] Run with `all-integration-tests` to ensure the Garden Linux test still works correctly.


.github/workflows/gardenlinux-bumper.yml
```diff
@@ -0,0 +1,63 @@
...
+        with:
+          token: '${{ secrets.ROBOT_ROX_GITHUB_TOKEN }}'
+          commit-message: Update Garden Linux VM
+          committer: RoxBot <roxbot@stackrox.com>
+          author: RoxBot <roxbot@stackrox.com>
+          branch: roxbot/gardenlinux-bump
+          signoff: false
+          delete-branch: true
+          title: Update Garden Linux VM
+          base: mauro/gardenlinux-vm-bumper
```
Molter73, line 40: - [ ] Remove this line before merge

utilities/gardenlinux-bumper/bumper.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def image_is_outdated(latest: (int, int, str), current: (int, int, str)):
+    latest_major = latest[0]
+    latest_minor = latest[1]
+
+    current_major = current[0]
+    current_minor = current[1]
+
+    return latest_major > current_major or (latest_major == current_major and latest_minor > current_minor)
```
Stringy, line 63: [nit] this can be slightly simplified by using tuple comparison

```
return (latest[0], latest[1]) > (current[0], current[1])
```

ansible/group_vars/all.yml
```diff
@@ -85,4 +85,4 @@ virtual_machines:
   garden-linux:
     project: sap-se-gcp-gardenlinux
     images:
-      - gardenlinux-gcp-gardener-prod-amd64-934-6-76ea662
+      - "{{ lookup('file', 'group_vars/gardenlinux-image.txt') }}"
```
Stringy, line 88: Do we ever wish to support more than one image (like we do for RHCOS)? It might be worth making `gardenlinux-image.txt` a json file (with just a list of versions in it) so we can add more images if necessary, but probably not worth doing in this PR

Stringy (approved):
Couple of very minor questions/comments but otherwise looks great!

ansible/group_vars/all.yml
Molter73: I also considered using a yaml file and using the [include_vars](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/include_vars_module.html) module, but had some trouble testing it and decided to go the path of least resistance (just use a file and call it a day).

Both these approaches would be nice to haves once we automate the RHCOS update.

utilities/gardenlinux-bumper/bumper.py
Molter73: I can probably use a sub-tuple instead, not sure what you think about readability though
```python
return latest[:2] > current[:2]
```

ansible/group_vars/all.yml
Molter73: As for the question on whether we want to support more than the latest version, no, I don't think we want to, latest one should be enough.

Stringy:
Just looked over the recent changes, and it still LGTM

PR #1072, opened 2023-03-27
ebpf probe changes for s390x enablement

kcrane:
## Description

This PR enables the BPF probe on the s390x architecture and also implements the socketcall syscall to catch network calls. Also this PR includes generic changes to fit to the [collector custom probe](https://github.com/stackrox/collector/blob/master/kernel-modules/probe/collector_probe.c).

Corresponding falcosecurity-libs PR https://github.com/stackrox/falcosecurity-libs/pull/44


Stringy (commented):
Just one refactoring suggestion in the probe, but otherwise it's looking good! Could you also bump the module version release candidate in `kernel-modules/MODULE_VERSION` to `2.4.0-rc2`

Molter73:
Run integration tests for Ubuntu and RHEL VMs in #1073 and they passed correctly. I'm now re-running with all other distros, but for the moment looks like the changes work on x86.

kcrane (commented):
We currently use the following structure:
```
struct sys_enter_setgid_args {
  __u64 pad;
  int __syscall_nr;
  gid_t gid;
};
```

To match with syscall format:
```
# cat /sys/kernel/tracing/events/syscalls/sys_enter_setgid/format
name: sys_enter_setgid
ID: 185
format:
	field:unsigned short common_type;	offset:0;	size:2;	signed:0;
	field:unsigned char common_flags;	offset:2;	size:1;	signed:0;
	field:unsigned char common_preempt_count;	offset:3;	size:1;	signed:0;
	field:int common_pid;	offset:4;	size:4;	signed:1;

	field:int __syscall_nr;	offset:8;	size:4;	signed:1;
	field:gid_t gid;	offset:16;	size:8;	signed:0;

print fmt: "gid: 0x%08lx", ((unsigned long)(REC->gid))
```

Notice __syscall_nr starts at offset 8 is 4 bytes (int) and gid starts at offset 16.  The compiler will pack adjacent fields if they have correct alignment for that type.  In this case __syscall_nr and gid will be packed together.  Whereas:
```
struct sys_enter_execve_args {
  __u64 pad;
  int __syscall_nr;
  const char* filename;
  const char* argv;
  const char* envp;
};
```
will not have __syscall_nr and filename packed together since filename needs to be on 8-byte boundary (pointer).  Explained here: http://www.catb.org/esr/structure-packing/

The following are syscalls I believe will be an issue due to fields being packed together:
```
struct sys_enter_setuid_args {
  __u64 pad;
  int __syscall_nr;
  uid_t uid;
};

struct sys_enter_setgid_args {
  __u64 pad;
  int __syscall_nr;
  gid_t gid;
};

struct sys_enter_setresuid_args {
  __u64 pad;
  int __syscall_nr;
  uid_t ruid;
  uid_t euid;
  uid_t suid;
};

struct sys_enter_setresgid_args {
  __u64 pad;
  int __syscall_nr;
  gid_t rgid;
  gid_t egid;
  gid_t sgid;
};
```
by changing gid_t and uid_t to “unsigned long”, it will force it to not be packed together.  All integration tests continue to pass with this commit.

kernel-modules/probe/collector_probe.c
```diff
@@ -198,6 +218,127 @@ PROBE_SIGNATURE("sched/", sched_process_exit, sched_process_exit_args) {
   return 0;
 }
 
+////////////////////////////////////////////////////////////////////////////////////
```
robbycochran, line 221: Can you add a reference to where these structs were copied from?

kernel-modules/probe/collector_probe.c
```diff
@@ -198,6 +218,127 @@ PROBE_SIGNATURE("sched/", sched_process_exit, sched_process_exit_args) {
...
+  int __syscall_nr;
+  long fd;
+  struct sockaddr* uservaddr;
+  long addrlen;
+};
+
+struct sys_enter_clone_args {
+  __u64 pad;
+  int __syscall_nr;
+#ifdef __s390x__
```
robbycochran, line 260: How do we keep track of these argument orderings?

kernel-modules/probe/collector_probe.c
```diff
@@ -227,17 +368,172 @@ static __always_inline int enter_probe(long id, struct sys_enter_args* ctx) {
...
+   * data.  Note that the syscall tracepoints do not provide any argument data
+   * in the syscall exit format.  Of course, trying to access argument data in
+   * filler results in BPF verifier errors.
+   *
+   * To provide arguments to the syscall enter and exit fillers, extract the
+   * arguments from syscall specific formats/structures to an sys_enter_args
+   * structure and use a BPF map to make them available to the enter/exit
+   * fillers.
+   */
+  switch (id) /* maybe use sc_evt here..... */
```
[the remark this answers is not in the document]
robbycochran: I agree, let's pull this switch out into an inlined function.

robbycochran (changes requested):

robbycochran:
Regarding struct packing, did we consider using a compiler directive instead of switching from `uid_t` to `unsigned int` ?

kernel-modules/probe/collector_probe.c
kcrane: > Regarding struct packing, did we consider using a compiler directive instead of switching from `uid_t` to `unsigned int` ?

I'm aware of a compiler directive to always pack (regardless of type alignment).  But not sure how that would help in using uid_t as it would guarantee that it would pack.  Is there another directive you have in mind?

Originally, I put padding in to explicitly "unpack" the structure (and keep the uid_t and gid_t types).  But found that we relied on implicit packing elsewhere in code and with endian differences between platforms the code looked a little verbose.  I put a comment at start of structure to remind people that implicit padding is assumed.   Also, in the syscall format, the gid and uid fields actually have a size of 8 bytes whereas uid_t and gid_t are 4 byte types.  Even though we won't have an 8 byte gid, the unsigned long actually makes sense here.

hbrueckner (approved):
Hi @kcrane , LGTM! Many thanks!

kernel-modules/probe/collector_probe.c
Stringy: Would the use of something like `__attribute__((aligned(8))` to force alignment/padding solve this problem?

kcrane: I think it would solve half the problem.  The alignment would be good, but since syscall format is 8 bytes and our type is 4 bytes, we would need to wrap everything in a \#ifdef w/ padding to handle big/little endian differences.

kernel-modules/probe/collector_probe.h
```diff
@@ -0,0 +1,283 @@
+#ifndef __COLLECTOR_PROBE_H
+#define __COLLECTOR_PROBE_H
+
+/* The structs below correspond to the syscall tracepoint format found 
+ * in /sys/kernel/tracing/events/syscalls
+ * example: cat /sys/kernel/tracing/events/syscalls/sys_enter_setgid/format
```
Stringy, line 6: To be clear, are these manually written based on the `format` files or is there a tool that can be used to generate the structures based on those files?

kernel-modules/probe/collector_probe.c
```diff
@@ -227,17 +248,23 @@ static __always_inline int enter_probe(long id, struct sys_enter_args* ctx) {
...
+    drop_flags = sc_evt->flags;
   }
 
-  evt_type = sc_evt->enter_event_type;
-  drop_flags = sc_evt->flags;
-
-  // To satisfy some verifier requiremnts in later parts of the falco plumbing/fillers,
-  // it is necessary to copy the context onto the stack.
-  memcpy(stack_ctx.args, _READ(ctx->args), sizeof(unsigned long) * NUM_SYS_ENTER_ARGS);
+  syscall_to_enter_args(id, ctx, &stack_ctx);
```
Stringy, line 267: To reduce uncertainty / complexity on other platforms, we should only perform this copying on s390x. I think a simple ifdef guard here should be enough.

```suggestion
#ifdef __s390x__
  syscall_to_enter_args(id, ctx, &stack_ctx);
#else
  evt_type = sc_evt->enter_event_type;
  drop_flags = sc_evt->flags;
  // To satisfy some verifier requiremnts in later parts of the falco plumbing/fillers,
  // it is necessary to copy the context onto the stack.
  memcpy(stack_ctx.args, _READ(ctx->args), sizeof(unsigned long) * NUM_SYS_ENTER_ARGS);
#endif
```

Stringy (changes requested):

kernel-modules/probe/collector_probe.h
hbrueckner: @Stringy those are manually written. I looked around some tooling for this but did not find anything useful. However, it looks like that [tplist](https://github.com/iovisor/bcc/blob/master/tools/tplist_example.txt) might help in future.

kernel-modules/probe/collector_probe.c
hbrueckner: @Stringy Hmm... my concern here is that you copy the trace point specific format into a `struct sys_enter_args stack_ctx;` under those assumptions. This actually caused my a couple of BPF verifier issues e.g. if the tp format structure is shorter in length than `struct sys_enter_args`. In the worst case, you copy more memory than necessary (not sure why this seems to work on non-s390x architectures). Finally, that was the reason why I introduced those tp specific formats and assign the parameters directly.

kernel-modules/probe/collector_probe.h
Stringy: Can we please update this comment to make it clear that this is the way these have been constructed

kernel-modules/probe/collector_probe.c
Stringy: Yeah I completely see where you're coming from, but for the time being I think it would be best to isolate this new behaviour to s390x and leave the other architectures as they are, since they have been consistently working for some time. We can certainly revisit in a subsequent release, but with our intention to move to CO.RE, this may be sufficient until that point.

kcrane: resolved in https://github.com/kcrane/collector/commit/04a208802a9bec1159e32c31006af01b61de8bd5

kernel-modules/probe/collector_probe.h
kcrane: resolved in https://github.com/kcrane/collector/commit/04a208802a9bec1159e32c31006af01b61de8bd5

kernel-modules/probe/collector_probe.c
```diff
@@ -121,10 +123,12 @@ static __always_inline int exit_probe(long id, struct sys_exit_args* ctx);
 // this if statement relies on short circuiting to simplify the definition
 // of the tracepoints. i.e. RHEL_RELEASE_VERSION will not be defined unless
 // RHEL_RELEASE_CODE is defined.
-#if !defined(RHEL_RELEASE_CODE) || RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(8, 0)
+#ifdef USE_COLLECTOR_CUSTOM_PROBES
```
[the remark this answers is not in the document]
kcrane: resolved with https://github.com/stackrox/collector/pull/1072/commits/0f7d3e511a4aacbe6e45b95ea961a3cdd067f947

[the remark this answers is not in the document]
hbrueckner: @kcrane Thanks!

kernel-modules/probe/collector_probe.c
```diff
@@ -121,10 +123,12 @@ static __always_inline int exit_probe(long id, struct sys_exit_args* ctx);
 // this if statement relies on short circuiting to simplify the definition
 // of the tracepoints. i.e. RHEL_RELEASE_VERSION will not be defined unless
 // RHEL_RELEASE_CODE is defined.
-#if !defined(RHEL_RELEASE_CODE) || RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(8, 0)
+#ifdef !defined(RHEL_RELEASE_CODE) || RHEL_RELEASE_CODE >= RHEL_RELEASE_VERSION(8, 0)
```
Stringy, line 126: this needs to be `#if` rather than `#ifdef`

Stringy (changes requested):

kernel-modules/probe/collector_probe.h
```diff
@@ -0,0 +1,296 @@
+#ifndef __COLLECTOR_PROBE_H
+#define __COLLECTOR_PROBE_H
+
+/* The structs below have been manually written to their corresponding syscall
+ * tracepoint format found in /sys/kernel/tracing/events/syscalls
+ * example: cat /sys/kernel/tracing/events/syscalls/sys_enter_setgid/format
+ *
+ * Note: will rely on implicit padding added by the compiler to ensure fields
```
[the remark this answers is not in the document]
kcrane: fixed in https://github.com/stackrox/collector/pull/1072/commits/fb069d33b2204d47ce293d138d61268915f62a9e

Stringy:
I've tested these changes on x86 here: https://github.com/stackrox/collector/pull/1170 and it's all green.

@kcrane - what's the progress on the GID issue on s390x? can I assist with that at all?

Stringy (approved):
LGTM, based on testing in https://github.com/stackrox/collector/pull/1170

If before the release we wish to disable eBPF on s390x again (depending on the outstanding GID bug) we can revert the changes in `HostInfo.h` but the changes in the probe can remain since they are already isolated.

PR #1080, opened 2023-03-30
Add the ability to run tests with variable collection methods

### LINKED ISSUE #1067: Integration testing
Investigate integration test results for BTF aware probes.

Part of #1014

Molter73:
## Description

More importantly, this allows to set the collection method to include the new core_bpf driver. This collection method and now be added for PRs using the 'run-corebpf-tests' label.

Fixes #1067 

## Checklist

## Testing Performed

- [x] Ensure running CI with `run-corebpf-tests` runs the tests with the `core_bpf` collection method.
- [x] Ensure running CI without `run-corebpf-tests` skips the tests with the `core_bpf` collection method.

### DIFF
--- a/.github/workflows/integration-tests-vm-type.yml
+++ b/.github/workflows/integration-tests-vm-type.yml
@@ -103,6 +103,10 @@ jobs:
             echo "COLLECTOR_LOG_LEVEL=trace" | tee -a "$GITHUB_ENV"
           fi
 
+          if [[ '${{ contains(github.event.pull_request.labels.*.name, 'run-corebpf-tests') }}' == 'true' ]]; then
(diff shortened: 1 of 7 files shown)


Molter73:
Just realized benchmarking is not being run for the CO-RE bpf probe (or at least it is not showing up in the commit), I'll try to add that as part of this PR as well.

Molter73:
Benchmarks seem to be running with the CO-RE bpf probe now, but they are not being shown in the report because it is not present in the baseline file (and it won't be until we run benchmarks on the master branch).

Stringy (approved):
LGTM! I think there's probably some more we can do to consolidate the uses of collection method throughout the playbooks (so there's less duplication) but that can be done in a follow up I think.

PR #1082, opened 2023-03-30
Bumps falco commit to include CMake patches

Stringy:
## Description

Includes stringification fix for cmake (which was broken downstream)


Molter73 (approved):
LGTM!

PR #1088, opened 2023-04-03
Increases driver download retries

Stringy:
## Description

Testing has revealed there is a situation where Sensor is up, but Central isn't, and this can result in Collectors restarting due to missing drivers. Whilst being unable to download drivers is an expected behaviour, increasing the timeouts will create a bigger window for components to start without having to restart collector each time.


ovalenti (approved):
I agree that this will help and shouldn't hurt, thanks :+1:

PR #1097, opened 2023-04-06
Jv enable processes listening on ports by default

JoukoVirtanen:
## Checklist

## Testing Performed

CI is sufficient


ovalenti (approved):
I checked that the ProcfsScraperTest applies the feature flag as "disable": `processesListeningOnPorts: 0` => no PLoP information added
 
That should be fine !

PR #1098, opened 2023-04-07
[WIP] Build multi-arch collector-slim images

anup-kodlekere:
## Description

This PR intends to add support for building multi-arch collector images with the use of docker buildx. It is part of a broader effort in enabling OSCI for IBM Power.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

### DIFF
--- a/.github/workflows/collector-slim.yml
+++ b/.github/workflows/collector-slim.yml
@@ -18,11 +18,18 @@ env:
   COLLECTOR_TAG: ${{ inputs.collector-tag }}
   DEFAULT_BUILDER_TAG: cache
   RHACS_ENG_IMAGE: quay.io/rhacs-eng/collector:${{ inputs.collector-tag }}
+  ARCH: linux/amd64,linux/ppc64le
 
 jobs:
   build-collector-image:
+
     name: Build the collector slim image
     runs-on: ubuntu-latest
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
 
     steps:
       - uses: actions/checkout@v3
@@ -59,9 +66,17 @@ jobs:
             echo "ADDRESS_SANITIZER=true" >> "$GITHUB_ENV"
           fi
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
       - name: Build images
         run: |
           make image
+
+      # - name: Push multiarch image to stackrox-io and rhacs-eng
 
       - name: Retag and push stackrox-io -slim
         uses: ./.github/actions/retag-and-push
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ MOD_VER_FILE=$(CURDIR)/kernel-modules/kobuild-tmp/MODULE_VERSION.txt
 LOCAL_SSH_PORT ?= 2222
 DEV_SSH_SERVER_KEY ?= $(CURDIR)/.collector_dev_ssh_host_ed25519_key
 
+ARCH ?= linux/amd64
+
 export COLLECTOR_VERSION := $(COLLECTOR_TAG)
 export MODULE_VERSION := $(shell cat $(CURDIR)/kernel-modules/MODULE_VERSION)
 
@@ -30,11 +32,11 @@ container-dockerfile-dev:
 .PHONY: builder
 builder:
 ifdef BUILD_BUILDER_IMAGE
-	docker build \
+	docker-buildx build --push --platform ${ARCH} \
 		--build-arg NPROCS=$(NPROCS) \
 		--cache-from quay.io/stackrox-io/collector-builder:cache \
 		--cache-from quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG) \
-		-t quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG) \
+		-t localhost:5000/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG) \
 		-f "$(CURDIR)/builder/Dockerfile" \
 		.
 else
@@ -62,11 +64,11 @@ build-drivers:
 
 image: collector unittest
 	make -C collector txt-files
-	docker build \
+	docker-buildx build --push --platform ${ARCH} \
 		--build-arg COLLECTOR_VERSION="$(COLLECTOR_TAG)" \
 		--build-arg MODULE_VERSION="$(MODULE_VERSION)" \
 		-f collector/container/Dockerfile \
-		-t quay.io/stackrox-io/collector:$(COLLECTOR_TAG) \
+		-t localhost:5000/stackrox-io/collector:$(COLLECTOR_TAG) \
 		$(COLLECTOR_BUILD_CONTEXT)
 
 image-dev: collector unittest container-dockerfile-dev
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ container-dockerfile-dev:
 .PHONY: builder
 builder:
 ifdef BUILD_BUILDER_IMAGE
-	docker-buildx build --push --platform ${ARCH} \
+	docker buildx build --push --platform ${ARCH} \
 		--build-arg NPROCS=$(NPROCS) \
 		--cache-from quay.io/stackrox-io/collector-builder:cache \
 		--cache-from quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG) \
@@ -64,7 +64,7 @@ build-drivers:
 
 image: collector unittest
 	make -C collector txt-files
-	docker-buildx build --push --platform ${ARCH} \
+	docker buildx build --push --platform ${ARCH} \
 		--build-arg COLLECTOR_VERSION="$(COLLECTOR_TAG)" \
 		--build-arg MODULE_VERSION="$(MODULE_VERSION)" \
 		-f collector/container/Dockerfile \


Makefile
```diff
@@ -69,6 +89,16 @@ image: collector unittest
...
 		$(COLLECTOR_BUILD_CONTEXT)
 
+ma-image: ma-collector unittest
+	make -C collector txt-files
+	docker buildx build --push --platform ${ARCH} \
+		--build-arg COLLECTOR_VERSION="$(COLLECTOR_TAG)" \
+		--build-arg MODULE_VERSION="$(MODULE_VERSION)" \
+		-f collector/container/Dockerfile \
+		-t quay.io/stackrox-io/collector:$(COLLECTOR_TAG)-slim \
+		-t quay.io/stackrox-io/collector:$(COLLECTOR_TAG)-base \
```
Molter73, line 99: CI is failing because we are missing a tag without the suffixes
```suggestion
		-t quay.io/stackrox-io/collector:$(COLLECTOR_TAG)-slim \
		-t quay.io/stackrox-io/collector:$(COLLECTOR_TAG)-base \
		-t quay.io/stackrox-io/collector:$(COLLECTOR_TAG) \
```

docs/how-to-start.md
```diff
@@ -37,6 +37,9 @@ Or pipe output of the make target:
 ```bash
 $ make image 2>&1 | sed -e "s|/src|${PWD}/collector/|g"
 ```
+*NOTE*: To build a multi-arch image, use the `ma-image` target instead which
+uses `docker buildx` to build and push the image to registry in a single step.
```
Molter73, line 41: Why do we need to push directly? Why can't we just build the image in our makefiles then push them in CI the way we do with the existing builds?

Makefile
```diff
@@ -41,9 +43,27 @@ else
 	docker pull quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG)
 endif
 
+.PHONY: ma-builder
+ma-builder:
+ifdef BUILD_BUILDER_IMAGE
+	docker buildx build --push --platform ${ARCH} \
```
Molter73, line 49: Related to my previous comment, seems like if we were to remove the `--push` option here we could keep a single target instead of duplicating everything. Also, would you mind elaborating on why we need to use buildx? Why can't we keep the regular `docker build` and let people choose what they want to use for building with the `DOCKER_BUILDKIT` env variable?

Makefile
```diff
@@ -69,6 +89,16 @@ image: collector unittest
 		-t quay.io/stackrox-io/collector:$(COLLECTOR_TAG) \
 		$(COLLECTOR_BUILD_CONTEXT)
 
+ma-image: ma-collector unittest
+	make -C collector txt-files
+	docker buildx build --push --platform ${ARCH} \
```
Molter73, line 94: Same as the previous comment.

Molter73 (commented):
Hey @anup-kodlekere! Thanks for this PR! I only have a few questions on some choices made here.

docs/how-to-start.md
anup-kodlekere: We need to push directly because docker does not allow loading the multi-arch manifest into it's engine (`docker image ls`) at the current time. To work around this, buildx pushes the multi-arch manifest directly to a registry.

Makefile
anup-kodlekere: Could you please provide an example of how that can be achieved? Because as far as I'm aware, docker build will only create a platform specific image, the one of the host.

anup-kodlekere:
@Molter73 I've removed the duplicate targets and introduced a few more changes. 
PTAL: https://github.com/stackrox/collector/actions/runs/4667373418/jobs/8263139202
I still need to figure out a way to push to both `stackrox-io` and `rhacs-eng`, so will look into that.

.github/workflows/collector-slim.yml
```diff
@@ -33,7 +34,9 @@ jobs:
         if: github.event_name != 'pull_request'
         run: |
           echo 'BUILD_BUILDER_IMAGE=true' >> "$GITHUB_ENV"
+          echo 'BUILD_MULTIARCH_IMAGE=true' >> "$GITHUB_ENV"
           echo "COLLECTOR_BUILDER_TAG=${DEFAULT_BUILDER_TAG}" >> "$GITHUB_ENV"
```
Molter73, line 38: There's a shellcheck lint error being reported here because there's more than 2 echo commands appending to the same file, could you rewrite it like this?
```suggestion
          {
            echo 'BUILD_BUILDER_IMAGE=true'
            echo 'BUILD_MULTIARCH_IMAGE=true'
            echo "COLLECTOR_BUILDER_TAG=${DEFAULT_BUILDER_TAG}"
          } >> "$GITHUB_ENV"
```

You can test this locally with [actionlint](https://github.com/rhysd/actionlint)

Molter73 (commented):
> I still need to figure out a way to push to both stackrox-io and rhacs-eng, so will look into that.

I'd assume the local intermediate layers are still available, so maybe you can add an`REGISTRY_ORGANIZATION` variable that defaults to `stackrox-io`, then `export REGISTRY_ORGANIZATION=rhacs-eng` and re-run the build steps, if the intermediate layers are there it should cache everything, tag the new images and push, right?

anup-kodlekere:
> > I still need to figure out a way to push to both stackrox-io and rhacs-eng, so will look into that.
> 
> I'd assume the local intermediate layers are still available, so maybe you can add an`REGISTRY_ORGANIZATION` variable that defaults to `stackrox-io`, then `export REGISTRY_ORGANIZATION=rhacs-eng` and re-run the build steps, if the intermediate layers are there it should cache everything, tag the new images and push, right?

Yes, I believe it should. I was checking that option also. Will push the changes and check the CI run.

Molter73 (commented):
The image build is working, but it looks like the ppc64le image might be using the x86 binary in it, there will need to be extra changes to the `collector` target [here](https://github.com/stackrox/collector/blob/fe1898c70670ed1708e71150e794fef1ef4920a2/collector/Makefile#L20-L35)

Molter73:
>Do you think building and pushing a multi-arch collector-builder will fix the issue?
>
>That, and also maybe using --platform for running platform specific build with docker run?.

You will need the builder for every platform, you'll need to use the `docker run --platform` for every arch, which will need to dump the build on separate directories and finally build the images with the corresponding arch.

anup-kodlekere:
> > Do you think building and pushing a multi-arch collector-builder will fix the issue?
> > That, and also maybe using --platform for running platform specific build with docker run?.
> 
> You will need the builder for every platform, you'll need to use the `docker run --platform` for every arch, which will need to dump the build on separate directories and finally build the images with the corresponding arch.

Hmm. I'm unaware of the inner workers of collector Makefiles. What is your opinion on using two parallel jobs for building architecture specific image and then one job to merge the result into one?

anup-kodlekere:
@Molter73 so I'm continuing with the first option i.e to make arch-specific directories and while running the unit tests, `AbortHandler` fails on qemu-emulation. If I run the test on native ppc hardware, it passes. What do you suggest we do here?

```
/tmp/collector/collector/test/AbortHandler_test.cpp:38: Failure
Death test: test_crash()
    Result: died but not with expected error.
  Expected: contains regular expression ".*AbortHandler.*AbortHandler_Crash_Test.*"
Actual msg:
[  DEATH   ]
[  FAILED  ] AbortHandler.Crash (123 ms)
[----------] 1 test from AbortHandler (124 ms total)
```

anup-kodlekere:
> Are you sure the binary being used is the ppc64le one?

Yes, it should be.

> Is that the only test being run? Could you try removing the test and running to see if that is the only one failing?

No, I'm running the full test suite, and that is the only test failing. If I change the regex from `".*AbortHandler.*AbortHandler_Crash_Test.*"` to `""`, the test passes. 

Perhaps I should push my commits and you can take a look at the CI run?

Molter73:
Looking at the amount of duplication going on for the ppc builds, I would strongly suggest looking into the matrix build strategy and build containers for different archs on parallel runners, it'll probably be a lot more clean and enable adding other archs in an easier way (arm and s390x are likely coming soonish).

That said, looks like the binaries being used are correct, we'll see what happens once CI finishes running.

anup-kodlekere:
Yeah I think the parallel builds should be much simpler to implement and also manage. Do you want me to continue in the same PR or should we close this with appropriate reasoning and work on a new PR?

Molter73:
> Yeah I think the parallel builds should be much simpler to implement and also manage. Do you want me to continue in the same PR or should we close this with appropriate reasoning and work on a new PR?

You can do either, personally I would keep going in this PR because I already have changes in here that might be useful and simply squash commits to make slightly adjust the history if needed, but you can go ahead and open a different PR if it's easier for you.

anup-kodlekere:
@Molter73 PTAL at the new changes and the CI run. Like I said, the `AbortHandler` is the only test that fails while running on emulation.

anup-kodlekere:
I made a change to `AbortHandler` which skips the test if detected platform is ppc64le:

```c
TEST(AbortHandler, Crash) {

  #if defined(__powerpc) || defined(__powerpc__) || defined(__powerpc64__) || defined(__POWERPC__) || defined(__ppc__) || defined(__PPC__) || defined(_ARCH_PPC)
    GTEST_SKIP() << "Skipping AbortHandler test on PowerPC";
  #endif

  // Install the AbortHandler, and verify that the stderr output will contain
  // something that looks like a stacktrace with AbortHander & Crash_Test frames.
  signal(SIGSEGV, AbortHandler);
  ASSERT_EXIT(test_crash(), ::testing::KilledBySignal(SIGSEGV),
              ".*AbortHandler.*AbortHandler_Crash_Test.*");
}
```

Molter73:
Hey @anup-kodlekere! Sorry for the delay.

For the time being and in the interest of moving the PR forward, the skip looks ok, I'd appreciate if we can narrow it down to check just one of those defines though.

anup-kodlekere:
@Molter73  I've got a few more questions. Currently the build is failing while creating multi-arch manifest for collector-builder. 

```
Run docker manifest create "quay.io/stackrox-io/collector-builder:${COLLECTOR_BUILDER_TAG}" \
error parsing name for manifest list quay.io/stacrox-io/collector-builder:: invalid reference format
Error: Process completed with exit code 1.
```

I initially assumed it was just syntactical and a wrapping the command around quotes would be enough, but it didn't help. I'm guessing it's because the `COLLECTOR_BUILDER_TAG`  env var isn't getting de-referenced, but don't know why. Any clue here?

Secondly, what would it take to make the `collector-builder:cache` image multi-arch? I don't see any explicit workflow that builds that image, or maybe I haven't looked hard enough. Thanks!

Molter73:
Hi @anup-kodlekere,

Environment variables set in a job are only available in that job, once you move on to a different job it gets removed. I'd say the easiest way to solve it in this case is to set the variable as an output of a job and then use it in the following steps, think it would look something like this:
```yaml
jobs:
  set-environment:
    name: Set environment variables
    runs-on: ubuntu-latest
    outputs:
      build-builder-image: ${{ steps.set-env.outputs.build-builder-image }}
      collector-builder-tag: ${{ steps.set-env.outputs.collector-builder-tag }}

      - name: Checks for main and release branches
        id: set-env
        run: |
          echo 'build-builder-image=true' >> "$GITHUB_OUTPUT"
          echo "collector-builder-tag=${DEFAULT_BUILDER_TAG}" >> "$GITHUB_OUTPUT"

# Both the PR and master checks will need to be unified in the previous step

  build-collector-image:
    name: Build the collector slim image
    needs: set-environment
    runs-on: ubuntu-latest
    strategy:
      matrix:
        arch: [amd64, ppc64le]
    env:
      BUILD_BUILDER_IMAGE: ${{ needs.set-environment.outputs.build-builder-image }}
      COLLECTOR_BUILDER_TAG: ${{ needs.set-environment.outputs.collector-builder-tag }}

# Set up buildx, qemu, build image and push

  create-multiarch-manifest:
    needs:
    - set-environment
    - build-collector-image
    name: Create Multiarch manifest
    runs-on: ubuntu-latest
    env:
      BUILD_BUILDER_IMAGE: ${{ needs.set-environment.outputs.build-builder-image }}
      COLLECTOR_BUILDER_TAG: ${{ needs.set-environment.outputs.collector-builder-tag }}
```

See here for more info on defining ouptuts: https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

As for the builder image, the `make image` target already depends on the builder target, so it gets build before the collector image, your changes should be enough for it to be building for power.

.github/workflows/collector-slim.yml
```diff
@@ -20,44 +20,69 @@ env:
...
+    steps:
+      - name: Checks PR, main and release branches
+        id: set-env
+        run: |
+
+          if [[ "${{ github.event_name != 'pull_request' }}" ]]; then
+            echo 'build-builder-image=true' >> "$GITHUB_OUTPUT"
+            echo "collector-builder-tag=${DEFAULT_BUILDER_TAG}" >> "$GITHUB_OUTPUT"
+
+          elif [[ "${{ github.event_name == 'pull_request' }}" ]]; then
```
Molter73, line 39: Similar to the previous comment
```suggestion
          elif [[ "${{ github.event_name }}" == 'pull_request' ]]; then
```

.github/workflows/collector-slim.yml
```diff
@@ -20,44 +20,69 @@ env:
...
+    steps:
+      - name: Checks PR, main and release branches
+        id: set-env
+        run: |
+
+          if [[ "${{ github.event_name }}" != 'pull_request' ]]; then
+            echo 'build-builder-image=true' >> "$GITHUB_OUTPUT"
+            echo "collector-builder-tag=${DEFAULT_BUILDER_TAG}" >> "$GITHUB_OUTPUT"
+
+          elif [[ "${{ github.event_name }}" == 'pull_request' ]]; then
```
Molter73, line 39: I believe we can do `else` directly here, since it will either not be `pull_request` and enter the first code block or be `pull_request` and enter this condition.
```suggestion
          else
```

.github/workflows/collector-slim.yml
```diff
@@ -20,44 +20,69 @@ env:
...
           submodules: true
 
-      - name: Checks for main and release branches
-        if: github.event_name != 'pull_request'
+      - name: Log env vars
         run: |
-          echo 'BUILD_BUILDER_IMAGE=true' >> "$GITHUB_ENV"
-          echo "COLLECTOR_BUILDER_TAG=${DEFAULT_BUILDER_TAG}" >> "$GITHUB_ENV"
+          echo "[LOG] - $BUILD_BUILDER_IMAGE"
+          echo "[LOG] - $COLLECTOR_BUILDER_TAG"
```
Molter73, line 79: [nitpick] Environment variables are already logged in GHA if you expand the command: 
image

Feel free to keep this step, but I would remove it.

.github/workflows/collector-slim.yml
```diff
@@ -67,41 +92,120 @@ jobs:
...
+          password: ${{ secrets.QUAY_STACKROX_IO_RW_PASSWORD }}  
+
+
+      - name: Create and push multiarch manifest for stackrox-io -slim
+        run: |
+          docker manifest create quay.io/stackrox-io/collector:${{ inputs.collector-tag }}-slim \
+                                 quay.io/stackrox-io/collector:${{ inputs.collector-tag }}-amd64-slim \
+                                 quay.io/stackrox-io/collector:${{ inputs.collector-tag }}-ppc64le-slim
+
+          docker manifest push quay.io/stackrox-io/collector:${{ inputs.collector-tag }}-slim
```
Molter73, line 161: [ultra-nit] We can probably refactor this into a small reusable action, since there are multiple blocks doing the same actions and only changing the image used. We can take care of it in a follow up PR though, not a blocker for this.

Makefile
```diff
@@ -30,15 +32,15 @@ container-dockerfile-dev:
 .PHONY: builder
 builder:
 ifdef BUILD_BUILDER_IMAGE
-	docker build \
+	docker buildx build --load --platform ${ARCH} \
```
Molter73, line 35: Do we absolutely need `buildx` here? The legacy builder is deprecated and, AFAICT, docker now defaults to buildkit. Keeping it here seems harmless but might interfere with the workflow of people using `docker` symlinked to `podman`.
```suggestion
	docker build --load --platform ${ARCH} \
```

Makefile
```diff
@@ -8,6 +8,8 @@ MOD_VER_FILE=$(CURDIR)/kernel-modules/kobuild-tmp/MODULE_VERSION.txt
 LOCAL_SSH_PORT ?= 2222
 DEV_SSH_SERVER_KEY ?= $(CURDIR)/.collector_dev_ssh_host_ed25519_key
 
+ARCH ?= amd64
```
Molter73, line 11: Please move this line to `Makefile-constants.mk`, otherwise running commands directly in the `collector/` subdirectory will break.

collector/test/AbortHandler_test.cpp
```diff
@@ -31,6 +31,11 @@ void test_crash() {
 }
 
 TEST(AbortHandler, Crash) {
+
+  #if defined(__powerpc64__)
+    GTEST_SKIP() << "Skipping AbortHandler test on ppc64le";
+  #endif
```
Molter73, line 37: There's a linting error here, running `make -C collector format` should solve it.

.github/workflows/collector-slim.yml
anup-kodlekere: Cool. I initially had kept it simply as `else` but thought of using an explicit condition. Will fix.

.github/workflows/collector-slim.yml
anup-kodlekere: Will fix in a follow up PR.

Makefile
anup-kodlekere: Okay. I just checked and it seems buildkit _is_ the default builder for docker. I can make this change. I assume the same needs to be done for the `image:`  target as well?

anup-kodlekere:
@Molter73 The default docker version in Actions is too low for builtdkit to be the default builder. 

`Version:           20.10.23+azure-2`

```
docker build --load --platform amd64 \
	--build-arg NPROCS=2 \
	--cache-from quay.io/stackrox-io/collector-builder:cache \
	--cache-from quay.io/stackrox-io/collector-builder:3.14.x-66-gd955a98fb1 \
	-t quay.io/stackrox-io/collector-builder:3.14.x-66-gd955a98fb1 \
	-f "/home/runner/work/collector/collector/builder/Dockerfile" \
	.
unknown flag: --load
See 'docker build --help'.
make: *** [Makefile:33: builder] Error 125
```
> [BuildKit](https://github.com/moby/buildkit) is an improved backend to replace the legacy builder. BuildKit is the default builder for users on Docker Desktop, and Docker Engine as of version 23.0.

Molter73:
> @Molter73 The default docker version in Actions is too low for builtdkit to be the default builder.
> 
> `Version: 20.10.23+azure-2`
> 
> ```
> docker build --load --platform amd64 \
> 	--build-arg NPROCS=2 \
> 	--cache-from quay.io/stackrox-io/collector-builder:cache \
> 	--cache-from quay.io/stackrox-io/collector-builder:3.14.x-66-gd955a98fb1 \
> 	-t quay.io/stackrox-io/collector-builder:3.14.x-66-gd955a98fb1 \
> 	-f "/home/runner/work/collector/collector/builder/Dockerfile" \
> 	.
> unknown flag: --load
> See 'docker build --help'.
> make: *** [Makefile:33: builder] Error 125
> ```
> 
> > [BuildKit](https://github.com/moby/buildkit) is an improved backend to replace the legacy builder. BuildKit is the default builder for users on Docker Desktop, and Docker Engine as of version 23.0.

Yeah, I just caught that, the `--load` option also doesn't seem to be supported by `podman build`, so we are already making a breaking change. Rollback to using `docker buildx build` and we'll deal with people failing to build collector with other tooling as they show up.

Molter73:
The latest failure is unrelated to your changes, there's some change on one of submodules that's causing the error.

Molter73:
I've opened stackrox/falcosecurity-libs#45, I'll open a PR based on this one that uses the branch on that PR for testing. If everything goes well, we'll merge that PR and re-running here should be enough to fix the error.

anup-kodlekere:
Okay. Also I've prepared the re-usable action to create and push manifests. Adding the changes here so you can take a look and decide whether we want it in this PR or some other:

create-and-push-manfest/action.yml:
```
name: create and push multiarch collector manifests
description: |
  This action will create multiarch manifests and push to the registries.

inputs:
  registry-org:
    description: |
      The registry in which the manifest will be pushed (stackrox-io/rhacs-eng)
    required: true
  tag:
    description: |
      The tag for the manifest.
    required: false
  ma-image:
    description: |
      The final multi-arch manifest image that will be created by combining
      arch specific images.
    required: true
  amd64-image:
    description: |
      The amd64 part of the multi-arch manifest.
    required: true
  ppc64le-image:
    description: |
      The ppc64le part of the multi-arch manifest.
    required: true
  username:
    description: |
      Username used to login to the repository the push is aimed at
    required: true
  password:
    description: |
      Password used to login to the repository the push is aimed at
    required: true

runs:
  using: "composite"
  steps:
    - uses: docker/login-action@v2
      with:
        registry: quay.io
        username: ${{ inputs.username }}
        password: ${{ inputs.password }}

    - shell: bash
      run: |

        docker manifest create "${{ inputs.registry-org }}/${{ inputs.ma-image }}${{ inputs.tag }} \ 
                                ${{ inputs.registry-org }}/${{ inputs.amd64-image }}${{ inputs.tag }} \
                                ${{ inputs.registry-org }}/${{ inputs.ppc64le-image }}${{ inputs.tag }}"

        docker manifest push "${{ inputs.registry-org }}/${{ inputs.ma-image }}${{ inputs.tag }}"


```


collector-slim.yml:
```
  create-multiarch-manifest:
    needs:
    - set-environment
    - build-collector-image
    name: Create Multiarch manifest
    runs-on: ubuntu-latest
    env:
      BUILD_BUILDER_IMAGE: ${{ needs.set-environment.outputs.build-builder-image }}
      COLLECTOR_BUILDER_TAG: ${{ needs.set-environment.outputs.collector-builder-tag }}

    steps:
      - name: Create and push multiarch manifest for stackrox-io -slim
        uses: ./.github/actions/create-and-push-manifest
        with:
          registry-org: quay.io/stackrox-io
          tag: -slim
          ma-image: collector:${{ inputs.collector-tag }}
          amd64-image: collector:${{ inputs.collector-tag }}-amd64
          ppc64le-image: collector:${{ inputs.collector-tag }}-ppc64le
          username: ${{ secrets.QUAY_STACKROX_IO_RW_USERNAME }}
          password: ${{ secrets.QUAY_STACKROX_IO_RW_PASSWORD }}


      - name: Create and push multiarch manifest for stackrox-io -base
        uses: ./.github/actions/create-and-push-manifest
        with:
          registry-org: quay.io/stackrox-io
          tag: -base
          ma-image: collector:${{ inputs.collector-tag }}
          amd64-image: collector:${{ inputs.collector-tag }}-amd64
          ppc64le-image: collector:${{ inputs.collector-tag }}-ppc64le
          username: ${{ secrets.QUAY_STACKROX_IO_RW_USERNAME }}
          password: ${{ secrets.QUAY_STACKROX_IO_RW_PASSWORD }}


      - name: Create and push multiarch manifest for builder to stackrox-io
        uses: ./.github/actions/create-and-push-manifest
        with:
          registry-org: quay.io/stackrox-io
          ma-image: collector-builder:${COLLECTOR_BUILDER_TAG}
          amd64-image: collector-builder:${COLLECTOR_BUILDER_TAG}-amd64
          ppc64le-image: collector-builder:${COLLECTOR_BUILDER_TAG}-ppc64le
          username: ${{ secrets.QUAY_STACKROX_IO_RW_USERNAME }}
          password: ${{ secrets.QUAY_STACKROX_IO_RW_PASSWORD }}

      
      - name: Create and push multiarch manifest for rhacs-eng -slim
        uses: ./.github/actions/create-and-push-manifest
        with:
          registry-org: quay.io/rhacs-eng
          tag: -slim
          ma-image: collector:${{ inputs.collector-tag }}
          amd64-image: collector:${{ inputs.collector-tag }}-amd64
          ppc64le-image: collector:${{ inputs.collector-tag }}-ppc64le


      - name: Create and push multiarch manifest for rhacs-eng -base
        uses: ./.github/actions/create-and-push-manifest
        with:
          registry-org: quay.io/rhacs-eng
          tag: -base
          ma-image: collector:${{ inputs.collector-tag }}
          amd64-image: collector:${{ inputs.collector-tag }}-amd64
          ppc64le-image: collector:${{ inputs.collector-tag }}-ppc64le

      - name: Create and push multiarch manifest for builder to rhacs-eng
        uses: ./.github/actions/create-and-push-manifest
        with:
          registry-org: quay.io/rhacs-eng
          ma-image: collector-builder:${COLLECTOR_BUILDER_TAG}
          amd64-image: collector-builder:${COLLECTOR_BUILDER_TAG}-amd64
          ppc64le-image: collector-builder:${COLLECTOR_BUILDER_TAG}-ppc64le
```

collector/CMakeLists.txt
```diff
@@ -115,7 +115,18 @@ set(DRIVER_DEVICE_NAME "${DRIVER_NAME}" CACHE STRING "Driver device name" FORCE)
...
 
+
+# ppc64le is not supported for the "modern" probe
+if(NOT CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "ppc64le")
+	set(BUILD_LIBSCAP_MODERN_BPF ON CACHE BOOL "Enable modern bpf engine" FORCE)
+endif()
+
+set(MODERN_BPF_EXCLUDE_PROGS "^(openat2|ppoll|setsockopt|getsockopt|clone3|io_uring_setup|nanosleep)$" CACHE STRING "Set of syscalls to exclude from modern bpf engine " FORCE)
+
+add_subdirectory(${PROJECT_SOURCE_DIR}/falcosecurity-libs EXCLUDE_FROM_DEFAULT_BUILD)
```
Molter73, line 126: These lines are duplicated with 129 to the end of the file, please remove these 2.

collector/CMakeLists.txt
```diff
@@ -115,7 +115,18 @@ set(DRIVER_DEVICE_NAME "${DRIVER_NAME}" CACHE STRING "Driver device name" FORCE)
 set(SCAP_BPF_PROGS_MAX "256" CACHE STRING "Number of BPF progams supported in Falco" FORCE)
 set(SCAP_HOST_ROOT_ENV_VAR_NAME "COLLECTOR_HOST_ROOT" CACHE STRING "Host root environment variable name" FORCE)
 
+
+# ppc64le is not supported for the "modern" probe
+if(NOT CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "ppc64le")
+	set(BUILD_LIBSCAP_MODERN_BPF ON CACHE BOOL "Enable modern bpf engine" FORCE)
+endif()
```
Molter73, line 122: These lines are meant to replace 128. Please remove line 128.

Molter73 (approved):
Thanks for all the effort in this @anup! LGTM!

PR #1113, opened 2023-04-14
Fix connscraper capabilities

Molter73:
## Description

connscraper needs DAC_OVERRIDE in order to read /proc correctly, so when using kernel modules we drop the capability after connscraper spawns.

## Checklist

## Testing Performed

- [x] Test the image in the main repo.


Stringy (approved):
LGTM!

PR #1116, opened 2023-04-17
Jv remove kernel module collection from integration tests

JoukoVirtanen:
## Description

Remove kernel modules from ansible playbooks.

## Checklist
- [ ] Investigated and inspected CI test results

## Testing Performed

CI is sufficient

PR #1117, opened 2023-04-18
Module version 2.5.0 / kernel module blocking

Stringy:
## Description

To set up for kernel module removal, block kernel module builds on new module versions. Also bump module version to 2.5.0

NOTE: do not merge until the following have merged (the main repo/OSCI PRs are required because some tests will begin failing if we bump the collector version with this new module version in it)

https://github.com/stackrox/stackrox/pull/5710
https://github.com/openshift/release/pull/38403
https://github.com/stackrox/collector/pull/1116


kernel-modules/BLOCKLIST
```diff
@@ -66,3 +66,7 @@
 6.2.* 2.2.0 *
 6.2.* 2.1.0 *
 6.2.* 2.0.1 *
+
```
Molter73, line 69: I would not leave an empty line here, should be harmless, but may also not be harmless (pick performance review comment, I know).

Molter73 (commented):
I think we should rollback the module version once we are done testing if we are not changing anything yet. At least it would be pretty odd to me if we are building 2 driver versions that are exactly the same, but incompatible according to its version.

Molter73 (approved):
LGTM!

PR #1125, opened 2023-04-20
Fix docker-compose example in docs

vladbologa:
## Description

Updated an example in the docs

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Tested by running `docker-compose up` using the yaml file in the docs.

### DIFF
--- a/docs/how-to-start.md
+++ b/docs/how-to-start.md
@@ -85,6 +85,10 @@ the host network in privileged mode. Collector in turn will try to download
 probes for your version of Linux kernel and start listening to events happening
 inside the container.
 
+The image `quay.io/rhacs-eng/grpc-server` can be built manually by running 
+`make mock-grpc-server-image` in the `https://github.com/stackrox/stackrox/`
+repository.
+
 ### Development with an IDE (CLion)
 
 #### Setup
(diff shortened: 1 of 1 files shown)


erthalion (approved):

docs/how-to-start.md
```diff
@@ -67,10 +67,12 @@ services:
       - /usr/lib:/host/usr/lib:ro
       - /sys/:/host/sys/:ro
       - /dev:/host/dev:ro
+      - type: tmpfs
+        target: /module
     depends_on:
       - grpc-server-debug
   grpc-server-debug:
-    image: stackrox/grpc-server:latest
+    image: quay.io/rhacs-eng/grpc-server:latest
```
Molter73, line 75: This is a private image unfortunately, so people outside the rhacs-eng org won't be able to use it. Maybe we can add a comment pointing at the main repo make target for people to know how to build it from source? In my experience that is usually fast enough.

vladbologa: Good point, I added a clarification below.

Molter73 (approved):
LGTM!

PR #1127, opened 2023-04-21
Improve compatibility with podman

Molter73:
## Description

A small set of changes that enables using podman as the build engine via the podman-docker package or by symlinking docker to podman.

The collector/falcosecurity-libs symlink is removed. It is not needed now that we use a global cmake project set at the root of the repository and was causing trouble with podman.

The '--cache-from' argument takes different types of arguments for podman and docker, since the way the builder works makes it relatively hard to actually take advantage of caching, we can directly remove them.

## Checklist

## Testing Performed
- [x] Run with `no-cache` to ensure removing the symlink doesn't affect driver builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,8 +32,6 @@ builder:
 ifdef BUILD_BUILDER_IMAGE
 	docker buildx build --load --platform ${PLATFORM} \
 		--build-arg NPROCS=$(NPROCS) \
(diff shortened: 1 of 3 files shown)


ovalenti (approved):
It runs smooth here !

Molter73:
The check-failures error is due to some debian packages being deleted by our crawler, it's a known issue and unrelated to these changes.

PR #1129, opened 2023-04-24
Adjust BTF adr

erthalion:
## Description

Include self-checks details, heuristics hard-stop.


ovalenti (approved):
It looks good to me !

PR #1138, opened 2023-04-25
Add an option to hard fail when the CO-RE bpf probe fails to insert

### LINKED ISSUE #1137: Temporary option to not fallback for MVP
Temporary do not fallback to eBPF if modern probes are failed to load. Turns out that e.g. verifier error is not preventing Collector from continuing. Note, that in general it's good, just not for MVP.

Part of #1011

Molter73:
## Description

As we've discussed on some weekly syncs, we should hard fail when the CO-RE bpf probe fails to insert in order to get more visibility on problems that might arise. A configuration option to rollback the behavior to fallback on regular eBPF is added.

Fixes #1137 

## Checklist

## Testing Performed
CI should be enough.


erthalion (approved):

Molter73:
- [ ] Rollback [ea24aae](https://github.com/stackrox/collector/pull/1138/commits/ea24aae6e8308f48c7d1e58b84fd7b7ac596515d) before merging.

PR #1150, opened 2023-04-27
Jv rox 16789 various collector integration tests for connections and endpoints

JoukoVirtanen:
## Description

Tests for connections and endpoints. Two containers with socat are started. One is the server and one is the container. The server container runs a command to listen for incoming messages. The client runs a command to send a message to the server. Many different variations of these commands can be easily tested and compared against expected results.

## Checklist
- [x] Some tests fail locally when run with kernel module probe

**Automated testing**
  - [x] Added integration tests

## Testing Performed
make -C integration-tests TestConnectionsAndEndpointsNormal
make -C integration-tests TestConnectionsAndEndpointsHigh

The following fails when run with kernel module probe
make -C integration-tests TestConnectionsAndEndpointsSourcePort
make -C integration-tests TestConnectionsAndEndpointsHighLowPorts


integration-tests/suites/connections-and-endpoints.go
```diff
@@ -0,0 +1,116 @@
...
+	val, err = s.Get(s.Server.ContainerID, networkBucket)
+	s.Require().NoError(err)
+	serverNetwork, err := common.NewNetworkInfo(val)
+	s.Require().NoError(err)
+	expectedLocalAddress = strings.Replace(s.Server.ExpectedNetwork[0].LocalAddress, "SERVER_IP", s.Server.IP, -1)
+	expectedRemoteAddress = strings.Replace(s.Server.ExpectedNetwork[0].RemoteAddress, "CLIENT_IP", s.Client.IP, -1)
+	assert.Equal(s.T(), expectedLocalAddress, serverNetwork.LocalAddress)
+	assert.Equal(s.T(), expectedRemoteAddress, serverNetwork.RemoteAddress)
+	assert.Equal(s.T(), "ROLE_SERVER", serverNetwork.Role)
+	assert.Equal(s.T(), s.Server.ExpectedNetwork[0].SocketFamily, serverNetwork.SocketFamily)
```
JoukoVirtanen, line 114: Still need to add asserts for endpoints.

integration-tests/suites/connections-and-endpoints.go
```diff
@@ -0,0 +1,115 @@
...
+	s.Client.ContainerID = common.ContainerShortID(longContainerID)
+	_, err = s.execContainer(serverName, []string{"/bin/sh", "-c", s.Server.Cmd})
+	s.Require().NoError(err)
+	time.Sleep(3 * time.Second)
+	s.Server.IP, err = s.getIPAddress(serverName)
+	s.Client.IP, err = s.getIPAddress(clientName)
+	s.Require().NoError(err)
+	clientCmd := strings.Replace(s.Client.Cmd, "SERVER_IP", s.Server.IP, -1)
+	_, err = s.execContainer(clientName, []string{"/bin/sh", "-c", clientCmd})
+	s.Require().NoError(err)
```
Stringy, line 67: Some empty lines here would make this a lot more readable, e.g.:

```suggestion
	serverName := s.Server.Name
	clientName := s.Client.Name
	
	longContainerID, err := s.launchContainer(serverName, socatImage, "/bin/sh", "-c", "/bin/sleep 300")
	s.Server.ContainerID = common.ContainerShortID(longContainerID)
	s.Require().NoError(err)
	
	longContainerID, err = s.launchContainer(clientName, socatImage, "/bin/sh", "-c", "/bin/sleep 300")
	s.Client.ContainerID = common.ContainerShortID(longContainerID)
	
	_, err = s.execContainer(serverName, []string{"/bin/sh", "-c", s.Server.Cmd})
	s.Require().NoError(err)
	
	time.Sleep(3 * time.Second)
	
	s.Server.IP, err = s.getIPAddress(serverName)
	s.Client.IP, err = s.getIPAddress(clientName)
	s.Require().NoError(err)
	
	clientCmd := strings.Replace(s.Client.Cmd, "SERVER_IP", s.Server.IP, -1)
	_, err = s.execContainer(clientName, []string{"/bin/sh", "-c", clientCmd})
	s.Require().NoError(err)
```

integration-tests/suites/connections-and-endpoints.go
```diff
@@ -0,0 +1,115 @@
...
+	clientName := s.Client.Name
+	longContainerID, err := s.launchContainer(serverName, socatImage, "/bin/sh", "-c", "/bin/sleep 300")
+	s.Server.ContainerID = common.ContainerShortID(longContainerID)
+	s.Require().NoError(err)
+	longContainerID, err = s.launchContainer(clientName, socatImage, "/bin/sh", "-c", "/bin/sleep 300")
+	s.Client.ContainerID = common.ContainerShortID(longContainerID)
+	_, err = s.execContainer(serverName, []string{"/bin/sh", "-c", s.Server.Cmd})
+	s.Require().NoError(err)
+	time.Sleep(3 * time.Second)
+	s.Server.IP, err = s.getIPAddress(serverName)
```
Stringy, line 62: we're not handling this error

integration-tests/suites/connections-and-endpoints.go
```diff
@@ -0,0 +1,115 @@
...
+	time.Sleep(30 * time.Second)
+
+	socatImage := common.QaImage("quay.io/rhacs-eng/qa", "socat")
+
+	serverName := s.Server.Name
+	clientName := s.Client.Name
+	longContainerID, err := s.launchContainer(serverName, socatImage, "/bin/sh", "-c", "/bin/sleep 300")
+	s.Server.ContainerID = common.ContainerShortID(longContainerID)
+	s.Require().NoError(err)
+	longContainerID, err = s.launchContainer(clientName, socatImage, "/bin/sh", "-c", "/bin/sleep 300")
```
Stringy, line 57: we're not handling this error

integration-tests/suites/connections-and-endpoints.go
```diff
@@ -0,0 +1,115 @@
...
+	s.PrintContainerStats(stats)
+	s.WritePerfResults("ConnectionsAndEndpoints", stats, s.metrics)
+}
+
+func (s *ConnectionsAndEndpointsTestSuite) TestConnectionsAndEndpoints() {
+	fmt.Println()
+	fmt.Println("serverName= ", s.Server.Name)
+	fmt.Println("clientName= ", s.Client.Name)
+	fmt.Println("serverCmd= ", s.Server.Cmd)
+	fmt.Println("clientCmd= ", s.Client.Cmd)
```
Stringy, line 91: probably want to remove these before merging

Stringy (commented):
Couple of minor points, but generally looking good

integration-tests/suites/connections-and-endpoints.go
JoukoVirtanen: Done

integration-tests/suites/connections-and-endpoints.go
JoukoVirtanen: Handled

integration-tests/suites/connections-and-endpoints.go
JoukoVirtanen: Handled

integration-tests/suites/connections-and-endpoints.go
```diff
@@ -0,0 +1,177 @@
...
+	})
+	sort.Slice(serverEndpoints, func(i, j int) bool { return endpointComparison(serverEndpoints[i], serverEndpoints[j]) })
+
+	for idx := range serverEndpoints {
+		assert.Equal(s.T(), s.Server.ExpectedEndpoints[idx].Protocol, serverEndpoints[idx].Protocol)
+		assert.Equal(s.T(), s.Server.ExpectedEndpoints[idx].Address, serverEndpoints[idx].Address)
+	}
+}
+
+func endpointComparison(endpoint1 common.EndpointInfo, endpoint2 common.EndpointInfo) bool {
```
Stringy, line 140: Is it possible to simplify this function? I'm not sure we need to compare all the fields to produce a deterministic sort order for the two lists. Perhaps something like:

```go
func endpointComparison(endpoint1, endpoint2 common.EndpointInfo) bool {
    if endpoint1.Address == nil {
        return false
    }
    if endpoint2.Address == nil {
        return true
    }
    
   addr1, addr2 := endpoint1.Address, endpoint2.Address
    return (addr1.AddressData < addr2.AddressData) && (addr1.Port < addr2.Port)
}
```

JoukoVirtanen: I think we need to compare the protocol as well. What if there are two endpoints that use the same port, but different protocols? The order would not be defined. I am not sure that we need IpNetwork, so I will take it out. Also I can do the assignment 

`addr1, addr2 := endpoint1.Address, endpoint2.Address`

I think that 
`(addr1.AddressData < addr2.AddressData) && (addr1.Port < addr2.Port)`

Is not correct. What if `addr1.AddressData < addr2.AddressData` is true, but `addr1.Port < addr2.Port` is false. In that case switching `endpoint1` and `endpoint2` will also be false. The function should have the property that `endpointComparison(endpoint1, endpoint2)` returns the opposite of `endpointComparision(endpoint2, endpoint1)`, but that is not true for your suggested code change.

Stringy: yeah, the snippet I posted wasn't tested; it was just a suggestion that we could sort by address and port rather than everything. Adding Protocol into the mix makes sense, as you suggest

PR #1159, opened 2023-05-10
ROX-14231: Enable building Oracle Linux 7 kernels

Molter73:
## Description

Before migrating to OSCI, we used to support building OL7UEK5 kernels, we had to get rid of this due to the unavailability of the image we used for the builder in that system. Since we are now building drivers on GHA and we don't have the limitation for the image, we should be able to support it again.

## Checklist

## Testing Performed

- [x] Ensure crawled kernels for OL7 UEK5 and 6 build drivers correctly.

### DIFF
--- a/.github/workflows/drivers.yml
+++ b/.github/workflows/drivers.yml
@@ -61,6 +61,9 @@ jobs:
         with:
           credentials_json: '${{ secrets.GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT }}'
 
+      - name: 'Set up Cloud SDK'
+        uses: 'google-github-actions/setup-gcloud@v1'
+
       - name: Create a mock cache
         if: ${{ !contains(github.event.pull_request.labels.*.name, 'no-cache') }}
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/init.yml
Molter73, line 1: - [ ] Rollback this file before merging

kernel-modules/KERNEL_VERSIONS
Molter73, line 1: - [ ] Rollback this file before merging

Molter73:
Looks like the service account doesn't have access to the staging bucket for bundles (testing manually does download the bundles correctly). @robbycochran could we add read permission to our service account for the `gs://stackrox-kernel-bundles-staging` bucket?

robbycochran:
I double-checked and the service account we are using does have permission to access the staging bucket. From the error, `ServiceException: 401 Anonymous caller...` -- anonymous caller may indicate we aren't using the service account credentials that are set up in a previous step?

> Looks like the service account doesn't have access to the staging bucket for bundles (testing manually does download the bundles correctly). @robbycochran could we add read permission to our service account for the `gs://stackrox-kernel-bundles-staging` bucket?

Molter73:
> I double-checked and the service account we are using does have permission to access the staging bucket. From the error, `ServiceException: 401 Anonymous caller...` -- anonymous caller may indicate we aren't using the service account credentials that are set up in a previous step?
> 
> > Looks like the service account doesn't have access to the staging bucket for bundles (testing manually does download the bundles correctly). @robbycochran could we add read permission to our service account for the `gs://stackrox-kernel-bundles-staging` bucket?

I was missing a step to setup gcloud before using it, it's now properly using the staging bucket and the drivers are built correctly AFAICT.

Molter73:
Drivers for OL7 UEK have been built here: https://github.com/stackrox/collector/actions/runs/4978555997

Stringy (approved):

PR #1160, opened 2023-05-10
Bumps download interval and timeouts

Stringy:
## Description

To improve start up robustness, the time out and interval for kernel driver downloads is increased:

- interval (1 second -> 5 seconds)
- Maximum timeout (90 seconds -> 3 minutes)
- Maximum attempts (90 -> 36)

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Tested locally and observed correct timeout / interval behavior.

### DIFF
--- a/collector/lib/GetKernelObject.cpp
+++ b/collector/lib/GetKernelObject.cpp
@@ -40,8 +40,9 @@ extern "C" {
 
 namespace collector {
 
-const int kNumDownloadRetries = 90;
-const int kMaxDownloadRetriesTime = 90;
+const int kMaxDownloadRetriesTime = 180;
+const int kMaxDownloadRetriesInterval = 5;
+const int kNumDownloadRetries = kMaxDownloadRetriesTime / kMaxDownloadRetriesInterval;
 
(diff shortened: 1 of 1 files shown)


collector/lib/GetKernelObject.cpp
```diff
@@ -40,8 +40,9 @@ extern "C" {
 
 namespace collector {
 
-const int kNumDownloadRetries = 90;
-const int kMaxDownloadRetriesTime = 90;
+const int kMaxDownloadRetriesTime = 180;
+const int kMaxDownloadRetriesInterval = 5;
+const int kNumDownloadRetries = kMaxDownloadRetriesTime / kMaxDownloadRetriesInterval;
```
erthalion, line 45: The assumption is that `max time = max retries * retry interval`, but isn't `max time = max retries * retry interval + processing time` is more correct? It doesn't break anything, but affects a bit many retries will be reached, right?

Stringy: yeah I suppose that's a fair assessment, though I'm not sure the processing time tends to be long enough to add too much variation in the overall wait time

erthalion (approved):

PR #1181, opened 2023-05-30
Convert cp command to more robust find + cp in prepare-src

Stringy:
## Description

The previous fix did not behave correctly on all shells, so this replaces the command with a crafted `find` version, which handles the expected differences between module versions.


ovalenti (approved):
Looks good to me !

PR #1186, opened 2023-06-02
Disable the import-users flag

ovalenti:
## Description

When the CRI is not available, the containers are never declared destroyed; and the user-manager keeps user information forever.

## Checklist

## Testing Performed

The memory allocation growth can be accelerated with:
```
while true; do kubectl run temp -i -t --image=debian:bookworm --rm --restart=Never --command -- sleep 1; done
```


Stringy (approved):
LGTM!

PR #1189, opened 2023-06-05
New environment variable for control set_import_users()

ovalenti:
## Description

ROX_COLLECTOR_SET_IMPORT_USERS defaults to false, and is used to configure sinsp and its user-mgmt subsystem.

This is intended as a safeguard in case we would need to reenable it; and was deactivated to circumvent a memory management issue (CRI not available => containers never freed in user.cpp)

## Checklist
- [x] Setting the env var changes the config and can be seen in the logs

### DIFF
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -57,6 +57,8 @@ BoolEnvVar set_processes_listening_on_ports("ROX_PROCESSES_LISTENING_ON_PORT", C
 
 BoolEnvVar core_bpf_hardfail("ROX_COLLECTOR_CORE_BPF_HARDFAIL", true);
 
+BoolEnvVar set_import_users("ROX_COLLECTOR_SET_IMPORT_USERS", false);
+
 }  // namespace
 
 constexpr bool CollectorConfig::kUseChiselCache;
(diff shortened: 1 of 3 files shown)


Stringy (approved):
LGTM!

erthalion (approved):
Looks good, but probably `...SET...` is not needed. `sinsp` uses it as verb, so `ROX_COLLECTOR_IMPORT_USERS` would be more straightforward.

ovalenti:
> Looks good, but probably `...SET...` is not needed. `sinsp` uses it as verb, so `ROX_COLLECTOR_IMPORT_USERS` would be more straightforward.

It makes sense. I based my choice on the _CURL_ flag, since it is similarly setting configuration for an other component, whereas flags like the core_bpf hard fail are behaviors of the collector logic itself.

erthalion:
> > Looks good, but probably `...SET...` is not needed. `sinsp` uses it as verb, so `ROX_COLLECTOR_IMPORT_USERS` would be more straightforward.
> 
> It makes sense. I based my choice on the _CURL_ flag, since it is similarly setting configuration for an other component, whereas flags like the core_bpf hard fail are behaviors of the collector logic itself.

Ah, I see. There seems to be some inconsistency here, so I don't mind it be either way.

PR #1194, opened 2023-06-08
Jv mock grpc server should report multiple connections

JoukoVirtanen:
## Description

Currently the mock grpc server only reports the latest connection that it sees for each container. We should have tests were multiple connections are seen per container. We should also make sure that when we expect to see one connection reported for a container, that there is only one connection for that container and not multiple connections. This PR updates the mock grpc server image to one that reports multiple connections per container and makes changes to the integration tests so that they get an array of connections rather than just a string representing one connection. Currently asserting that we see the expected number of connection and that all of the connections are ones that we expect to see causes many of the tests to become flaky or fail. For now we just do logging when the expected number of connections does not match with the actual number of connections and compare just the last connection that we see with the last expected connection. This will be fixed in a future PR. repeated_network_flow.go was already looking at multiple connections per container, by reading the grpc server logs, and this method of getting the connections was not changed for it.

See the corresponding PR in stackrox/stackrox https://github.com/stackrox/stackrox/pull/6308

## Checklist
- [x] Ran integration tests locally

## Testing Performed

Ran integration tests locally. Ran the following script for a couple of hours

#!/usr/bin/env bash
set -eou pipefail

while true; do
        sudo make -C integration-tests TestConnectionsAndEndpointsNormal COLLECTOR_TAG=3.15.x-92-g26a65a7bdf
        sudo make -C integration-tests TestConnectionsAndEndpointsHighLowPorts COLLECTOR_TAG=3.15.x-92-g26a65a7bdf
        sudo make -C integration-tests TestConnectionsAndEndpointsServerHigh COLLECTOR_TAG=3.15.x-92-g26a65a7bdf
        sudo make -C integration-tests TestConnectionsAndEndpointsSourcePort COLLECTOR_TAG=3.15.x-92-g26a65a7bdf
        sudo make -C integration-tests TestConnectionsAndEndpointsUDPNormal COLLECTOR_TAG=3.15.x-92-g26a65a7bdf
        sudo make -C integration-tests TestConnectionsAndEndpointsUDPNoReuseaddr COLLECTOR_TAG=3.15.x-92-g26a65a7bdf
        sudo make -C integration-tests TestConnectionsAndEndpointsUDPNoFork COLLECTOR_TAG=3.15.x-92-g26a65a7bdf
        echo 
        echo 
        echo 
        echo 
done

Also ran `make` and `sudo make -C integration-tests TestRepeatedNetworkFlow` a few times.

## Future work

Change the existing tests so that they have asserts on all expected connections and the number of expected connections. A couple of Jira tickets were created for this

https://issues.redhat.com/browse/ROX-17964
https://issues.redhat.com/browse/ROX-18803

### DIFF
--- a/integration-tests/images.yml
+++ b/integration-tests/images.yml
@@ -1,5 +1,5 @@
 images:
-  grpc-server: quay.io/rhacs-eng/grpc-server:3.73.x-95-gd43176a427
+  grpc-server: quay.io/rhacs-eng/grpc-server:4.0.x-658-g2922c23f34
 
   performance-phoronix: quay.io/rhacs-eng/collector-performance:phoronix
   performance-json-label: quay.io/rhacs-eng/collector-performance:json-label
--- a/integration-tests/suites/base.go
+++ b/integration-tests/suites/base.go
@@ -232,21 +232,6 @@ func (s *IntegrationTestSuiteBase) getPort(containerName string) (string, error)
 	return "", fmt.Errorf("no port mapping found: %v %v", rawString, portMap)
 }
 
-func (s *IntegrationTestSuiteBase) Get(key string, bucket string) (val string, err error) {
-	if s.db == nil {
-		return "", fmt.Errorf("Db %v is nil", s.db)
-	}
-	err = s.db.View(func(tx *bolt.Tx) error {
-		b := tx.Bucket([]byte(bucket))
-		if b == nil {
-			return fmt.Errorf("Bucket %s was not found", bucket)
-		}
-		val = string(b.Get([]byte(key)))
-		return nil
-	})
-	return
-}
-
 func (s *IntegrationTestSuiteBase) GetProcesses(containerID string) ([]common.ProcessInfo, error) {
 	if s.db == nil {
 		return nil, fmt.Errorf("Db %v is nil", s.db)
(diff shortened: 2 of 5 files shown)


ovalenti:
Having more than one connection reported for a test-case is absolutely something we want. However, with PR's approach, I am not sure that the validation of the result is going to be reliable. Also, it does not solve our flackyness issues with variable execution times, AFAIUI.

How about importing the grpc mock "inside" the test environment instead and dropping the database ?

With this approach:
- we would not rely on an external mock grpc server image
- we could implement scenarios based on expected events instead of relying on hypothetically timed steps
- since we would have scenarios, one test could cover much more functionality
- the individual tests could execute much faster
- the GRPC mock is written in Golang, so integrating it with the existing testing logic is hopefully "reinventing the wheel".

What do you think @JoukoVirtanen ?

JoukoVirtanen:
> Having more than one connection reported for a test-case is absolutely something we want. However, with PR's approach, I am not sure that the validation of the result is going to be reliable. Also, it does not solve our flackyness issues with variable execution times, AFAIUI.
> 
> How about importing the grpc mock "inside" the test environment instead and dropping the database ?
> 
> With this approach:
> 
> * we would not rely on an external mock grpc server image
> * we could implement scenarios based on expected events instead of relying on hypothetically timed steps
> * since we would have scenarios, one test could cover much more functionality
> * the individual tests could execute much faster
> * the GRPC mock is written in Golang, so integrating it with the existing testing logic is hopefully "reinventing the wheel".
> 
> What do you think @JoukoVirtanen ?

I agree with incorporating the mock grpc server directly into the integration tests, but I think that could be done in another PR.

PR #1196, opened 2023-06-14
Removes mapped_id, to fix platforms that require sys_enter and sys_exit

Stringy:
## Description

When using sys_enter and sys_exit, `mapped_id` remained set to the placeholder (-1) which was then used to determine high-level syscall filtering. This results in every event being dropped.

This PR removes mapped_id, in favor of populating or using the existing `id`.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)

### DIFF
--- a/kernel-modules/probe/collector_probe.c
+++ b/kernel-modules/probe/collector_probe.c
@@ -224,7 +224,6 @@ static __always_inline int enter_probe(long id, struct sys_enter_args* ctx) {
   enum ppm_event_type evt_type = PPME_GENERIC_E;
   int drop_flags = UF_ALWAYS_DROP;
   struct sys_enter_args stack_ctx = {.id = id};
-  long mapped_id = id;
 
   if (bpf_in_ia32_syscall()) {
     return 0;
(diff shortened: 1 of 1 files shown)


kernel-modules/probe/collector_probe.c
```diff
@@ -316,7 +314,7 @@ static __always_inline int exit_probe(long id, struct sys_exit_args* ctx) {
   }
```
ovalenti, line 314: There is a `mapped_id` reference remaining.

ovalenti (changes requested):

kernel-modules/probe/collector_probe.c
Stringy: dang, good catch! thanks

Stringy:
cc @kcrane - these changes may affect s390x, could you have a look?

ovalenti (approved):
lgtm !

PR #1199, opened 2023-06-15
ROX-15297: centralizes test configuration

Stringy:
## Description

Environment variables and configuration options now managed centrally rather than across the testing files.

Includes an image store, by way of a simple yaml file, to centralize the declaration of images between tests and avoid reuse.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Tested locally, otherwise depend on CI to exercise these changes.

### DIFF
--- a/integration-tests/suites/config/images.go
+++ b/integration-tests/suites/config/images.go
@@ -15,6 +15,8 @@ func (i *ImageStore) CollectorImage() string {
 	return ReadEnvVar(envCollectorImage)
 }
 
+// ImageByKey looks up an image from the store, and panics
+// if the image does not exist.
 func (i *ImageStore) ImageByKey(key string) string {
 	if img, ok := i.Images[key]; ok {
 		return img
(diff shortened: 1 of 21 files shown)


integration-tests/suites/config/images.go
```diff
@@ -0,0 +1,70 @@
...
+
+type ImageStore struct {
+	Images map[string]string
+}
+
+func (i *ImageStore) CollectorImage() string {
+	return ReadEnvVar(envCollectorImage)
+}
+
+func (i *ImageStore) GRPCImage() string {
```
ovalenti, line 18: I think I understand that the GRPC mock is so much used everywhere that it deserves its own method. My first guts feeling was that every image could have a dedicated method; the advantage being that the compiler can check for typos in method names whereas it cannot with keys (and also a source browser can give you all usage locations).

But after more thinking, a string-search is rather effective, the tests will spot typos and you might have found just the right balance.

ovalenti (approved):
This is great, thanks !

integration-tests/suites/config/images.go
Stringy: That is a very good point, it's just as static as the other images (I think Collector is the only one that's dynamic?) I'll remove this method and just get the image in the same way as all the others

PR #1206, opened 2023-06-19
Fixes local default in integration test config

Stringy:
## Description

Correctly sets the default REMOTE_HOST_TYPE. Without this change, the tests will segfault if a REMOTE_HOST_TYPE is not provided.


ovalenti (approved):
That should fix it :D

PR #1207, opened 2023-06-20
Ansible playbook for local development setup.

Stringy:
## Description

I've found myself doing 'local' development on various GCP VMs, particularly to test driver builds. This playbook will install basic dev dependencies and clone collector into the user's home directory, checking out a different branch if COLLECTOR_BRANCH environment variable is set


ovalenti (approved):
As far as I understand it, this should be very handy :D

PR #1209, opened 2023-06-20
Refactors GHA to reduce wait times for testing

Stringy:
## Description

Makes the ppc64le build and multi-arch manifest step into separate workflows, to allow tests to run while the slow ppc64le build runs. We currently only test x86 upstream, and only build amd64 full images upstream. 

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,9 +58,7 @@ jobs:
 
   build-collector-slim-ppc64le:
     uses: ./.github/workflows/collector-slim.yml
-    needs:
-      - init
-      - build-collector-slim-amd64
+    needs: init
     with:
       collector-tag: ${{ needs.init.outputs.collector-tag }}
       collector-image: ${{ needs.init.outputs.collector-image }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,7 +58,9 @@ jobs:
 
   build-collector-slim-ppc64le:
     uses: ./.github/workflows/collector-slim.yml
-    needs: init
+    needs:
+      - init
+      - build-collector-slim-amd64
     with:
       collector-tag: ${{ needs.init.outputs.collector-tag }}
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -47,12 +47,24 @@ jobs:
...
       collector-tag: ${{ needs.init.outputs.collector-tag }}
       collector-image: ${{ needs.init.outputs.collector-image }}
+      arch: amd64
+    secrets: inherit
+
+  build-collector-slim-ppc64le:
+    uses: ./.github/workflows/collector-slim.yml
+    needs:
+      - init
+      - build-collector-slim-amd64
```
ovalenti, line 63: Why is slim-ppc64le depending on slim-amd64 ?

Stringy: It was to force them to run sequentially rather than in parallel, mostly as a way of ensuring that amd64 runs first. There is a possibility that there aren't enough workers to run both at the same time, which could mean that ppc64le runs first, which is back to being extremely slow again (I did see this in testing.) 

It's probably not strictly necessary though

ovalenti: Understood, though I never saw such a case happen in GHA.

I believe that we could leave those run in parallel and keep the graph clean.

ovalenti (approved):
The matrix build arch was appealing, but there was something wrong with the fact that archs had to be enumerated for the multiarch manifest, for instance.

Overall, things are better now, thanks !

ovalenti:
The [integration test tag](https://github.com/stackrox/collector/blob/e73e74db59ea7732bd50447e86988f2b672d550b/.github/workflows/main.yml#L84) in main.yml may need to include the `-amd64` suffix as well

```
collector-tag: ${{ needs.init.outputs.collector-tag }}-amd64
```

PR #1210, opened 2023-06-20
Fixes selinux disabling on coreos and rhcos

Stringy:
## Description

Fixes check to disable selinux on fedora-coreos and rhcos


ovalenti (approved):
Good catch :+1:

PR #1212, opened 2023-06-25
Jv log events that are too old

JoukoVirtanen:
## Description

Adds logging for when the event time doesn't agree with collector time. This logging includes events that are too old or are in the future. Also adds metrics for these cases.

## Checklist
- [x] Inspect CI logs for unreasonable timestamps
      No unreasonable timestamps were found
- [x] Ran integration test locally and saw the new logging
- [x] Ran integration test locally and saw the new counters in the metrics

### Logging

Ran integration tests locally. Saw the following in the logs

[INFO    2023/08/10 21:55:09] (SignalServiceClient.cpp:20) Trying to establish GRPC stream for signals ...
[INFO    2023/08/10 21:55:09] (SignalServiceClient.cpp:38) Successfully established GRPC stream for signals.
[WARNING 2023/08/10 21:55:09] [Throttled] (SysdigService.cpp:214) Event of type 293 is unreasonably old. It's timestamp is 2023-08-10T14:55:10.022197Z
[INFO    2023/08/10 21:55:09] (SelfCheckHandler.cpp:18) Found self-check process event.
[INFO    2023/08/10 21:55:09] (SelfCheckHandler.cpp:43) Found self-check connection event.

### Metrics

Modified integration test so that collector would not terminate

        //err = s.collector.TearDown()
        //s.Require().NoError(err)

Ran integration test with `make`

Then got metrics

 curl localhost:9090/metrics
 
 Saw the following
 
\# HELP rox_collector_counters Collector counters
\# TYPE rox_collector_counters gauge
rox_collector_counters{type="event_timestamp_distant_past"} 1620
rox_collector_counters{type="procfs_could_not_open_proc_dir"} 0
rox_collector_counters{type="event_timestamp_future"} 0
rox_collector_counters{type="procfs_could_not_open_fd_dir"} 0
rox_collector_counters{type="procfs_could_not_read_cmdline"} 0
rox_collector_counters{type="process_info_miss"} 0
rox_collector_counters{type="net_conn_updates"} 3
rox_collector_counters{type="process_lineage_string_total"} 2323
rox_collector_counters{type="procfs_could_not_read_exe"} 0
rox_collector_counters{type="net_conn_deltas"} 2
rox_collector_counters{type="net_conn_inactive"} 2
rox_collector_counters{type="rate_limit_flushing_counts"} 0
rox_collector_counters{type="net_cep_updates"} 0
rox_collector_counters{type="procfs_could_not_get_socket_inodes"} 0
rox_collector_counters{type="net_cep_deltas"} 0
rox_collector_counters{type="net_known_ip_networks"} 0
rox_collector_counters{type="procfs_could_not_get_network_namespace"} 0
rox_collector_counters{type="procfs_could_not_open_pid_dir"} 0
rox_collector_counters{type="process_lineage_sqr_total"} 200
rox_collector_counters{type="net_cep_inactive"} 0
rox_collector_counters{type="process_lineage_total"} 192
rox_collector_counters{type="net_known_public_ips"} 0
rox_collector_counters{type="process_lineage_counts"} 200
rox_collector_counters{type="process_info_hit"} 0

## Future work

Will look into creating integration tests that check the metrics created here.


collector/lib/SysdigService.cpp
```diff
@@ -223,6 +226,24 @@ void SysdigService::Start() {
...
+void LogUnreasonableEventTime(int64_t time_micros, sinsp_evt* evt) {
+  int64_t time_diff;
+  int64_t evt_ts = evt->get_ts() / 1000UL;
+  int64_t max_past_time = 3600000000;  // One hour in microseconds. This is generous.
+  int64_t max_future_time = 5000000;   // 5 seconds in microseconds. A time in a little in the future is suspicious.
+
+  time_diff = time_micros - evt_ts;
+  if (time_diff > max_past_time) {
+    CLOG_THROTTLED(WARNING, std::chrono::seconds(10)) << "Event of type " << evt->get_type() << " is unreasonably old. It's timestamp is " << google::protobuf::util::TimeUtil::MicrosecondsToTimestamp(evt_ts);
+    COLLECTOR_ASSERT(false);
```
JoukoVirtanen, line 238: Perhaps the logging statement and assert could be combined. See above. Perhaps something other than false could be passed in, like `(time_diff > max_past_time)`.

Molter73 (commented):
I feel like implementing our own assert is an unnecessary complexity, asserts are meant to stop code when something goes wrong in a debugging/dev cycle, so there would be no point IMO in compiling collector with release flags but still have this other type of assert be able to stop execution.

If the problem is that the regular asserts sometimes causes collector to stop when compiled with debug flags, we should investigate and fix these (even if the assert comes from falco and means we need to make a fix upstream), as opposed to overengineering our code.

collector/lib/CollectorAssert.h
```diff
@@ -0,0 +1,38 @@
...
+*/
+
+#pragma once
+
+#ifdef COLLECTOR_ASSERT
+
+#  define COLLECTOR_ASSERT(X)                                                                                                                                  \
+    do {                                                                                                                                                       \
+      if (!(X)) {                                                                                                                                              \
+        collector::logging::LogMessage(__FILE__, __LINE__, false, collector::logging::LogLevel::FATAL) << "Assertion failed: " << #X << " evaluated to false"; \
```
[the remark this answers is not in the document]
JoukoVirtanen: I have removed the assert.

collector/lib/SysdigService.cpp
JoukoVirtanen: I have removed the assert.

collector/lib/SysdigService.cpp
```diff
@@ -212,6 +233,7 @@ void SysdigService::Run(const std::atomic<ControlValue>& control) {
     if (!evt) continue;
 
     auto process_start = NowMicros();
+    LogUnreasonableEventTime(process_start, evt);
```
JoukoVirtanen, line 236: Move to after `if (!signal_handler.ShouldHandle(evt)) continue;`

ovalenti (approved):
This is in the hot path, but the computation cost is very minimal.

Please, move the `LogUnreasonableEventTime call to after the Handlers filtering as you suggested, then it's good to merge !

JoukoVirtanen:
> This is in the hot path, but the computation cost is very minimal.
> 
> Please, move the `LogUnreasonableEventTime call to after the Handlers filtering as you suggested, then it's good to merge !

Done

PR #1228, opened 2023-07-05
Allow collector to fallback to regular eBPF by default

### LINKED ISSUE #1224: [ROX-18272] Reinstate EBPF fallbacks
Change "hard failure" mode to more user-friendly "soft failure" in the following cases:

- [ ] Heuristic hard fail
- [ ] Fallback to the next available collection method during the probe load

See [ROX-18272](https://issues.redhat.com/browse/ROX-18272) for more high-level details.

Molter73:
## Description

Change the default collector configuration so that it falls back to eBPF when the CO-RE BPF driver fails to be inserted.

Fixes #1224

## Checklist

## Testing Performed

CI should be enough since the change is pretty straightforward.

### DIFF
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -55,7 +55,7 @@ BoolEnvVar set_enable_core_dump("ENABLE_CORE_DUMP", false);
 // If true, add originator process information in NetworkEndpoint
 BoolEnvVar set_processes_listening_on_ports("ROX_PROCESSES_LISTENING_ON_PORT", CollectorConfig::kEnableProcessesListeningOnPorts);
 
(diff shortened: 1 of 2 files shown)


ovalenti (commented):
This seems quite safe, indeed :)

I think that we should set `ROX_COLLECTOR_CORE_BPF_HARDFAIL=1` for the integration tests. Otherwise, it will end-up masking failures with core_bpf. What do you think ?

Molter73:
> I think that we should set `ROX_COLLECTOR_CORE_BPF_HARDFAIL=1` for the integration tests. Otherwise, it will end-up masking failures with core_bpf. What do you think ?

Makes sense, let me find the easiest way to implement it.

ovalenti (approved):
It looks good to me !

PR #1242, opened 2023-07-13
Print checksum of drivers being used

Molter73:
## Description

Have collector print the sha256 checksum for the driver it uses. Same as #1237, this is part of the groundwork for unifying the upstream and downstream built drivers. By printing the checksum for a driver being used it'll be easier for us to check where the driver comes from in case a user reports an issue.

A small script for downloading drivers from GCP and validating checksums is also provided, in hopes that if we ever need to validate the source of a driver it will make the work somewhat easier.

## Checklist

**Automated testing**
  - [x] Added unit tests

## Testing Performed

- [x] Manually checked `driver-checksum.sh` properly validates files and cleans up after itself.

### DIFF
--- a/utilities/driver-checksum.sh
+++ b/utilities/driver-checksum.sh
@@ -1,6 +1,21 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+# A small script for quickly validating driver SHA256 checksums.
+# See the 'usage' function or run './driver-checksum.sh -h' for a
+# comprehensive explanation on how to use the script.
+
+set -eo pipefail
+
+function usage() {
+    printf "usage: driver-checksum.sh source checksum driver\n"
+    printf "\n"
+    printf "    source\tSpecify if the drivers exists as a file or on gcp\n"
+    printf "          \t  Values: gcp, file\n"
+    printf "    checksum\tA SHA256 checksum to be compared against the driver's\n"
+    printf "    driver\tThe path to the driver on GCP or the local fs\n"
+    printf "\n"
+    printf "    -h, --help\tThis help message\n"
+}
 
 function process_gcp_driver() {
     checksum="${1}"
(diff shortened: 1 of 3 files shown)


utilities/driver-checksum.sh
```diff
@@ -0,0 +1,56 @@
...
+    if ! sha256sum -c "${temp_file}"; then
+        echo >&2 "Checksum validation failed"
+        retval=1
+    fi
+
+    rm -f "${temp_file}"
+    return $retval
+}
+
+SOURCE=${1}
```
JoukoVirtanen, line 40:
```suggestion
SOURCE=${1} # Valid options are "gcp" or "file"
```

utilities/driver-checksum.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
```
JoukoVirtanen, line 4: Add a comment about what this script does and include a couple of examples.

JoukoVirtanen (approved):
LGTM! You could a couple of comments to the script file to explain what it does.

utilities/driver-checksum.sh
JoukoVirtanen, line 4:
```suggestion

# Takes in the source type of the driver, either gcp or file, a sha256 checksum, and the location of the file
# and checks if the sha256 checksum of the file actually matches the provided checksum.
# E.g ./driver-checksum.sh file 3ea61b22b3123871b203cd352c60884e336d63777145bfdad20388aa74b71a06 ../kernel-modules/container/kernel-modules/collector-ebpf-4.18.0-477.13.1.el8_8.x86_64.o.g
```

JoukoVirtanen: Perhaps add a comment explaining what the script is for and how to use it.

utilities/driver-checksum.sh
JoukoVirtanen: I think SOURCE_TYPE is clearer than SOURCE. This will need to be changed elsewhere as well.

Molter73: I tend to think of "source" as "the place the driver we are testing comes from", so IMO source is actually more accurate than source_type which would imply there is a source somewhere else in the code, WDYT?

utilities/driver-checksum.sh
Molter73: I added a small comment on the purpose of the script and a `usage` function describing what each argument should be, hopefully that is enough 🙂

JoukoVirtanen: Good idea.

PR #1243, opened 2023-07-13
ROX-18736 Report connections only when they are established

ovalenti:
## Description

The Collector should report connections that can actually be used to transfer information.
In the case of a TCP connection in asynchronous mode, we have to postpone declaring the connection and
wait until we know the state of the socket.

- When connect returns `INPROGRESS`, we do not declare a connection anymore.
- If the socket is pending and a getsockopt(SO_ERROR) returns a success, we declare this connection

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [x] Added integration tests
  - [x] Added regression tests

### DIFF
--- a/collector/lib/NetworkSignalHandler.cpp
+++ b/collector/lib/NetworkSignalHandler.cpp
@@ -26,17 +26,36 @@ EventMap<Modifier> modifiers = {
 };
 
 }  // namespace
-
+/*
+ * Socket connection life-cycle scenarii:
+ * - synchronous:
+ *   - events:  connect()/accept() >= 0 --> close()/shutdown() = 0
+ *     fd_info: connected                   connected
+ *     result:  ADD                         REMOVE
+ *   - events:  connect()/accept() < 0
+ *     fd_info: failed
+ *     result:  nil
+ *   - events:  accept() >= 0
+ * - asynchronous:
+ *   - events:  connect() < 0 (E_INPROGRESS) --> getsockopt() ok --> shared with synchronous
+ *     fd_info: pending                          connected
+ *     result:  nil                              ADD
+ *   - events:  connect() < 0 (other)
+ *     fd_info: failed
+ *     result:  nil
+ */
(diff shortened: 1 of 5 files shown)


ovalenti:
Reminder :  add getsockopt to the captured syscalls of custom-probe

collector/lib/NetworkSignalHandler.cpp
```diff
@@ -20,22 +20,32 @@ EventMap<Modifier> modifiers = {
...
+  if (!fd_info) return std::nullopt;
+
+  if (fd_info->is_socket_failed()) {
+    return std::nullopt;
+  }
+
+  if (fd_info->is_socket_pending()) {
+    CLOG(DEBUG) << "Ignoring pending connection until we determine its status.";
+    return std::nullopt;
+  }
```
Stringy, line 41: so my understanding of this is that when the socket is pending, the result that we get on line 43 will be some errno value which is why we're first checking `fd_info`? Is this correct?

Regardless, I think it'd be helpful to have a little comment description of the socket lifecycle and expected results/behaviours

Stringy (commented):
Generally looks great, though I think we need an integration test for this behaviour. Not sure how easy this would be to setup, however.

collector/lib/NetworkSignalHandler.cpp
ovalenti: > so my understanding of this is that when the socket is pending, the result that we get on line 43 will be some errno value which is why we're first checking `fd_info`? Is this correct?

This is correct, we now handle a case where a negative result has more meaning that just "fail". That is connect() returning with errno=E_INPROGRESS

> Regardless, I think it'd be helpful to have a little comment description of the socket lifecycle and expected results/behaviours

Indeed, this will be appreciated when I come back to this in 6 months :sweat_smile: 
I added some charts and comments, if you want to take a look -> https://github.com/stackrox/collector/commit/1a493e7a5214793b6e2e84c17bc2259720a8d347

PR #1244, opened 2023-07-14
Cache collector install steps

Molter73:
## Description

When building the collector image, a final-step.sh is used to install some packages depending on the base image being used. Because this script is called after the required binaries and scripts are copied, modifying collector invalidates the layer cache for this step, making the build process take a bit longer.

This commit moves some commands shared by the regular and dev images into the dockerfile itself, renames the scripts to install.sh and calls it before copying files that might be subject to modification. This should slightly speed up dev iteration, as the collector image can cache the installation step (usually the step that takes the longest).

## Checklist

## Testing Performed

CI should be enough.


Stringy (approved):
LGTM!

PR #1245, opened 2023-07-17
Added s390x to multiarch collector slim build WIP

jdao-rh:
Added s390x arch to multiarch build of collector-slim


jdao-rh:
duplicate of https://github.com/stackrox/collector/pull/1261

PR #1246, opened 2023-07-18
Use non-docker actionlint commit hook

Molter73:
## Description

The actionlint-docker commit hook has some issues when running using podman from macos, it tries to mount the repository in a way that is not supported and fails. Because fixing this upstream would be annoying and installing actionlint locally shouldn't be a big deal, we should start using the native binary option. This is also preferable to committing with --no-verify and relaying on people running the check manually.

For instructions on how to install actionlint see: https://github.com/rhysd/actionlint/blob/main/docs/install.md

## Checklist

## Testing Performed

This is a change that only affects developer workflow.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,8 +12,12 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-python@v3
       - uses: mfinelli/setup-shfmt@v2
+
+      - name: Install actionlint
+        run: |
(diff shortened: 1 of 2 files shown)


Stringy:
I think we'll need to install `actionlint` in the GHA runner for the lint workflow (unless it is there by default?)

Molter73:
> I think we'll need to install `actionlint` in the GHA runner for the lint workflow (unless it is there by default?)

You are right, it's not in the list of installed software, I'll run some tests and get back to you.

Stringy (approved):

PR #1248, opened 2023-07-19
ROX-18548: Use downstream built drivers for image and support-packages

Molter73:
## Description

With this change we'll start packing downstream build drivers for collector and in the support-packages for driver versions 2.6.0+. This provides a clean break even if the drivers themselves will not have any changes.

## Checklist

## Testing Performed
- [x] Check the support-package attempts to use the downstream built drivers. ([CI run](https://github.com/stackrox/collector/actions/runs/5622712929/job/15239395160?pr=1248))
- [x] Check the full images have the downstream built drivers. ([CI run](https://github.com/stackrox/collector/actions/runs/5643937916/job/15288258722))


.github/workflows/init.yml
```diff
@@ -125,3 +131,40 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs:
+    - common-variables
+    if: github.event_name == 'push' && (github.ref_type == 'tag' || startsWith(github.ref_name, 'release-'))
+    steps:
+      - uses: actions/checkout@v3
+      - name: Ensure downstream drivers exist for releases
+        run: |
+          DRIVER_VERSION="$(cat ${{ github.workspace }}/kernel-modules/MODULE_VERSION)"
+          DRIVER_COUNT="$(gsutil ls "gs://${{ needs.common-variables.outputs.cpaas-drivers-bucket }}/${DRIVER_VERSION}/" | wc -l)"
```
JoukoVirtanen, line 145: There should probably be some error handling in case the bucket doesn't exist. Also I am not sure that this won't return an error if there are no files in the bucket. `grep asdf some_file.txt | wc -l` returns an error if asdf doesn't appear in the file. Might this have the same issue?

.github/workflows/main.yml
```diff
@@ -94,11 +97,14 @@ jobs:
...
       branch-name: ${{ needs.init.outputs.branch-name }}
-      drivers-bucket: ${{ needs.init.outputs.drivers-bucket }}
+      upstream-drivers-bucket: ${{ needs.init.outputs.drivers-bucket }}
+      downstream-drivers-bucket: ${{ needs.init.outputs.cpaas-drivers-bucket }}/x86_64
       support-packages-bucket: ${{ needs.init.outputs.support-packages-bucket }}
       support-packages-index-bucket: ${{ needs.init.outputs.support-packages-index-bucket }}
       public-support-packages-bucket: ${{ needs.init.outputs.public-support-packages-bucket }}
-    if: always() && (github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'test-support-packages'))
+    if: |
+      (success() || failure()) &&
```
JoukoVirtanen, line 106: Perhaps this needs a comment. It looks like this might always evaluate to true.

kernel-modules/support-packages/04-create-support-packages.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -exuo pipefail
```
JoukoVirtanen, line 3: Remember to change back.

kernel-modules/support-packages/04-create-support-packages.sh
```diff
@@ -22,10 +22,24 @@ compress_files() (
     zip -r "${output_file}" .
 )
 
+use_downstream() {
+    IFS='.' read -ra version <<< "${1%-*}"
```
JoukoVirtanen, line 26: There should be some input validation. Are we getting three integers?

JoukoVirtanen:
Was this error fixed https://github.com/stackrox/collector/actions/runs/5669610864/job/15362790259

.github/workflows/init.yml
Molter73: I tested this, if the bucket doesn't exist an error message is printed to stderr and no lines are counted in stdout, so the check for 0 is still valid.

.github/workflows/main.yml
Molter73: I'll add one, the reason for this change is that cancelling a workflow still ran the support-package step, which shouldn't be the case, this new condition skips the step when the workflow is cancelled but still runs it if everything is fine or a previous step failed.

kernel-modules/support-packages/04-create-support-packages.sh
Molter73: Why wouldn't we get 3 integers? The input this function gets is the driver version that has been changed to a `M.m.p` format for quite some time now.

Molter73:
> Is there a way to view the kernel versions added, replaced, and unchanged in the support package? If this is only available in logs should we add a metadata file to the gcp bucket?

I'll add something to the support-package step.

.github/actions/support-package-driver-matrix/driver-matrix.py
```diff
@@ -0,0 +1,149 @@
...
+
+def main(kernels: dict, file, available: str):
+    for line in sys.stdin:
+        line = line.rstrip()
+        process_line(kernels, available, line)
+
+    if file:
+        json.dump(kernels, file)
+    else:
+        print(json.dumps(kernels))
```
Stringy, line 128: we can use json.dump in both branches here - 

```
if file:
    json.dump(kernel, file)
else:
    json.dump(kernel, sys.stdout)
```

Or better still, we can pass in either file or stdout in the caller and then we don't need to care where we're dumping

.github/actions/support-package-driver-matrix/driver-matrix.py
```diff
@@ -0,0 +1,149 @@
...
+
+    file = None
+    if args.update:
+        if os.path.isfile(args.update):
+            with open(args.update, 'r') as f:
+                kernels = json.load(f)
+
+        file = open(args.update, "w")
+
+    main(kernels, file, available)
```
Stringy, line 149: [nit] it feels a bit redundant that we're passing both the kernel list and the file to `main()` by reference. Since we have access to the kernels dictionary at this level, and main doesn't need to care about the file, can't we just do the json dumping here? Or if not, it might make sense to pass all the responsibility of loading and updating files to the main function

.github/workflows/init.yml
```diff
@@ -125,3 +131,40 @@ jobs:
...
             } >> "$GITHUB_OUTPUT"
           fi
+
+  downstream-check:
+    runs-on: ubuntu-latest
+    needs:
+    - common-variables
+    if: github.event_name == 'push' && (github.ref_type == 'tag' || startsWith(github.ref_name, 'release-'))
+    steps:
+      - uses: actions/checkout@v3
```
Stringy, line 141: [ultranit] can we have a whitespace gap between these steps

.github/workflows/support-packages.yml
```diff
@@ -46,13 +50,23 @@ jobs:
...
           branch-name: ${{ inputs.branch-name }}
 
+      - name: Create driver-matrix.json
+        uses: ./.github/actions/support-package-driver-matrix
+        with:
+          metadata-dir: /tmp/support-packages/metadata/module-versions
+          upstream-drivers-bucket: gs://${{ inputs.upstream-drivers-bucket }}
+          downstream-drivers-bucket: gs://${{ inputs.downstream-drivers-bucket }}
+
+      - run: jq . /tmp/driver-matrix.json
```
Stringy, line 60: is this for debugging/logging purposes?

.github/actions/support-package-driver-matrix/action.yml
```diff
@@ -0,0 +1,38 @@
...
+          mod_ver="$(basename "$mod_ver_dir")"
+
+          # Add the regular drivers
+          gsutil ls "${{ inputs.upstream-drivers-bucket }}/${mod_ver}/*" \
+              | ${{ github.action_path }}/driver-matrix.py -u /tmp/driver-matrix.json
+
+          if use_downstream "$mod_ver"; then
+            # Update the matrix with downstream drivers
+            gsutil ls "${{ inputs.downstream-drivers-bucket }}/${mod_ver}/*" \
+                | ${{ github.action_path }}/driver-matrix.py -u /tmp/driver-matrix.json -d
```
Stringy, line 36: [nit] can we use the long-form arg names to make it a bit clearer (same on line 31)

.github/actions/support-package-driver-matrix/driver-matrix.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-u', '--update', help='File to be updated')
+    parser.add_argument('-d', '--downstream', action='store_true',
+                        help='Mark available drivers as "downstream"')
+    args = parser.parse_args()
+
+    available = 'downstream' if args.downstream else 'upstream'
```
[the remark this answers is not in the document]
Molter73: Yeah, you can tell I put this script together in like 15 minutes, I'll change the name to something more meaningful.

.github/actions/support-package-driver-matrix/driver-matrix.py
```diff
@@ -0,0 +1,149 @@
...
+
+import argparse
+import json
+import os
+import sys
+import re
+
+ebpf_re = re.compile(r'(\d+\.\d+\.\d+)/collector-ebpf-(\d+\.\d+\.\d+.*)\.o\.gz$')
+kmod_re = re.compile(r'(\d+\.\d+\.\d+)/collector-(\d+\.\d+\.\d+.*)\.ko\.gz$')
+unavailable_re = re.compile(r'(\d+\.\d+\.\d+)/\.collector-ebpf-(\d+\.\d+\.\d+.*)\.unavail$')
```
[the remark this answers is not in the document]
Molter73: I considered doing some regex magic in order to pick up if the driver is a kmod or ebpf, then I remembered regex magic is one of the primary sources of evil in the universe. I'll try to de-duplicate code a bit but no promises.

.github/actions/support-package-driver-matrix/driver-matrix.py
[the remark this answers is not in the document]
Molter73: I'll rename this as `source` as in "the source for this driver is either upstream, downstream or unavailable". Will also unify it across all function.s

kernel-modules/support-packages/04-create-support-packages.sh
JoukoVirtanen: You could do something like

    if [[ ! $1 =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
        echo "Error: Invalid version number format. Expected format: major.minor.patch"
        return 1
    fi
    
  Let's not make assumptions about what the format might be like in the future.

.github/actions/support-package-driver-matrix/driver-matrix.py
```diff
@@ -0,0 +1,141 @@
...
+            }
+        else:
+            dver[driver_type] = source
+
+
+def process_line(kernels: dict,
+                 source: str,
+                 line: str):
+    (driver_path, driver) = os.path.split(line)
+    (_, driver_version) = os.path.split(driver_path)
```
Stringy, line 93: [nit] redundant parenthesis

```suggestion
    driver_path, driver = os.path.split(line)
    _, driver_version = os.path.split(driver_path)
```

Stringy (approved):
LGTM!

PR #1249, opened 2023-07-19
Add IBM Z VM support to collector ansible

levivic:
Using IBM Cloud to manage IBM Z virtual instance for Collector upstream CI. @Molter73

### DIFF
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -54,6 +54,7 @@ summarized below:
 | Type          | Families       |
 | ------------- | -------------- |
 | rhel          | rhel-7 <br> rhel-8 |
+| rhel-s390x    | rhel-8-6-s390x | 
 | rhel-sap      | rhel-8-4-sap-ha <br> rhel-8-6-sap-ha |
 | cos           | cos-stable <br> cos-beta <br> cos-dev |
 | sles          | sles-12 <br> sles-15 |
@@ -102,6 +103,9 @@ The following environment variables may be used to modify some behavior:
 | JOB_ID | A unique identifier to de-conflict VM names | the current user's username |
 | COLLECTOR_TEST | Which integration test make target to run. (e.g. integration-test-process-network) | ci-integration-tests |
 | VM_TYPE | Which kind of VMs to create on GCP (as listed above.) By default, will only build rhel VMs. Use 'all' to build an inventory containing every kind of VM. | rhel |
+| IC_API_KEY | API key for IBM Cloud. See [doc](https://cloud.ibm.com/docs/account?topic=account-userapikey&interface=ui) |
+| IC_REGION | The IBM Cloud region where you want to create your resources. See [doc](https://cloud.ibm.com/docs/workload-protection?topic=workload-protection-regions)|
+|||
 
 Note: other environment variables that may affect the operation of the integration tests
 can be used to modify behavior. See [the integration tests README](../integration-tests/README.md) for details.
@@ -113,7 +117,7 @@ root of this directory structure. It should contain key/value pairs of variable
 names and credentials to be used in the playbooks. Currently, the only required
 credentials are quay_username and quay_password, which are created by make
 from the environment variables `QUAY_RHACS_ENG_RO_USERNAME` and `QUAY_RHACS_ENG_RO_PASSWORD`
-to match CI variables.
+to match CI variables. If you are using IBM Cloud to create IBM Z RHEL instances, you will also need to specifiy enviroment variables `RHEL_SUBSCRIPTION_USERNAME` and `RHEL_SUBSCRIPTION_PASSWORD` to register your RHEL system.
 
 To create your own, for dev the format should be:
 
--- a/ansible/dev/group_vars/all.yml
+++ b/ansible/dev/group_vars/all.yml
@@ -1,5 +1,6 @@
 ---
 
+# GCP setup
 gcp_project: "stackrox-dev"
 gcp_auth_kind: "application"
 gcp_service_account_file: null
@@ -14,3 +15,14 @@ ansible_user: "{{ lookup('env', 'USER') }}"
 gcp_default_labels:
   stackrox-ci: "false"
 
+
+# IBM Cloud for IBM Z setup
+# vpc_name: lei-vpc-toronto
+vpc_id: r038-c4f8e233-4b56-44bd-af71-140ebd1664e9
+# subnet_name: stackrox-ci-sn2
+subnet_id: 02r7-68814545-4e23-4fe9-b557-a557f03a2b71
+zone: ca-tor-2
+# ssh_key_name: acs-sshkey
+ssh_key_id: r038-fb0260c7-c01d-45c8-8026-7d50042943b9
+# vsi_resource_group: stackrox-ci-resource-group
+vsi_resource_group_id: 1a33a6a9bd6e498f8115e9b1064bfa97 
--- a/ansible/roles/create-s390x-vm/tasks/main.yml
+++ b/ansible/roles/create-s390x-vm/tasks/main.yml
@@ -1,0 +1,74 @@
+- name: Fetch the variables from var file
+  include_vars:
+    file: vars.yml
+
+- name: Check for existing VSI
+  ibm.cloudcollection.ibm_is_instance_info:
+    name: "{{ vsi_name }}"
+  failed_when:
+    - vsi.rc != 0
+    - '"No Instance found" not in vsi.stderr'
+  register: vsi
+
+- name: Configure VSI
+  ibm.cloudcollection.ibm_is_instance:
+    name: "{{ vsi_name }}"
+    state: available
+    id: "{{ vsi.resource.id | default(omit) }}"
+    vpc: "{{ vpc_id }}"
+    profile: "{{ vsi_profile }}"
+    image: "{{ vsi_image_id }}"
+    resource_group: "{{ vsi_resource_group_id }}"
+    keys:
+      - "{{ ssh_key_id }}"
+    primary_network_interface:
+     - subnet: "{{ subnet_id }}"
+    zone: "{{ zone }}"
+  register: vsi
+
+- name: Check for existing Floating IP
+  ibm.cloudcollection.ibm_is_floating_ip_info:
+    name: "{{ vsi_name }}-fip"
+  failed_when:
+    - fip.rc != 0
+    - '"No floatingIP found" not in fip.stderr'
+  register: fip
+
+- name: Configure Floating IP Address
+  ibm.cloudcollection.ibm_is_floating_ip:
+    name: "{{ vsi_name }}-fip"
+    state: available
+    id: "{{ fip.resource.id | default(omit) }}"
+    resource_group: "{{ vsi_resource_group_id }}"
+    target: "{{ vsi.resource.primary_network_interface[0]['id'] }}"
+  register: fip
+
+- name: Print Floating IP Address
+  debug:
+    msg: "{{ vsi_name }} IP Address: {{ fip.resource.address }}"
+
+- name: Add VSI to Ansible inventory
+  add_host:
+    name: "{{ fip.resource.address }}"
+    ansible_user: root
+    groups:
+      - job_id_{{ job_id }}
+      - platform_{{ test_platform }}
+    ansible_ssh_extra_args: "-o StrictHostKeyChecking=no"
+    ansible_ssh_private_key_file: "~/.ssh/acs-sshkey_rsa.prv"
+    vm_config: "{{ vm_config }}"
+    vm_image_family: "{{ vm_image_family }}"
+    vm_collection_method: "{{ vm_collection_method }}"
+
+- name: Register server with Red Hat
+  tags: attach_subscription
+  community.general.redhat_subscription:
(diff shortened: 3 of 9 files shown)


Molter73:
cc @Stringy for his ansible expertise.

ansible/README.md
```diff
@@ -113,7 +117,7 @@ root of this directory structure. It should contain key/value pairs of variable
 names and credentials to be used in the playbooks. Currently, the only required
 credentials are quay_username and quay_password, which are created by make
 from the environment variables `QUAY_RHACS_ENG_RO_USERNAME` and `QUAY_RHACS_ENG_RO_PASSWORD`
-to match CI variables.
+to match CI variables. If you are using IBM Cloud to create IBM Z RHEL instances, you will also need to specifiy enviroment variables `RHEL_SUBSCRIPTION_USERNAME` and `RHEL_SUBSCRIPTION_PASSWORD` to register your RHEL system.
```
Molter73, line 120: I believe we can use `REDHAT_USERNAME` and `REDHAT_PASSWORD`, our CI has those setup to a RHSM account for the team.
```suggestion
to match CI variables. If you are using IBM Cloud to create IBM Z RHEL instances, you will also need to specifiy enviroment variables `REDHAT_USERNAME` and `REDHAT_PASSWORD` to register your RHEL system.
```

ansible/roles/create-s390x-vm/vars.yml
```diff
@@ -0,0 +1,8 @@
+---
+#rhel-8-6-s390x
+vsi_image_id: r038-0d89bf54-c28f-4dc4-bec0-0f8e03a74de3
+vsi_profile: bz2-2x8
+vsi_name: "{{ vm_name }}"
+
+redhat_username: "{{ lookup('env', 'RHEL_SUBSCRIPTION_USERNAME', default='') }}"
+redhat_password: "{{ lookup('env', 'RHEL_SUBSCRIPTION_PASSWORD', default='') }}"
```
Molter73, line 8:
```suggestion
redhat_username: "{{ lookup('env', 'REDHAT_USERNAME', default='') }}"
redhat_password: "{{ lookup('env', 'REDHAT_PASSWORD', default='') }}"
```

ansible/roles/create-vm/tasks/main.yml
```diff
@@ -32,7 +32,22 @@
       vm_image_family: "{{ vm_family }}"
       vm_collection_method: "{{ vm_collection_method }}"
 
-- name: Create VM
+- name: Determin the arch
```
Molter73, line 35: Typo
```suggestion
- name: Determine the arch
```

ansible/roles/create-s390x-vm/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+    - vsi.rc != 0
+    - '"No Instance found" not in vsi.stderr'
+  register: vsi
+
+- name: Configure VSI
+  ibm.cloudcollection.ibm_is_instance:
+    name: "{{ vsi_name }}"
+    state: available
+    id: "{{ vsi.resource.id | default(omit) }}"
+    vpc: "{{ vpc_id }}"
```
Molter73, line 18: Any way we could create this id dynamically to spin up VMs on demand? Or are we stuck with the single VM? Our integration tests expect to be running on their own, so having a single VM in an environment where multiple CI runs could be triggered in parallel will be a problem, if we are to use a single VM, have you put some thought on how we could coordinate multiple CI jobs?

Molter73 (commented):
Hey @levivic! Sorry for the delay on the review, I've left a few questions in the PR.

Also, is there a link to the documentation for the `ibm.cloudcollection` ansible collector? All I could find is [this](https://galaxy.ansible.com/ibm/cloudcollection) and it doesn't seem to have much wrt explaining how to use it.

ansible/roles/create-s390x-vm/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+    resource_group: "{{ vsi_resource_group_id }}"
+    target: "{{ vsi.resource.primary_network_interface[0]['id'] }}"
+  register: fip
+
+- name: Print Floating IP Address
+  debug:
+    msg: "{{ vsi_name }} IP Address: {{ fip.resource.address }}"
+
+- name: Add VSI to Ansible inventory
+  add_host:
```
Stringy, line 51: How does this work with the wider inventory code? and how does this persist between executions of playbooks?

At the moment the process tends to be to create all the VMs first, then in a subsequent execution we run the tests (and a final step destroys the VMs.)

ansible/roles/create-s390x-vm/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+    groups:
+      - job_id_{{ job_id }}
+      - platform_{{ test_platform }}
+    ansible_ssh_extra_args: "-o StrictHostKeyChecking=no"
+    ansible_ssh_private_key_file: "~/.ssh/acs-sshkey_rsa.prv"
+    vm_config: "{{ vm_config }}"
+    vm_image_family: "{{ vm_image_family }}"
+    vm_collection_method: "{{ vm_collection_method }}"
+
+- name: Register server with Red Hat
```
Stringy, line 63: This feels like a provisioning step rather than a creation step

Stringy (commented):
I've added some questions about interactions with the existing dynamic inventories. I also have some other general comments that don't really fit into the code itself:

1. Rather than separate roles for s390 VM creation/deletion, could we move those into new task files within the existing create/destroy roles (in a similar way to how the provision-vm role works)
2. Can we have some additional notes in the README about how we're using the IBM cloud provisioning, or a link to specific IBM cloud docs. It is not immediately clear to me what we are creating, why we need a floating IP, etc.
3. Will the integration of these changes in GHA come in a follow up PR?

levivic:
> Also, is there a link to the documentation for the `ibm.cloudcollection` ansible collector? All I could find is [this](https://galaxy.ansible.com/ibm/cloudcollection) and it doesn't seem to have much wrt explaining how to use it.
 
Unfortunately, that's also the only reference I could find as well since it's release announcement in https://www.ibm.com/cloud/blog/announcements/ibm-cloud-collection-for-ansible. And I was mostly reading though its github examples and code https://github.com/IBM-Cloud/ansible-collection-ibm/tree/master/examples/simple-vm-ssh to educate myself.

ansible/roles/create-s390x-vm/tasks/main.yml
levivic: The vm creation is solely determined by `vsi_name`, equivalent to `vm_name` for x86 vms (https://github.com/stackrox/collector/blob/master/ansible/roles/create-vm/tasks/main.yml#L35-L38). If x86 is creating these ID dynamically during the CI process, then it should be the same for the z vms.

ansible/README.md
```diff
@@ -102,6 +103,9 @@ The following environment variables may be used to modify some behavior:
 | JOB_ID | A unique identifier to de-conflict VM names | the current user's username |
 | COLLECTOR_TEST | Which integration test make target to run. (e.g. integration-test-process-network) | ci-integration-tests |
 | VM_TYPE | Which kind of VMs to create on GCP (as listed above.) By default, will only build rhel VMs. Use 'all' to build an inventory containing every kind of VM. | rhel |
+| IC_API_KEY | API key for IBM Cloud. See [doc](https://cloud.ibm.com/docs/account?topic=account-userapikey&interface=ui) |
+| IC_REGION | The IBM Cloud region where you want to create your resources. See [doc](https://cloud.ibm.com/docs/workload-protection?topic=workload-protection-regions)|
```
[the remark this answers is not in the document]
levivic: These are two envs defined and used widely by IBM Cloud and its plugins, including the `ansbile cloudcollection` we are using here, i.e., https://github.com/IBM-Cloud/ansible-collection-ibm/tree/master/examples/simple-vm-ssh#set-api-key-and-region. So they are the default envs to be used.

ansible/roles/create-s390x-vm/tasks/main.yml
levivic: I think I heard Robby mentioned the x86 vms are created (through gcp) with subscription attached. And i don't see there is a registration process for x86 vm creation/provision. So I consider the registration one part of the vm creation process.

Stringy: For the purposes of these roles and playbooks, the distinction elsewhere is whether or not a connection to the VM is established for the step to take place. Anything that involves configuring the VM should be performed under the provisioning role, and I think that registering the subscription fits within that.

ansible/roles/create-s390x-vm/tasks/main.yml
levivic: I did test the vm-lifecycle playbook (i.e., creation, provision, destroy) with the `add_host` approach, and it was working ok. I expect the integration test (i.e,. vm creation, vm provision, test, vm destroy) should be ok, as long as they are in the same playbook. But this hasn't been tested yet since we don't have test environment ready (e.g., no multiarch image for building or testing).

ansible/roles/create-s390x-vm/tasks/main.yml
levivic: sure, I have followed the suggestion and made the subscription under provisioning role.

levivic:
Hi @Stringy , to answer your additional questions:

> 1. Rather than separate roles for s390 VM creation/deletion, could we move those into new task files within the existing create/destroy roles (in a similar way to how the provision-vm role works)

Done

> 2. Can we have some additional notes in the README about how we're using the IBM cloud provisioning, or a link to specific IBM cloud docs. It is not immediately clear to me what we are creating, why we need a floating IP, etc.

Done.

> 3. Will the integration of these changes in GHA come in a follow up PR?

Exactly, this PR (i.e., geting vm access ready) is only the first step towards the GHA CI. Once it's merged, we will work on the GHA integration.

ansible/roles/provision-vm/tasks/redhat.yml
```diff
@@ -1,13 +1,31 @@
...
+  community.general.redhat_subscription:
+    state: present
+    username: "{{ redhat_username }}"
+    password: "{{ redhat_password }}"
+    auto_attach: true
+  register: registration
+  retries: 2
+  delay: 30
+  until: registration is not failed
+  when: arch.stdout == "s390x"
```
Stringy, line 21: I think we can use `ansible_architecture` here, instead of finding this information ourselves

Stringy: It might also make sense to add this as a label on the VM during creation so we have this information throughout the various playbooks - I can see that it's used a fair bit and each time we're running `arch` to get it, so would be good to reduce this duplication a bit

ansible/dev/group_vars/all.yml
```diff
@@ -14,3 +15,14 @@ ansible_user: "{{ lookup('env', 'USER') }}"
...
+# IBM Cloud setup
+# vpc_name: lei-vpc-toronto
+vpc_id: r038-c4f8e233-4b56-44bd-af71-140ebd1664e9
+# subnet_name: stackrox-ci-sn3
+subnet_id: 02s7-0fca7f4f-7e53-40d5-a087-f8cb5d7ab682
+zone: ca-tor-3
+# ssh_key_name: acs-sshkey
+ssh_key_id: r038-fb0260c7-c01d-45c8-8026-7d50042943b9
+# vsi_resource_group: stackrox-ci-resource-group
+vsi_resource_group_id: 1a33a6a9bd6e498f8115e9b1064bfa97
```
Stringy, line 28: Will there be separate values for these vars for CI? If they're going to be static between dev and ci profiles, they should be moved to `group_vars/all.yml` otherwise we'll need these in `ci/group_vars/all.yml` as well

Stringy (commented):
Thanks @levivic it's looking like it's pretty much there, just a couple more questions/comments

ansible/dev/group_vars/all.yml
levivic: They should be the same for both `ci` and `dev`. So moved them under `group_vars/all.yml`. Thanks for the suggestion.

ansible/roles/provision-vm/tasks/redhat.yml
levivic: Good idea, i have set the arch info as group var so it's shared among playbooks/roles/tasks.

ansible/roles/create-s390x-vm/tasks/main.yml
levivic: After reading the workflow, I agree the inventory management is needed. Therefore, adding this commite https://github.com/stackrox/collector/pull/1249/commits/cebcaa94059e25defcc151ff6f4ed73e0c5c2bd2

Stringy: Looks good! One minor point, could we add `ansible/{dev,ci}/inventory_ibmcloud_z.yml` to `.gitignore` ? It's unlikely that it'll be committed but worth enforcing I think.

levivic: Done!

PR #1251, opened 2023-07-21
ADDED - Dockerfile for cbl-mariner2 kernel-modules

Smuggla:
## Testing Performed
Created Probes uses in stackrox deployments on "Azure Linux" [Mariner](https://github.com/microsoft/CBL-Mariner)
They are built upon new kernel deployed by Azure daily.
These can be indexed/downloaded at https://marinerprobes.azureedge.net/

### DIFF
--- a/kernel-modules/build/cbl-mariner2.Dockerfile
+++ b/kernel-modules/build/cbl-mariner2.Dockerfile
@@ -1,0 +1,25 @@
+FROM --platform=linux/amd64 mcr.microsoft.com/cbl-mariner/base/core:2.0 AS cbl-mariner-2.0-base
+
+RUN tdnf -y update && \
+    tdnf -y install \
+        binutils \
+        clang \
+        cmake \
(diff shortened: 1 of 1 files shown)


Smuggla:
Tagging @maddosaurus

Molter73:
Hi @Smuggla! Thanks for your PR!

Adding a new driver builder usually incurs extra effort on our end to make sure the builder and the drivers it builds actually work as intended. Further, with the current changes this dockerfile will not be used by our CI at all. So before we move forward I have a few questions:
- Have you tried building the CBL-mariner drivers using the existing Fedora builder? Their clang version differ (mariner uses 12 and fedora 14), but they should be close enough.
- Is the end goal here for us to officially support mariner? If so, please open an issue so we can further discuss the benefits of this change and the additional effort required for it.
- Finally, the latest version of collector has a BTF enabled probe that should be able to run on supported systems without an extra probe, have you tried running stackrox with it?

PR #1255, opened 2023-07-25
ROX-18860: force Core eBPF on s390x RHEL 8.4 kernels

kcrane:
## Description

eBPF is not supported on RHEL 8.4 kernels [ROX-18058](https://issues.redhat.com/browse/ROX-18058).  Force Core eBPF to be used.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

* Tested locally on rhel8.4 and rhel8.6 systems using a collector integration test and verified in collector logs the expected behaviour (rhel8.4 uses core and rhel8.6 uses ebpf probe).
* Tested on ocp 4.10 cluster running stackrox tests
* Tested on ocp 4.12 cluster running stackrox tests


collector/lib/HostInfo.h
```diff
@@ -146,6 +149,8 @@ struct KernelVersion {
   std::string release;
   // the entire version string (as in `uname -v`)
   std::string version;
+  // the kernel machine string (as in `uname -a`)
```
Stringy, line 152: [nit] `-a` gives everything, so a more precise command to get the machine would be `-m`

collector/lib/HostHeuristics.cpp
```diff
@@ -80,9 +80,39 @@ class DockerDesktopHeuristic : public Heuristic {
...
+    std::string os_id = host.GetOSID();
+
+    if (k.machine != "s390x") {
+      return;
+    }
+
+    if (os_id == "rhel" || os_id == "centos") {
+      // example release version: 4.18.0-305.88.1.el8_4.s390x
+      // build_id = 305
+      if (k.release.find(".el8.") != std::string::npos) {
```
Stringy, line 98: This would not match the example release version given - does this mean that the example is wrong or should this line be something like `k.release.find(".el8")`?

collector/lib/HostHeuristics.cpp
```diff
@@ -80,9 +80,39 @@ class DockerDesktopHeuristic : public Heuristic {
...
 
+class S390XHeuristic : public Heuristic {
+ public:
+  // S390X does not support eBPF ealier than 4.18.0-348 (rhel8.5) so we switch to use corebpf
+  // instead.
+  void Process(HostInfo& host, const CollectorConfig& config, HostConfig* hconfig) const {
+    auto k = host.GetKernelVersion();
+    std::string os_id = host.GetOSID();
+
+    if (k.machine != "s390x") {
```
Stringy, line 91: Might as well also check whether we're already configured to use core-bpf to avoid further processing

Stringy (commented):
Quick final question: once we have merged #1249 (and added the necessary GHA bits) will the changes in the PR be sufficiently exercised?

Stringy (approved):
You'll need to rebase to pick up some CI fixes in master, but this looks good to me and we can merge once CI passes.

robbycochran:
@kcrane It would be helpful to add to the PR description the test setup used to validate these changes.

collector/lib/HostHeuristics.cpp
```diff
@@ -57,9 +57,37 @@ class DockerDesktopHeuristic : public Heuristic {
...
+    std::string os_id = host.GetOSID();
+
+    if (k.machine != "s390x" || config.GetCollectionMethod() == CollectionMethod::CORE_BPF) {
+      return;
+    }
+
+    // example release version: 4.18.0-305.88.1.el8_4.s390x
+    // build_id = 305
+    if (k.release.find(".el8") != std::string::npos) {
+      if (k.kernel == 4 && k.major == 18 && k.minor == 0) {
```
robbycochran, line 75: I'm curious, why check for kernel version and build_id and not `el8_4`?

kcrane: followed same pattern as HostInfo.cpp isRHEL76 function.

robbycochran: For RHEL 7, the kernel version is always `.el7.` whether it is 7.5 or 7.6 etc. But for RHEL 8+ the kernel build includes the minor number. I think checking for el8_4 would be more reliable.

kcrane: ok, I can change this, rebuild and retest if you feel like we didn't capture the appropriate releases.

To be clear, you want the release, kernel version and build id check to be removed in place of searching the release string for ".el8_4."  ?

robbycochran: Yes, that is more straightforward. Especially after looking a little closer, the major build id doesn't change on patch releases.

| RHEL version | Kernel prefix | 
| -- | -- |
| RHEL 8.3 | 4.18.0-240 |
| RHEL 8.4 | 4.18.0-305 |
| RHEL 8.5 | 4.18.0-348 |
| RHEL 8.6 | 4.18.0-372 |

collector/test/HostInfoTest.cpp
```diff
@@ -114,45 +114,45 @@ TEST(KernelVersionTest, TestVersionStringPopulated) {
...
   std::string os_id = "rhel";
   EXPECT_TRUE(hasEBPFSupport(kernel, os_id));
 
   os_id = "coreos";
   EXPECT_FALSE(hasEBPFSupport(kernel, os_id));
 
-  kernel = KernelVersion("5.10.0", "");
+  kernel = KernelVersion("5.10.0", "", "");
   EXPECT_TRUE(hasEBPFSupport(kernel, os_id));
 }
```
robbycochran, line 158: Please also add unit tests for s390x on 8.4.

collector/lib/HostHeuristics.cpp
kcrane: pushed change for checking  "el8_4" in release string instead of checking on kernel version and build id.

collector/test/HostInfoTest.cpp
kcrane: added tests for HostInfo (for machine id) and HostHeuristics (for 8.4 probe switch)

robbycochran (approved):

PR #1259, opened 2023-07-27
Adds rhel-s390x to GHA integration tests

Stringy:
## Description

Follow up of #1249 to add rhel-s390x to integration tests in GHA.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Relying on CI.

### DIFF
--- a/.github/workflows/integration-tests-vm-type.yml
+++ b/.github/workflows/integration-tests-vm-type.yml
@@ -147,8 +147,6 @@ jobs:
         run: |
           make -C "${{ github.workspace }}/ansible" create-ci-vms
 
-          cat ansible/ci/inventory_ibmcloud_z.yml
-
       - name: Run Tests
         if: ${{ ! inputs.run-benchmarks }}
         run: |
--- a/.github/workflows/integration-tests-vm-type.yml
+++ b/.github/workflows/integration-tests-vm-type.yml
@@ -147,6 +147,8 @@ jobs:
         run: |
           make -C "${{ github.workspace }}/ansible" create-ci-vms
 
+          cat ansible/ci/inventory_ibmcloud_z.yml
+
       - name: Run Tests
         if: ${{ ! inputs.run-benchmarks }}
         run: |
--- a/.github/workflows/integration-tests-vm-type.yml
+++ b/.github/workflows/integration-tests-vm-type.yml
@@ -128,7 +128,7 @@ jobs:
             {
               echo "REDHAT_USERNAME=${REDHAT_USERNAME}"
               echo "REDHAT_PASSWORD=${REDHAT_PASSWORD}"
-            } > "${GITHUB_ENV}"
+            } >> "${GITHUB_ENV}"
           fi
         env:
           IBM_S390X_API_KEY: ${{ secrets.IBM_CLOUD_S390X_API_KEY }}
--- a/.github/workflows/integration-tests-vm-type.yml
+++ b/.github/workflows/integration-tests-vm-type.yml
@@ -123,8 +123,17 @@ jobs:
               echo "IC_REGION=ca-tor"
             } >> "${GITHUB_ENV}"
           fi
+
+          if [[ '${{ inputs.vm_type }}' =~ rhel ]]; then
+            {
+              echo "REDHAT_USERNAME=${REDHAT_USERNAME}"
+              echo "REDHAT_PASSWORD=${REDHAT_PASSWORD}"
+            } > "${GITHUB_ENV}"
+          fi
         env:
           IBM_S390X_API_KEY: ${{ secrets.IBM_CLOUD_S390X_API_KEY }}
+          REDHAT_USERNAME: ${{ secrets.REDHAT_USERNAME }}
+          REDHAT_PASSWORD: ${{ secrets.REDHAT_PASSWORD }}
 
       - name: Create VMs
         if: ${{ ! inputs.run-benchmarks }}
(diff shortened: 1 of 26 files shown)


ansible/group_vars/all.yml
```diff
@@ -1,8 +1,11 @@
 ---
 
+arch_info: "{{ 's390x' if 's390x' in test_platform else 'x86'}}"
+
 gcp_ssh_key_file: "{{ lookup('env', 'GCP_SSH_KEY_FILE', default='~/.ssh/google_compute_engine') }}"
+ibmcloud_ssh_key_file: "{{ lookup('env', 'IBMCLOUD_SSH_KEY_FILE', default='~/.ssh/acs-sshkey_rsa.prv') }}"
 
-ansible_ssh_private_key_file: "{{ gcp_ssh_key_file }}"
+ansible_ssh_private_key_file: "{{ ibmcloud_ssh_key_file if arch_info == 's390x' else gcp_ssh_key_file }}"
```
Molter73, line 8: In the interest of making it easier when we want to add ppc64le VMs, maybe we want to check for not x86?
```suggestion
ansible_ssh_private_key_file: "{{ ibmcloud_ssh_key_file if arch_info != 'x86' else gcp_ssh_key_file }}"
```
When the time comes to add arm, this will probably be incredibly messy.

ansible/roles/create-vm/templates/ibm-cloud-inventory.j2
Molter73, line 1: This honestly look like black magic, I'll assume is right, but dang...

ansible/roles/create-vm/templates/ibm-cloud-inventory.j2
```diff
@@ -0,0 +1,11 @@
+{% for item in groups %}
+
+[{{item}}]
+{%   for entry in groups[item] %}
```
Molter73, line 4: Could we do something like this? I think jinja templates are basically python and this might be a little more readable IMO.
```suggestion
{% for item, entries in groups.items() %}

[{{item}}]
{%   for entry in entries %}
```

ansible/roles/provision-vm/tasks/fedora-coreos.yml
```diff
@@ -40,4 +40,4 @@
   wait_for_connection:
     delay: 10
     sleep: 10
-    timeout: 60
+    timeout: 120
```
Molter73, line 43: Why do we need to change the timeout for fedora? Is it taking longer to process the playbooks?

ansible/roles/provision-vm/tasks/redhat.yml
```diff
@@ -1,13 +1,38 @@
...
+  community.general.redhat_subscription:
+    state: present
+    username: "{{ redhat_username }}"
+    password: "{{ redhat_password }}"
+    auto_attach: true
+  register: registration
+  retries: 2
+  delay: 30
+  until: registration is not failed
+  when: arch_info == "s390x"
```
Molter73, line 25: Maybe we should turn this into a block and have a single `when` condition?

ansible/roles/provision-vm/tasks/redhat.yml
```diff
@@ -1,13 +1,38 @@
...
+
+- debug:
+    var: distro
```
Molter73, line 28: Do we need to keep this debug statement? Doesn't hurt, I know, but still.

Molter73 (commented):
Overall looks good! Left a few comments here and there.

.github/workflows/integration-tests-vm-type.yml
```diff
@@ -96,9 +99,21 @@ jobs:
...
           GCP_SSH_KEY: ${{ secrets.GCP_SSH_KEY }}
           GCP_SSH_KEY_PUB: ${{ secrets.GCP_SSH_KEY_PUB }}
+          IBM_S390X_SSH_KEY: ${{ secrets.IBM_CLOUD_S390X_SSH_PRIVATE_KEY }}
+
+      - name: DEBUG verify ansible versions
+        run: |
+          ansible --version
+          ansible-galaxy collection list
+          python3 --version
+          python --version
```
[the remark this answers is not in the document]
Stringy: yes, definitely - must have missed this in my final pass

ansible/roles/create-vm/templates/ibm-cloud-inventory.j2
Stringy: that is a good point. I also want to try to make this a bit more readable by having the host variables on individual lines. Might mean converting this to yaml, but I think that's probably fine

ansible/roles/provision-vm/tasks/fedora-coreos.yml
Stringy: I'm currently trying to figure that out - the error seems to be that it hasn't fully rebooted yet at this point. Not sure what the root cause is currently

ansible/roles/provision-vm/tasks/redhat.yml
Stringy: good point, I'll remove

.github/workflows/integration-tests-vm-type.yml
```diff
@@ -109,11 +124,33 @@ jobs:
...
           fi
 
+          if [[ '${{ inputs.vm_type }}' =~ s390x ]]; then
+            {
+              echo "IC_API_KEY=${IBM_S390X_API_KEY}"
+              echo "IC_REGION=ca-tor"
+            } >> "${GITHUB_ENV}"
+          fi
+
+          if [[ '${{ inputs.vm_type }}' =~ rhel ]]; then
```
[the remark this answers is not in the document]
Stringy: they're unused I think, but following what you've previously suggested about setting even if we're not using them for a specific configuration - I think I'll just set them. The s390x vars above will be different on Power though, so I'll keep that check in

Molter73 (approved):
LGTM!

levivic:
Thanks @Stringy @Molter73!

PR #1264, opened 2023-07-28
Skip multiarch build in PRs

Molter73:
## Description

This PR moves steps for building and pushing the builder and slim images to an ansible playbook. The reason for this change is twofold:
- Initially, we want to be able to skip the multiarch steps in PRs that take a very long time and hinder feedback cycles, while also keeping the amount of complex checks in GHA to a minimum.
- Once we get access to native s390x and ppc64le machines, we want to move the build steps to use those, and having the workflow as an ansible playbook should make this transition easier.

## Checklist

## Testing Performed

- [x] Run with `skip-multiarch-builds` should skip ppc64le and s390x builds. ([CI run](https://github.com/stackrox/collector/actions/runs/5725025725/job/15512748936?pr=1264))
- [x] Run without `skip-multiarch-builds` should build the multiarch images and manifest. ([CI run](https://github.com/stackrox/collector/actions/runs/5725397137?pr=1264))
- [x] Run with `build-builder-image` should build and push the builder image alongside the collector images. ([CI run](https://github.com/stackrox/collector/actions/runs/5725025725/job/15512748936?pr=1264))


Makefile
Molter73, line 1: The changes in this file are needed because ansible is icky about environment variables being defined or not, so having a check on an actual value is just easier. From a dev experience point of view, the change should not affect workflows, since defining `BUILD_BUILDER_IMAGE` with any value other than `false` should still trigger the builder image build.

.github/workflows/collector-slim.yml
```diff
@@ -78,53 +70,54 @@ jobs:
...
+      - name: Build images
+        if: |
+          github.event_name == 'push' ||
+          matrix.arch == 'amd64' ||
+          !contains(github.event.pull_request.labels.*.name, 'skip-multiarch-builds')
         run: |
-          docker tag "quay.io/stackrox-io/collector-builder:${COLLECTOR_BUILDER_TAG}" "quay.io/rhacs-eng/collector-builder:${COLLECTOR_BUILDER_TAG}-${{ matrix.arch }}"
-          docker push "quay.io/rhacs-eng/collector-builder:${COLLECTOR_BUILDER_TAG}-${{ matrix.arch }}"
+          ansible-galaxy collection install community.general
+          ansible-galaxy collection install community.docker
```
Stringy, line 111: consider adding these to a `ansible-requirements.yml` file and installing with `ansible-galaxy collection install -r ansible-requirements.yml`

.github/workflows/collector-slim.yml
```diff
@@ -78,53 +70,54 @@ jobs:
...
-          username: ${{ secrets.QUAY_RHACS_ENG_RW_USERNAME }}
-          password: ${{ secrets.QUAY_RHACS_ENG_RW_PASSWORD }}
-
-      - name: Push builder to rhacs-eng
-        if: github.event_name == 'push' || needs.set-environment.outputs.collector-builder-tag != env.DEFAULT_BUILDER_TAG
+      - name: Build images
+        if: |
+          github.event_name == 'push' ||
+          matrix.arch == 'amd64' ||
+          !contains(github.event.pull_request.labels.*.name, 'skip-multiarch-builds')
```
Stringy, line 108: This makes me unreasonably happy

ansible/build-collector.yml
Stringy, line 1: These additions do raise an interesting question about ansible documentation and how we document each playbook, especially as we add more complex plays. For now, could you add a section to the README about this playbook (what it's for, a bit of detail on the args it expects, etc) and I'll figure out a better way of documenting playbooks more generally (interested to hear any ideas if you have them)

Stringy (approved):
Couple of pretty minor things, and some wider thoughts/questions, but otherwise LGTM! I look forward to not having to wait for multiarch builds all the time :)

ansible/build-collector.yml
Molter73: I'll add the section, we should also start considering whether we would benefit from running `ansible-lint` in our CI pipeline if we decide to lean even harder on playbooks.

PR #1265, opened 2023-07-31
Moves most notifications to oncall channel

Stringy:
## Description

Final (?) shift of notifications over to oncall channel. team-acs-collector-notifications now reserved for github notifications and the one notification about MODULE_VERSION (i.e. notifications that are not the responsibility of the signal handler)


Molter73 (commented):
I think you need to change `SLACK_COLLECTOR_NOTIFICATIONS_WEBHOOK` to `SLACK_COLLECTOR_ONCALL_WEBHOOK` as well

PR #1270, opened 2023-08-02
Prevent cache builders for specific archs from overwriting the multiarch manifest

Molter73:
## Description

After #1264 got merged, a weird behavior showed up on the master branch, the `cache` tag for the builder images that is meant to be the multiarch manifest for the latest builder image was being linked to the amd64 image only. Investigation showed that tagging and pushing via ansible is updating the `RepoTags` field to contain both the base image and the arch specific image.

RepoTags before #1264:
```
podman inspect --type image --format='{{ .RepoTags }}' quay.io/stackrox-io/collector-builder:cache-amd64
[quay.io/stackrox-io/collector-builder:cache-amd64]
```

RepoTags after #1264:
```
podman inspect --type image --format='{{ .RepoTags }}' quay.io/stackrox-io/collector-builder:cache-amd64
[quay.io/stackrox-io/collector-builder:cache quay.io/stackrox-io/collector-builder:cache-amd64]
```

This causes quay to directly point the `cache` tag to the `cache-amd` tag, when the behavior we want is to leave the `cache` tag unchanged.

In order to fix this, we can untag the `cache` tag before pushing, causing the `cache-amd` tag to go back to the way it was before the change.

## Checklist

## Testing Performed

- [x] Manually check the generated images have the correct `RepoTags` information.

### DIFF
--- a/.github/workflows/collector-slim.yml
+++ b/.github/workflows/collector-slim.yml
@@ -108,7 +108,7 @@ jobs:
           !contains(github.event.pull_request.labels.*.name, 'skip-multiarch-builds')
         run: |
           ansible-galaxy install -r ansible/requirements.yml
-          ansible-playbook -v \
+          ansible-playbook \
             --connection local \
             -i localhost, \
             --limit localhost \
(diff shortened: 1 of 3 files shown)


Stringy (approved):
LGTM!

PR #1271, opened 2023-08-04
Build multiarch full collector images

Molter73:
## Description

Use the downstream built drivers for P and Z alongside the downstream built images to build full collectors images.

## Checklist

## Testing Performed

- [x] Skipping multiarch builds in PRs should build a single x86 full image.


Stringy (approved):
LGTM! I think on the basis that this is currently passing CI and that final changes are in #1308 for upstream testing, we can probably merge this

PR #1273, opened 2023-08-04
ROX-18856: enable support for getsockopt

ovalenti:
## Description

We need the getsockopt_x syscall event in order to know when the status of an synchronous connection is known (#1243)

- _core-bpf_: cherry-pick a commit from upstream for a verifier failure (struct __kernel_timespec being changed in recent kernels). 
- _ebpf_: register getsockopt in the custom probe

## Checklist


Stringy (approved):
Fairly uncontentious changes :) LGTM

PR #1277, opened 2023-08-08
ROX-18838: Remove docker socket mountpoint from collector container.

ovalenti:
## Description

The current definition of the Collector daemonset contains a volume to access to the Docker socket. This is not used anymore and could cause security concern.

This PR removes the mountpoint parameters in the CI tests to check that Collector is not affected by this change.

## Checklist

### DIFF
--- a/integration-tests/suites/common/collector_manager.go
+++ b/integration-tests/suites/common/collector_manager.go
@@ -33,6 +33,7 @@ type CollectorManager struct {
 
 func NewCollectorManager(e Executor, name string) *CollectorManager {
 	collectorOptions := config.CollectorInfo()
+	runtimeOptions := config.RuntimeInfo()
 	image_store := config.Images()
 
 	collectionMethod := config.CollectionMethod()
@@ -50,9 +51,9 @@ func NewCollectorManager(e Executor, name string) *CollectorManager {
 		env["MODULE_DOWNLOAD_BASE_URL"] = "https://collector-modules.stackrox.io/612dd2ee06b660e728292de9393e18c81a88f347ec52a39207c5166b5302b656"
 	}
 	mounts := map[string]string{
-		// The presence of this file disables an optimisation, which turns off podman runtime parsing.
+		// The presence of this socket disables an optimisation, which would turn off podman runtime parsing.
 		// https://github.com/falcosecurity/libs/pull/296
-		"/run/podman/podman.sock:ro": "/tmp/dummy_podman_socket_mount",
+		"/run/podman/podman.sock:ro": runtimeOptions.Socket,
 		"/host/proc:ro":              "/proc",
 		"/host/etc:ro":               "/etc/",
 		"/host/usr/lib:ro":           "/usr/lib/",


integration-tests/suites/config/config.go
```diff
@@ -5,7 +5,6 @@ const (
 	CollectionMethodCoreBPF = "core-bpf"
 
 	runtimeDefaultCommand = "docker"
-	runtimeDefaultSocket  = "/var/run/docker.sock"
```
Molter73, line 8: I think this is messing with this other line.
https://github.com/stackrox/collector/blob/485640ef687aeb3064ea5722fb3ddac6c093438b/integration-tests/suites/config/config.go#L131

You probably want to add it back in.

Molter73: And the one in integration-tests/suites/config/env.go

ovalenti: Argh, I missed that usage of the config object. I restored it and the CI now seems happy.

Molter73:
Looks like RHCOS is showing the same behavior I was getting on my Fedora VM, I wonder if there's something about the format for the cgroups in that platform that is not supported by Falco and we were falling back to using the container engine for getting that data somehow 🤔 

It'd be great if we could somehow log what container runtime is being used, I suspect crun is not supported by Falco and is what I'm using on my environment.
```
  ociRuntime:
    name: crun
    package: crun-1.8.6-1.fc38.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 1.8.6
      commit: 73f759f4a39769f60990e7d225f561b4f4f06bcf
      rundir: /run/crun
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +LIBKRUN +WASM:wasmedge +YAJL
```

Molter73:
Ok, I've been doing some debugging and I've found why this is failing on systems with podman.

Usually, the container ID is detected by the container engine in falco, some of the engines are run by default, but podman has this little jewel: https://github.com/stackrox/falcosecurity-libs/blob/653e24c25d10c1a7ac16455ec41d0c6bc5bfd219/userspace/libsinsp/container_engine/docker/podman.cpp#L177-L182

That piece of code is quite weird, but the basic idea is, "if the podman socket doesn't exists, don't try to match podman type cgroups", causing Falco to not be able to parse container IDs on systems running podman. We will need to figure out a way to either:
- Mock the collector filesystem so it thinks it has podman always.
- Have some compile time check that allows us to bypass that check.

ovalenti:
Good catch @Molter73 !

> That piece of code is quite weird, but the basic idea is, "if the podman socket doesn't exists, don't try to match podman type cgroups"

I was searching for the reason behind this particular exception, and it would be a performance trick: https://github.com/falcosecurity/libs/pull/296

ovalenti:
Also, none of the end-user runtime environments we support uses podman, nor docker... so I wonder whether the problem is in the code, ... or the test environment :grimacing:

Molter73:
> Good catch @Molter73 !
> 
> > That piece of code is quite weird, but the basic idea is, "if the podman socket doesn't exists, don't try to match podman type cgroups"
> 
> I was searching for the reason behind this particular exception, and it would be a performance trick: [falcosecurity/libs#296](https://github.com/falcosecurity/libs/pull/296)

Yeah, I assumed something of the sort, since it has to look for the socket in multiple users. 

> Also, none of the end-user runtime environments we support uses podman, nor docker... so I wonder whether the problem is in the code, ... or the test environment 😬

For the time being, I think we can keep the podman mount to trick the integration tests into passing, but we'll need to look for a better solution. I'll talk to folks upstream, cause there is a mask that you can set at runtime to skip any engines you don't use, so I think it makes sense to move the check out of the libraries and into Falco itself.

The sad part in all this is that we will still have the code for querying container APIs, even if they fail and there won't be an easy way to decouple them.

Molter73 (approved):
LGTM!

ovalenti:
TIL: yet an other difference between podman and docker
https://github.com/containers/podman/issues/6234#issuecomment-628842681

Molter73 (approved):

PR #1282, opened 2023-08-10
ROX-18716: integration tests for asynchronous connections

ovalenti:
## Description

- when an asynchronous connection is successful, it is reported
- when an asynchronous connection fails, it should not be reported. It currently is, hence the test is disabled. But #1243 will fix it !

## Checklist


integration-tests/suites/async_connections.go
```diff
@@ -0,0 +1,96 @@
...
+
+	err = s.collector.TearDown()
+	s.Require().NoError(err)
+
+	s.db, err = s.collector.BoltDB()
+	s.Require().NoError(err)
+}
+
+func (s *AsyncConnectionTestSuite) TearDownSuite() {
+	s.cleanupContainer([]string{"server", "client"})
```
JoukoVirtanen, line 75: Do you need to add nginx here?

integration-tests/suites/base.go
```diff
@@ -282,6 +282,24 @@ func (s *IntegrationTestSuiteBase) GetProcesses(containerID string) ([]common.Pr
 	return processes, nil
 }
 
+func (s *IntegrationTestSuiteBase) HasNetworks(containerID string) (bool, error) {
```
JoukoVirtanen, line 285: I don't think this function is needed.

integration-tests/suites/async_connections.go
```diff
@@ -0,0 +1,96 @@
...
+		s.Require().NoError(err)
+
+		assert.Equal(s.T(), 1, len(networkInfos))
+	} else {
+		// expect no connections at all
+		hasNetwork, err := s.HasNetworks(s.clientContainer)
+		s.Require().NoError(err)
+
+		assert.True(s.T(), !hasNetwork, "A failed asynchronous connection should not be reported")
+	}
```
JoukoVirtanen, line 95:
```suggestion
        networkInfos, err := s.GetNetworks(s.clientContainer)
	if !s.BlockConnection {
		// expect one connection		
		s.Require().NoError(err)

		assert.Equal(s.T(), 1, len(networkInfos))
	} else {
		// expect no connections at all
		s.Require().Error(err)
	}
```

integration-tests/suites/base.go
ovalenti: GetNetworks() returns an error when there are no networks. Do you prefer to change this and return an empty set ?

integration-tests/suites/async_connections.go
ovalenti: Well, that could be any kind of error. Overall, not receiving networks does not seem like an error to me, especially when you expect that there will be no networks reported.

integration-tests/suites/base.go
ovalenti: I implemented as suggested below, nevemind.

integration-tests/suites/async_connections.go
ovalenti: Implemented as suggested.

integration-tests/container/async-connection/curl-test.c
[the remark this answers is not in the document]
ovalenti: I checked that the `curl` tool makes the calls we need, and there is an option to set the connection timeout :tada: 

Switched to making use of the existing image.

integration-tests/suites/async_connections.go
```diff
@@ -0,0 +1,97 @@
...
+
+	time.Sleep(5 * time.Second) // TODO use the endpoint declaration
+
+	target := s.serverIP
+
+	if s.BlockConnection {
+		target = "10.255.255.1"
+	}
+
+	containerID, err = s.launchContainer("client", image_store.QaImageByKey("qa-schedule-curls"), "curl", "--connect-timeout", "5", fmt.Sprintf("http://%s/", target))
```
JoukoVirtanen, line 64: I don't think you need to use the qa-schedule-curls image here. It is a bit confusing as it seems that you are using the script, schedule-curls.sh, the image has. The qa-schedule-curls image uses the pstauffer/curl image. Perhaps you can use that image directly.

integration-tests/suites/async_connections.go
JoukoVirtanen: Maybe you could check for the specific error? Maybe in a different PR we could make it so that not finding the bucket is not an error. I will approve as is. There is another unrelated comment that I left that I would like to be resolved before merging.

ovalenti: I made an attempt to return empty set; but expectations that the set is empty or not-empty are rather hard to read. I believe that if we do this in a subsequent PR, it would still be beneficial if we have functions that expect empty or something, instead of expect(x.len() > 0)

Stringy (approved):

PR #1283, opened 2023-08-11
Improve ansible output

Stringy:
## Description

To reduce the amount of useless information in the ansible output (setup tasks, skipped tasks etc) this PR enables the `selective` stdout callback which will only print tasks that are explicitly opt-in via the `print_action` task, with the exception of failures which are always printed

As a first pass, this PR also opts-in with integration testing tasks and build tasks, but none of the VM creation or provisioning task. This is by no means set in stone and if more tasks make sense for logging then we can opt-in.

The advantage should be that the CI logs are more readable and we can identify flakes/failures more easily. 

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed
Tested locally. Example output, including some successes and failures:

```
ansible-playbook \
		-i dev \
		-e job_id="ghutton" \
		-e @secrets.yml \
		--tags run-tests \
		integration-tests.yml
............
# run-test-target : Run integration tests *********************************************************************************
  * collector-dev-any-rhel-8-ghutton- FAILED!!! -- non-zero return code -----------------------
.
# run-test-target : Output integration test log ***************************************************************************
  * collector-dev-any-rhel-8-ghutton- changed=False ---------------------------------------------
    docker rm -fv container-stats benchmark collector grpc-server 2>/dev/null || true
    go version
    go version go1.20.2 darwin/amd64
    set -o pipefail; go test -tags bench -timeout 60m -count=1 -v -run TestProcessNetwork 2>&1 | tee -a integration-test.log
    === RUN   TestProcessNetwork
        require.go:794:
            	Error Trace:	process_network.go:74
            	            				suite.go:64
            	            				integration_test.go:14
            	Error:      	Received unexpected error:
            	            	invalid character '@' looking for beginning of value
            	Test:       	TestProcessNetwork
    --- FAIL: TestProcessNetwork (44.48s)
    FAIL
    exit status 1
    FAIL	github.com/stackrox/collector/integration-tests	45.108s
# run-test-target : Check results *****************************************************************************************
  * collector-dev-any-rhel-8-ghutton- FAILED!!! -- Tests failed -------------------------------
  * collector-dev-any-rhel-9-ghutton- changed=True --  ----------------------------------------
..
# run-test-target : Output integration test log ***************************************************************************
  * collector-dev-any-rhel-9-ghutton- changed=False ---------------------------------------------
    docker rm -fv container-stats benchmark collector grpc-server 2>/dev/null || true
    go version
    go version go1.20.2 darwin/amd64
    set -o pipefail; go test -tags bench -timeout 60m -count=1 -v -run TestProcessNetwork 2>&1 | tee -a integration-test.log
    === RUN   TestProcessNetwork
    === RUN   TestProcessNetwork/TestNetworkFlows
    ServerDetails from Bolt: 2902e6934451 {LocalAddress::80 RemoteAddress:172.17.0.4 Role:ROLE_SERVER SocketFamily:SOCKET_FAMILY_UNKNOWN CloseTimestamp:(timestamp: nil Timestamp)}
    ServerDetails from test: 2902e6934451 172.17.0.3, Port: 80
    ClientDetails from Bolt: 2902e6934451 {LocalAddress: RemoteAddress:172.17.0.3:80 Role:ROLE_CLIENT SocketFamily:SOCKET_FAMILY_UNKNOWN CloseTimestamp:(timestamp: nil Timestamp)}
    ClientDetails from test: 5e0273a68238 172.17.0.4
    === RUN   TestProcessNetwork/TestProcessLineageInfo
    === RUN   TestProcessNetwork/TestProcessViz
    CPU: Container grpc-server, Mean 0.016250000000000004, StdDev 0.036571847095819475
    CPU: Container collector, Mean 3.945714285714285, StdDev 6.519131515338966
    Writing perf.json
    --- PASS: TestProcessNetwork (109.26s)
        --- PASS: TestProcessNetwork/TestNetworkFlows (0.00s)
        --- PASS: TestProcessNetwork/TestProcessLineageInfo (0.00s)
        --- PASS: TestProcessNetwork/TestProcessViz (0.00s)
    PASS
    ok  	github.com/stackrox/collector/integration-tests	109.821s

# STATS *******************************************************************************************************************
collector-dev-any-rhel-8-ghutton    : ok=8	changed=2	failed=1	unreachable=0	rescued=1	ignored=0
collector-dev-any-rhel-9-ghutton    : ok=10	changed=5	failed=0	unreachable=0	rescued=0	ignored=0
make: *** [integration-tests] Error 2 
```

### DIFF
--- a/ansible/ci-build-collector.yml
+++ b/ansible/ci-build-collector.yml
@@ -12,6 +12,8 @@
       community.general.make:
         chdir: "{{ lookup('env', 'GITHUB_WORKSPACE') }}"
         target: image
+      # ensure this action is printed
+      tags: [print_action]
 
     - name: Retag collector image to arch specific
       community.docker.docker_image:
@@ -71,6 +73,7 @@
         - "  {{ collector_image }}-{{ arch }}-base"
         - "  {{ ansible_env.RHACS_ENG_IMAGE }}-{{ arch }}-slim"
         - "  {{ ansible_env.RHACS_ENG_IMAGE }}-{{ arch }}-base"
+      tags: [print_action]
 
     - name: Logout of quay.io
       community.docker.docker_login:
--- a/ansible/ci-build-collector.yml
+++ b/ansible/ci-build-collector.yml
@@ -22,6 +22,10 @@
             msg: "{{ build_result.stdout }}"
           tags: [print_action]
 
+        # propagate the failure (otherwise the tasks below will run)
+        - ansible.builtin.fail:
+            msg: Build failed
+
     - name: Retag collector image to arch specific
       community.docker.docker_image:
         name: "{{ collector_image }}"
--- a/ansible/ci-build-collector.yml
+++ b/ansible/ci-build-collector.yml
@@ -8,12 +8,19 @@
     COLLECTOR_TAG: "{{ ansible_env.COLLECTOR_TAG }}"
 
   tasks:
-    - name: Build the collector image
-      community.general.make:
-        chdir: "{{ lookup('env', 'GITHUB_WORKSPACE') }}"
-        target: image
-      # ensure this action is printed
-      tags: [print_action]
+    - block:
+        - name: Build the collector image
+          community.general.make:
+            chdir: "{{ lookup('env', 'GITHUB_WORKSPACE') }}"
+            target: image
+          register: build_result
+          # ensure this action is printed
+          tags: [print_action]
+      rescue:
+        - name: Output built log
+          debug:
+            msg: "{{ build_result.stdout }}"
+          tags: [print_action]
 
     - name: Retag collector image to arch specific
       community.docker.docker_image:
--- a/ansible/ci-build-collector.yml
+++ b/ansible/ci-build-collector.yml
@@ -16,15 +16,11 @@
           register: build_result
           # ensure this action is printed
           tags: [print_action]
-      rescue:
+      always:
         - name: Output built log
           debug:
             msg: "{{ build_result.stdout }}"
           tags: [print_action]
-
-        # propagate the failure (otherwise the tasks below will run)
-        - ansible.builtin.fail:
-            msg: Build failed
 
     - name: Retag collector image to arch specific
       community.docker.docker_image:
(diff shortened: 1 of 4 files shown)


ansible/ci-build-collector.yml
```diff
@@ -13,6 +13,8 @@
       community.general.make:
         chdir: "{{ lookup('env', 'GITHUB_WORKSPACE') }}"
         target: image
+      # ensure this action is printed
+      tags: [print_action]
```
Stringy, line 17: thinking about it, we'll also need a task to always print the output of this make target (or conditionally on failure only?)

Molter73: I think printing the output on failure is what we want, success just means the image was built correctly.

Stringy: it might be worth doing that with the tests as well

ansible/ci-build-collector.yml
```diff
@@ -8,10 +8,23 @@
...
+          community.general.make:
+            chdir: "{{ lookup('env', 'GITHUB_WORKSPACE') }}"
+            target: image
+          register: build_result
+          # ensure this action is printed
+          tags: [print_action]
+      rescue:
+        - name: Output built log
+          debug:
+            msg: "{{ build_result.stdout }}"
```
Molter73, line 22: Do we maybe need `build_result.stderr` as well? I would assume compilation errors would go to that stream

Molter73 (approved):
Overall LGTM! Just a minor question, but otherwise no reason not to merge IMO.

ansible/ci-build-collector.yml
Stringy: I just double checked this behaviour and if the build task fails above, we get all the stderr output printed by default, which actually makes me feel like this rescue block is unnecessary because we probably only care about the errors and not the full build log. Alternatively, if we also want the full build log, we can change this to `always` - WDYT?

Molter73: Sounds good, we can keep build log, should make it a little easier to understand what is going on during failures.

PR #1284, opened 2023-08-11
Skip builder image build when it is unmodified

Molter73:
## Description

Skip the builder image build when no changes would make modifications to the actual image (i.e.: no change to third_party libraries and the builder directory itself).

## Checklist

## Testing Performed
- [x] Run with `build-builder-image` should force build the builder.
- [x] Run with a change to either the `third_party` or `builder` directories should build the builder. ([CI run](https://github.com/stackrox/collector/actions/runs/5834100265?pr=1284))
- [x] Running with no changes to the `third_party` or `builder` directories should skip building the builder. ([CI run](https://github.com/stackrox/collector/actions/runs/5853649433/job/15870801596?pr=1284))

### DIFF
--- a/.github/workflows/collector-builder.yml
+++ b/.github/workflows/collector-builder.yml
@@ -8,6 +8,10 @@ on:
         required: true
         description: |
           The tag used to build the collector image
+    outputs:
+      collector-builder-tag:
+        description: The builder tag used by the build
+        value: ${{ jobs.build-builder-image.outputs.collector-builder-tag }}
 
 env:
   COLLECTOR_TAG: ${{ inputs.collector-tag }}
--- a/.github/workflows/collector-builder.yml
+++ b/.github/workflows/collector-builder.yml
@@ -1,0 +1,137 @@
+name: Collector builder image build
+
+on:
+  workflow_call:
+    inputs:
+      collector-tag:
+        type: string
+        required: true
+        description: |
+          The tag used to build the collector image
+
+env:
+  COLLECTOR_TAG: ${{ inputs.collector-tag }}
+  DEFAULT_BUILDER_TAG: cache
+
+jobs:
+  builder-needs-rebuilding:
+    name: Determine if builder image needs to be built
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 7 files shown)


.github/workflows/collector-builder.yml
```diff
@@ -0,0 +1,141 @@
...
+      - name: Define builder tag
+        id: builder-tag
+        run: |
+          COLLECTOR_BUILDER_TAG="${DEFAULT_BUILDER_TAG}"
+          if [[ "${{ github.event_name }}" == 'pull_request' ]]; then
+            COLLECTOR_BUILDER_TAG="${{ inputs.collector-tag }}"
+          fi
+
+          echo "COLLECTOR_BUILDER_TAG=${COLLECTOR_BUILDER_TAG}" >> "$GITHUB_ENV"
+          echo "collector-builder-tag=${COLLECTOR_BUILDER_TAG}" >> "$GITHUB_OUTPUT"
```
ovalenti, line 77: When the builder does not need to be rebuilt, how is the `collector-build-tag` going to be provided to the collector build ?

Molter73: I'll test it, but when skipping I think the output turns into an empty string and our makefiles turn an empty `COLLECTOR_BUILDER_TAG` env variable to the default (`cache`):
https://github.com/stackrox/collector/blob/d6cd0ff497582e56104911fdb5af966cdbeee077/Makefile-constants.mk#L2-L4

Molter73: Actually, in the latest CI run, that is what happened:
image

https://github.com/stackrox/collector/actions/runs/5856587306/job/15876735578?pr=1284

ovalenti (approved):
This is going to save so much time and resources !

Apart from the nitpick comment for the default value of the builder tag, this look's good to me :+1:

.github/workflows/collector-builder.yml
ovalenti: I think that skipping setting of the value when we are in the default case would be coherent (have the default value's "value" in one place). But, hey, it works !

PR #1289, opened 2023-08-16
Remove stale pull of builder image with skip-multiarch

ovalenti:
## Description

We used to pull the builder image in the `skip-multiarch-builds`, which now breaks when the builder's build is skipped (builder tag not set).

This seems uneeded now, so let's just remove it.


Molter73 (approved):
Sorry I missed this one! Thanks for putting together the PR!

LGTM!

PR #1291, opened 2023-08-16
Automated integration test images builds

Stringy:
## Description

Automates the creation of the containers in use by the integration tests. This will occur only when the files in `integration-tests/container` change, rather than on every build and will only push the images on master rather than in each run of a PR.

This also includes the creation of multi-arch manifests for each of the images.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed
None yet.


integration-tests/suites/process_network.go
Stringy, line 1: The changes here are directly related to the new `containerstats` image which no longer uses `docker:dind` but instead is based off `stream9`

.github/workflows/integration-test-containers.yml
```diff
@@ -0,0 +1,136 @@
...
+            # are tagged, and that tag is passed to the integration test and applied to all images.
+            #
+            dirs=$(find integration-tests/container/ -type d -mindepth 1 -maxdepth 1)
+            qa_tag="$(echo "${GITHUB_SHA}" | cut -c1-8)"
+          else
+            #
+            # On master, we have the luxury of only building things that have changed,
+            # because the QA tag is not applied (instead we'll update the floating tags)
+            #
+            dirs=$(dirname ${{ needs.test-containers.outputs.changed_files }} | sort | uniq)
```
ovalenti, line 56: We will have to remember that subdirectories are forbidden here, this is probably acceptable.

.github/workflows/integration-test-containers.yml
```diff
@@ -0,0 +1,136 @@
...
+            qa_tag=
```
ovalenti, line 57: push action on release branches may rebuild and override master's QA images, and vice-versa

integration-tests/container/benchmark/Makefile
```diff
@@ -7,9 +10,22 @@ COLLECTOR_QA_BENCHMARK_TAG=phoronix-$(COLLECTOR_QA_TAG)
...
+.PHONY: build-and-push
+build-and-push: build
+	@docker push quay.io/rhacs-eng/collector-performance:$(COLLECTOR_QA_BENCHMARK_TAG)-$(IMAGE_ARCH)
+
+.PHONY: manifest
+manifest:
+	@docker manifest create quay.io/rhacs-eng/collector-performance:$(COLLECTOR_QA_BENCHMARK_TAG) \
+		quay.io/rhacs-eng/collector-performance:$(COLLECTOR_QA_BENCHMARK_TAG)-amd64 \
+		quay.io/rhacs-eng/collector-performance:$(COLLECTOR_QA_BENCHMARK_TAG)-s390x \
+		quay.io/rhacs-eng/collector-performance:$(COLLECTOR_QA_BENCHMARK_TAG)-ppc64le
```
ovalenti, line 29: [nit] not sure about this, but we could have a list of archs in `Makefile-constants.mk` and iterate over it. It would avoid changing the list everywhere upon addition. However, a change to the constants is not watched for modifications, and the current code is readable... so definitely not sure :smile:

ovalenti (changes requested):
This is a very nice PR and some nice tricks here (loving it :sweat_smile: ).

I am only concerned by the potential collision between master's qa-tag and release branches, what do you think ?

integration-tests/container/QA_TAG
```diff
@@ -0,0 +1 @@
+1.0.0
```
ovalenti, line 1: What scheme do you envision for increasing the version ? My lack of opinion here would lead me use one plain integer for master, then something like `release-x.x-<integer>` for the release branches.

ovalenti (changes requested):
I think that this simple and powerful, and addresses all the corner-cases I can think of:
- it handles the "PR on top of a PR"
- release branches backwards compatibility

And as bonus:
- since master is not a sliding tag, no need to update the QA_TAG when starting release branches until we actually touch the images in that branch.
- saves a lot of space as compared to the previous strategy.

I will be happy to approve when the extraneous comment in .github/workflows/integration-test-containers.yml is fixed; and maybe a touch of documentation ?

ovalenti (approved):
Once the integration-tests pass, this is good to go for me !

It was not easy, but I think that the result is awesome, thanks @Stringy :+1:

ovalenti:
Sorry, wrong button...

The latest failure seems due to an entrypoint definition in the socat container.

This got me a bit further: https://github.com/stackrox/collector/commit/82182e737ae6f641d219ddd7cc9d9fae8b5bb6bc

PR #1298, opened 2023-08-21
Check the correct subdirectory for downstream drivers

Molter73:
## Description

The subdirectory being checked for downstream drivers is wrong, it should check the `x86_64` subdirectory.

## Checklist

## Testing Performed

- [x] Manually checked the command correctly returns the drivers in the bucket.


Molter73:
The CI failures are due to a failing kernel that is also found on master, unrelated to the changes in this PR.

erthalion (approved):

PR #1302, opened 2023-08-21
Plot benchmark data for collector

Molter73:
## Description

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -49,15 +49,17 @@ jobs:
           ./integration-tests/scripts/baseline/main.py --test integration-tests/perf-all.json \
             | sort \
             | awk -f ./integration-tests/scripts/baseline/format.awk > benchmark.md
-
-          delimiter=$(openssl rand -hex 8)
-          {
-            echo "PERF_TABLE<<${delimiter}"
-            cat benchmark.md
-            echo "${delimiter}"
-          } >> "$GITHUB_ENV"
         env:
           GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT: "${{ secrets.GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT }}"
+
+      - name: Plot benchmark graphs
+        if: github.event_name == 'pull_request'
+        run: |
+          python3 -m pip install -r utilities/perf-plotter/requirements.txt
+          ./utilities/perf-plotter/plot-perf.py integration-tests/perf-all.json -o /tmp/perf.png
+
+          # upload perf graph to public GCP bucket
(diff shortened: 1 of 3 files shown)


utilities/perf-plotter/plot-perf.py
```diff
@@ -0,0 +1,96 @@
...
+    # Create a single legend for both subplots
+    handles, labels = memplot.get_legend_handles_labels()
+    fig.legend(handles, labels, loc='lower right')
+    fig.set_figheight(12)
+    fig.set_figwidth(12)
+
+    cpuplot.set_ylabel('CPU usage (%)')
+    memplot.set_ylabel('Memory usage (MiB)')
+
+    if len(output) == 0:
```
Stringy, line 81: [nit]

```suggestion
    if output:
```

utilities/perf-plotter/plot-perf.py
```diff
@@ -0,0 +1,96 @@
...
+import argparse
+import json
+import sys
+import datetime as dt
+
+import matplotlib.pyplot as plt
+import matplotlib.dates as md
+
+
+def bytes_to_float(bytes: str) -> float:
```
Stringy, line 18: [nit] this function can return None if the suffix doesn't match `B|Mi|Ki|MiB|KiB` - I think either we want to remove the `else` so we always try to do `float(bytes) / 1000000` if the other statements don't match, or explicitly throw an exception. It's not a _huge_ concern though, given how this script is used

There is also the ability to match on `Mi|Ki` which I think is also unwanted (but again, not huge concern given the inputs and the intended use)

.github/workflows/benchmarks.yml
```diff
@@ -49,16 +49,30 @@ jobs:
...
-            echo "PERF_TABLE<<${delimiter}"
-            cat benchmark.md
-            echo "${delimiter}"
-          } >> "$GITHUB_ENV"
         env:
           GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT: "${{ secrets.GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT }}"
 
+      - name: Plot benchmark graphs
+        run: |
+          python3 -m pip install -r utilities/perf-plotter/requirements.txt
```
Stringy, line 57: Consider moving this to the `Install python deps` step, to keep all the dependency setup in the same place

Stringy (approved):
LGTM! Couple of minor comments but nothing that should hold up merging.

utilities/perf-plotter/plot-perf.py
Molter73: Yeah, I thought about raising a runtime exception, then realized I didn't really care enough to put the effort in 🤣 , I'll tweak it a bit.

PR #1305, opened 2023-08-23
Add scan-build target

Molter73:
## Description

The new target will compile the collector and connscrape binaries with scan-build. This target is also added to CI as its own job.

## Checklist

## Testing Performed

TBD


.github/workflows/main.yml
```diff
@@ -91,6 +91,34 @@ jobs:
...
+    env:
+      COLLECTOR_BUILDER_TAG: ${{ needs.build-builder-image.outputs.collector-builder-tag }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+
+      - name: Run scan build
+        run: |
+          set -x
```
JoukoVirtanen, line 106: We probably don't want that in CI

JoukoVirtanen:
I think running this locally and showing the output and running CI is sufficient. You could run locally with and without the two code fixes that you made and paste the output here. If you do that and rebase and address the one small comment I made, I will approve.

JoukoVirtanen:
Also if CI fails without the code fixes, that is good and should be noted in the PR. If CI passes without the code fixes, that should also be noted.

.github/workflows/main.yml
Molter73: Yup, I was debugging something apparently, I'll remove it.

PR #1313, opened 2023-09-04
ROX-19901: Removes BoltDB & mock gRPC server from tests, in favor of in-tree gRPC server

Stringy:
## Description

In order to simplify testing and provide a more deterministic mechanism for retrieving events from Collector, this PR makes use of gRPC directly, rather than going via the mock gRPC server and BoltDB.

This PR is large, and as a result its scope is constrained only to the in-tree gRPC server and associated test refactoring.

Things that will be fixed in a follow up and NOT in this PR:

- simplified test setup
- deterministic event generation
- setup / teardown stability

## Checklist

**Automated testing**
~- [ ] Added unit tests~

## Testing Performed

Run all tests locally against a RHEL-9 VM using Collector 3.16.0 other platforms will be covered by CI, and the results will be investigated accordingly.

- [ ] run CI on all platforms
- [ ] run CI on s390x and Power


integration-tests/suites/mock_sensor/server.go
```diff
@@ -0,0 +1,460 @@
...
+
+// LiveLineages returns a channel that can be used to read live
+// process lineage events
+func (m *MockSensor) LiveLineages() <-chan *storage.ProcessSignal_LineageInfo {
+	return m.lineageChannel
+}
+
+// ProcessLineages returns a list of all processes that have been received for
+// a given container ID
+func (m *MockSensor) ProcessLineages(containerID string) []types.ProcessLineage {
```
Molter73, line 123: Don't we need to take `m.processMutex` here?

integration-tests/suites/mock_sensor/server.go
```diff
@@ -0,0 +1,460 @@
...
+
+// LiveConnections returns a channel that can be used to read live
+// connection events
+func (m *MockSensor) LiveConnections() <-chan *sensorAPI.NetworkConnection {
+	return m.connectionChannel
+}
+
+// Connections returns a list of all connections that have been received for
+// a given container ID
+func (m *MockSensor) Connections(containerID string) []types.NetworkInfo {
```
Molter73, line 156:
```suggestion
func (m *MockSensor) Connections(containerID string) []types.NetworkInfo {
	m.networkMutex.Lock()
	defer m.networkMutex.Unlock()
```

integration-tests/suites/mock_sensor/server.go
```diff
@@ -0,0 +1,460 @@
...
+
+// Liveendpoints returns a channel that can be used to read live
+// endpoint events
+func (m *MockSensor) LiveEndpoints() <-chan *sensorAPI.NetworkEndpoint {
+	return m.endpointChannel
+}
+
+// Endpoints returns a list of all endpoints that have been received for
+// a given container ID
+func (m *MockSensor) Endpoints(containerID string) []types.EndpointInfo {
```
Molter73, line 189:
```suggestion
func (m *MockSensor) Endpoints(containerID string) []types.EndpointInfo {
	m.networkMutex.Lock()
	defer m.networkMutex.Unlock()
```

integration-tests/suites/mock_sensor/server.go
```diff
@@ -0,0 +1,460 @@
...
+	ipPortPair := utils.NetworkPeerID{
+		Address: utils.IPFromBytes(addr.GetAddressData()),
+		Port:    uint16(addr.GetPort()),
+	}
+	return ipPortPair.String()
+}
+
+func (m *MockSensor) logEvent(format string, args ...interface{}) {
+	m.logFile.WriteString(fmt.Sprintf(format+"\n", args...))
+}
```
Molter73, line 460: I would assume go has some logging library we could use somewhere 🤔

Molter73 (commented):
This is great! The tests look a lot more streamlined are easier to follow IMO.

Left a few comments and we need to address some of the commented out checks we have before merging, but overall looking great

integration-tests/suites/listening_ports.go
```diff
@@ -16,24 +17,23 @@ type ProcessListeningOnPortTestSuite struct {
...
-	s.Require().NoError(err)
+	collector.Env["COLLECTOR_CONFIG"] = `{"logLevel":"debug","turnOffScrape":false,"scrapeInterval":1}`
+	collector.Env["ROX_PROCESSES_LISTENING_ON_PORT"] = "true"
 
-	err = s.collector.Launch()
-	s.Require().NoError(err)
-	time.Sleep(30 * time.Second)
+	s.StartCollector(false)
+
+	time.Sleep(60 * time.Second)
```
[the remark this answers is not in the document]
Stringy: This was me trying to nail down the process timing problem. Turns out even if you wait for five minutes between collector starting and starting the plop container, we still miss the process. I'll revert for now, and this is likely to change again after further investigations

integration-tests/suites/common/collector_manager.go
```diff
@@ -172,23 +147,8 @@ func (c *CollectorManager) getAllContainers() (string, error) {
...
-		"--name", "grpc-server",
-		"--network=host",
-		"--privileged",
-		"-v", "/tmp:/tmp:rw",
-		"--user", user.Uid + ":" + user.Gid,
-		c.GRPCServerImage,
-	}
-	_, err := c.executor.Exec(cmd...)
-	return err
+	// c.Sensor.Start()
```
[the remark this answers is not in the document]
Stringy: Yeah there's a lot of grpc stuff still in here, but it's not needed so I'll get rid of it

integration-tests/suites/mock_sensor/server.go
Stringy: I originally avoided the log library because I didn't want to mess with logging elsewhere, but obviously Logger structs exist and I am idiot 🤣 I've implemented this locally now and it also gives us timestamps on the events which could make debugging of timing problems a lot easier

integration-tests/suites/base.go
```diff
@@ -57,6 +60,55 @@ type PerformanceResult struct {
...
+	if !disableGRPC {
+		s.Sensor().Start()
+	}
+
+	s.Require().NoError(s.Collector().Setup())
+	return s.Collector().Launch()
+}
+
+// StopCollector will tear down the collector container and stop
+// the MockSensor if it was started.
```
JoukoVirtanen, line 75: I don't think this stops the mock sensor.

JoukoVirtanen:
I tried running this locally and got this error again 

```
Retrying ([docker pull quay.io/rhacs-eng/collector-performance:stats-]) (1 of 3) Error: Command Failed: docker pull quay.io/rhacs-eng/collector-performance:stats-
```

Can you rebase to fix this issue?

integration-tests/suites/duplicate_endpoints.go
```diff
@@ -61,27 +40,23 @@ func (s *DuplicateEndpointsTestSuite) killSocatProcess(port int) {
...
 	time.Sleep(30 * time.Second)
 
 	processImage := config.Images().QaImageByKey("qa-socat")
 
 	containerID, err := s.launchContainer("socat", processImage, "TCP-LISTEN:80,fork", "STDOUT")
-
 	s.Require().NoError(err)
+
 	s.serverContainer = common.ContainerShortID(containerID)
-	s.waitForEndpoints()
```
JoukoVirtanen, line 84: This can't be deleted. Read the comment for SetupSuite. The times of the scrapes have to be known and the times at which the listening endpoints are opened and closed have to be coordinated with the scrapes. Otherwise you will have flakes.

integration-tests/suites/duplicate_endpoints.go
```diff
@@ -93,27 +68,16 @@ func (s *DuplicateEndpointsTestSuite) SetupSuite() {
...
 }
 
 func (s *DuplicateEndpointsTestSuite) TestDuplicateEndpoints() {
-	processes, err := s.GetProcesses(s.serverContainer)
-	s.Require().NoError(err)
-	endpoints, err := s.GetEndpoints(s.serverContainer)
-	s.Require().NoError(err)
-
-	assert.Equal(s.T(), 2, len(endpoints))
-	assert.Equal(s.T(), 10, len(processes))
```
JoukoVirtanen, line 118: It is a good idea to still check the number of expected processes.

Stringy: Agreed. See this thread for the current issue affecting process reporting in some of these tests: https://redhat-internal.slack.com/archives/CFMQ5C2TT/p1694699857526889

integration-tests/suites/duplicate_endpoints.go
Stringy: The problem is the scrape timings are impossible to predict at this level. The current test does not strictly abide by the timings outlined in the comment, and `waitForEndpoints` does not account for the time required to download the database. I think this test needs a bit of a rethink, especially with the removal of the database.

Molter73 (commented):
Just a few more comments, but I think we'll be able to merge this soon-ish.

integration-tests/integration_test.go
```diff
@@ -40,7 +40,8 @@ func TestRepeatedNetworkFlow(t *testing.T) {
 		NumIter:                11,
 		SleepBetweenCurlTime:   2,
 		SleepBetweenIterations: 1,
-		ExpectedReports:        []bool{true, false},
```
JoukoVirtanen, line 43: I think it is good to check the order. When first working on these tests, I did have some more complicated tests for this. Perhaps in these three the order is not as important.

integration-tests/integration_test.go
```diff
@@ -98,7 +101,7 @@ func TestProcfsScraperNoScrape(t *testing.T) {
 
 func TestProcfsScraperDisableFeatureFlag(t *testing.T) {
 	connScraperTestSuite := &suites.ProcfsScraperTestSuite{
-		TurnOffScrape:               false,
+		TurnOffScrape:               true,
```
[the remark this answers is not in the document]
JoukoVirtanen: `DisableFeatureFlag` refers to `RoxProcessesListeningOnPort` being false. I don't think we should change the value here. I don't see a reason for it. You can make another test with TurnOffScrape=true and RoxProcessesListeningOnPort=false if you like, but this test should be kept. The name of the test could be changed to something like `ScrapeWithProcessesListeningOnPort`.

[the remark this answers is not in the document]
JoukoVirtanen: Running this locally the test now fails with TurnOffScrape=false and RoxProcessesListeningOnPort=false. The reason is that previously if the process originator was nil the endpoint would also be nil, but now with the code changes, if the originator is nil the endpoint is not nil, so expecting that if RoxProcessesListeningOnPort=false that the endpoint is nil is not correct. This can be easily fixed and the test should pass.

integration-tests/suites/duplicate_endpoints.go
JoukoVirtanen: If the comments are not correct and you correct them? I think the line `s.Sensor().ExpectEndpointsN(s.T(), s.serverContainer, 60*time.Second, 1)` should give the same functionality as before.  This test has a longer scrape interval than the other tests to mitigate the problems that you outlined. The major problem that you outlined, is with a problem with the old method. What problems do you think there are with the test as it currently is?

integration-tests/suites/duplicate_endpoints.go
```diff
@@ -93,27 +65,17 @@ func (s *DuplicateEndpointsTestSuite) SetupSuite() {
...
+	s.Sensor().ExpectEndpointsN(s.T(), s.serverContainer, 10*time.Second, 2)
+	s.Sensor().ExpectProcessesN(s.T(), s.serverContainer, 10*time.Second, 8)
```
JoukoVirtanen, line 80: The sleep here should be 20 seconds, like it was in the original test. We need to make sure that at the end of the scrape interval there are no extra endpoints reported. Also we should simply assert that after 20 seconds we have the endpoints and processes that we expect. We should not end the test early if we get the endpoints that we expect. There might be additional unexpected endpoints reported. The point of this test was that we were getting duplicate endpoints reported and this test was supposed to catch that bug. I think the way you have this now might not catch the bug this test was intended to catch.

integration-tests/suites/mock_sensor/expect_conn.go
```diff
@@ -0,0 +1,169 @@
...
+		return true
+	}
+
+	timer := time.After(timeout)
+
+loop:
+	for {
+		select {
+		case <-timer:
+			return assert.Fail(t, "timed out waiting for networks")
```
JoukoVirtanen, line 105: Should be timed out waiting for endpoints

integration-tests/suites/missing_proc_scrape.go
```diff
@@ -21,23 +19,15 @@ func (s *MissingProcScrapeTestSuite) SetupSuite() {
 		s.Require().NoError(err, "Failed to create fake proc directory")
 	}
 
-	s.executor = common.NewExecutor()
-	s.collector = common.NewCollectorManager(s.executor, s.T().Name())
-
-	s.collector.Mounts["/host/proc:ro"] = fakeProcDir
+	// Mount the fake proc directory created by 'create-fake-proc.sh'
```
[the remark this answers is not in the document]
Stringy: I think the comment is out-dated - I'll remove it for now

integration-tests/suites/duplicate_endpoints.go
Stringy: To be clear, this is not a sleep, but a worst-case limit to wait for additional endpoints/processes. The original 20 second sleep remains in the SetupSuite of this test. 

I'll add some additional asserts here to verify that no more processes or endpoints are reported than the ones we expect.

integration-tests/suites/duplicate_endpoints.go
Stringy: I think there has been a misunderstanding on my part on how this test works and the way in which collector reports endpoints. I'll update the docs and the general flow of this test to better reflect the behaviour under test.

integration-tests/suites/duplicate_endpoints.go
```diff
@@ -47,73 +30,79 @@ func (s *DuplicateEndpointsTestSuite) killSocatProcess(port int) {
...
+// therefore this test is strictly concerned with the scrape interval.
+//
+// The desired time line of events in this test is the following:
+//
+// for time (t), and scrape interval (i)
+//
+// 1. Start a process that opens a port, 80
+// 2. Wait for the endpoint to be reported, this is t=0
+// 3. Start a process that opens a different port, 81
+// 4. At t=i (the scape interval), port 81 should be reported
```
JoukoVirtanen, line 64:
```suggestion
// 4. At t=i (the scrape interval), port 81 should be reported
```

integration-tests/suites/duplicate_endpoints.go
JoukoVirtanen: Looks good.

integration-tests/suites/base.go
```diff
@@ -425,27 +361,61 @@ func (s *IntegrationTestSuiteBase) StartContainerStats() {
...
 }
+
+// isSelinuxPermissiveNeeded returns whether or not a given VM requires
+// SELinux permissive mode. e.g. rhel, or fedora-coreos
+func (s *IntegrationTestSuiteBase) isSelinuxPermissiveNeeded() bool {
+	vmType := config.VMInfo().Config
+	if strings.Contains(vmType, "coreos") || strings.Contains(vmType, "rhcos") {
+		return true
+	}
+	if strings.Contains(vmType, "rhel-7") {
```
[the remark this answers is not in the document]
Stringy: We could probably remove it now (?)

integration-tests/integration_test.go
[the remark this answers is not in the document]
Stringy: Fixed

integration-tests/suites/base.go
```diff
@@ -425,27 +361,61 @@ func (s *IntegrationTestSuiteBase) StartContainerStats() {
...
-	maxCount := 10
-
-	output, _ := s.executor.Exec("stat", file, "2>&1")
-	fmt.Println(output)
-	for !strings.Contains(output, "No such file or directory") {
-		time.Sleep(1 * time.Second)
-		count += 1
-		if count == maxCount {
+	timer := time.After(10 * time.Second)
+	ticker := time.NewTicker(time.Second)
```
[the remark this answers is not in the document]
Stringy: Agreed. I think we should rethink the communications between tests and test containers now that we're running on the same host and will (soon) be running in a container. For now though, I think we can leave this as-is

integration-tests/suites/types/network.go
```diff
@@ -0,0 +1,14 @@
...
+	LocalAddress   string
+	RemoteAddress  string
+	Role           string
+	SocketFamily   string
+	CloseTimestamp string
+}
+
+func (n *NetworkInfo) IsActive() bool {
+	// no close timestamp means the connection is open, and active
+	return n.CloseTimestamp == "(timestamp: nil Timestamp)"
```
Molter73, line 13: There's a `nilTimeStamp` constant in `integration-tests/suites/base.go`, could somehow move it into this file and test against that directly?

Molter73 (approved):
LGTM!

JoukoVirtanen (approved):
LGTM

PR #1318, opened 2023-09-11
Fixes integration tests on noexec mounted directories

Stringy:
## Description

With the tests now running directly on the host, some hosts mount `/tmp` `noexec` which means we get permission denied errors when trying to run the integration tests from that location. This PR changes the location to the user's home directory to avoid the `noexec` constraint. 

Longer term, tests should be containerized to avoid this issue altogether

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Tested locally against garden linux, will rely on CI to test the remainder of the platforms.

### DIFF
--- a/ansible/roles/run-test-target/vars/main.yml
+++ b/ansible/roles/run-test-target/vars/main.yml
@@ -1,6 +1,6 @@
 ---
 
 integration_tests_root: "{{ collector_root }}/integration-tests"
-remote_tests_root: /tmp/collector-tests/
+remote_tests_root: "{{ ansible_env.HOME }}/collector-tests/"
 remote_tests_binary: "{{ remote_tests_root }}/collector-tests"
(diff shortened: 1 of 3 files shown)


ansible/roles/run-test-target/vars/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 
 integration_tests_root: "{{ collector_root }}/integration-tests"
-remote_tests_root: /tmp/collector-tests/
+remote_tests_root: "{{ ansible_env.HOME }}/collector-tests/"
```
Molter73, line 4: [ultra-nit] Think you can remove the trailing `/`, it's causing the tests to look for `collector-tests//collector-tests`

Molter73 (approved):

PR #1325, opened 2023-09-12
ROX-19605: new feature-flag to control aggregation of unmatched IPs

ovalenti:
## Description

A new feature-flag ROX_ENABLE_EXTERNAL_IPS environment variable controls whether external network connections which do not match any defined known network are aggregated in the default bucket.

The default value is false, which corresponds to the previous behavior. 

## Checklist

**Automated testing**
  - [x] Added unit tests

### DIFF
--- a/collector/lib/ConnTracker.cpp
+++ b/collector/lib/ConnTracker.cpp
@@ -91,15 +91,19 @@ IPNet ConnectionTracker::NormalizeAddressNoLock(const Address& address) const {
     return network;
   }
 
-  // Otherwise, associate it to "rest of the internet".
-  switch (address.family()) {
-    case Address::Family::IPV4:
-      return IPNet(canonical_external_ipv4_addr, 0, true);
-    case Address::Family::IPV6:
-      return IPNet(canonical_external_ipv6_addr, 0, true);
-    default:
-      return {};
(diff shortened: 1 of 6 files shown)


Molter73 (approved):
LGTM!

ovalenti:
Since the other parts of the system are not going to play nice with only this change in collector, it looks like we will need a feature-flag that covers several places. I renamed it and also reverted the logic.

Could you please take an other look and see if everything's fine, @Molter73 ?

collector/lib/ConnTracker.cpp
```diff
@@ -91,7 +91,11 @@ IPNet ConnectionTracker::NormalizeAddressNoLock(const Address& address) const {
     return network;
   }
 
-  // Otherwise, associate it to "rest of the internet".
+  if (enable_external_ips_) {
+    return IPNet(address, 0, true);
+  }
+
+  // associate it to "rest of the internet".
```
Molter73, line 98: [ultimate-nit] We can now leave the comment as it was.

Molter73 (approved):
I actually like this approach better, LGTM!

Molter73:
On a separate note, mind updating the PR description? It still names the old env variable.

PR #1326, opened 2023-09-12
Use the effective URL for logging when downloading a driver

Molter73:
## Description

Loosely related to ROX-19614, when sensor and collector are deployed to a namespace other than stackrox, the URL logged by collector points to it trying to download the driver from sensor.stackrox.svc when it should be using sensor.<namespace>.svc.

This change adds a GetEffectiveURL method to the FileDownloader class, so we can now print the correct URL the driver will be downloaded from.

## Checklist

**Automated testing**
  - [x] Added unit tests

## Testing Performed

Deployed the image to a non standard namespace and checked the collector logs:
```
[INFO    2023/09/12 10:38:20] Sensor configured at address: sensor.rhacs-operator.svc:443
[INFO    2023/09/12 10:38:20] Attempting to connect to Sensor
[INFO    2023/09/12 10:38:20] Successfully connected to Sensor.
[INFO    2023/09/12 10:38:20] Module version: 2.6.0
[INFO    2023/09/12 10:38:20] Config: collection_method:ebpf, scrape_interval:30, turn_off_scrape:0, hostname:kind-control-plane, processesListeningOnPorts:1, logLevel:INFO, set_import_users:0, collect_connection_status:1
[INFO    2023/09/12 10:38:20] Candidate drivers:
[INFO    2023/09/12 10:38:20] collector-ebpf-6.4.13-200.fc38.x86_64.o
[INFO    2023/09/12 10:38:20] Attempting to download collector-ebpf-6.4.13-200.fc38.x86_64.o
[INFO    2023/09/12 10:38:20] Attempting to download kernel object from https://sensor.rhacs-operator.svc:443/kernel-objects/2.6.0/collector-ebpf-6.4.13-200.fc38.x86_64.o.gz
```

### DIFF
--- a/collector/lib/FileDownloader.cpp
+++ b/collector/lib/FileDownloader.cpp
@@ -4,6 +4,7 @@
 #include <algorithm>
 #include <chrono>
 #include <fstream>
+#include <optional>
 #include <string_view>
 #include <unistd.h>
 #include <utils.h>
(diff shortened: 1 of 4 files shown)


ovalenti (commented):
Since we are providing the FileDownloader with the connect_to info, I think it would be simpler to store the bits of information and build the connect_to using those internally, instead of having to parse the string.

collector/lib/FileDownloader.cpp
```diff
@@ -110,16 +140,25 @@ FileDownloader::~FileDownloader() {
...
   }
 
+  if (url_ != nullptr) {
+    curl_url_cleanup(url_);
+  }
+
   curl_global_cleanup();
 
-  curl_slist_free_all(connect_to_);
+  connect_to_.reset();
```
ovalenti, line 149: [ultra-nit] This seems superfluous

ovalenti (approved):
Thanks for making the changes suggested :+1: 

That looks good !

collector/lib/FileDownloader.cpp
Molter73: You are right, I'll remove it.

PR #1328, opened 2023-09-13
Improves QA tag processing to better support local development

Stringy:
## Description

Fixes issues where the QA tag was not set correctly for local testing.

## Checklist

## Testing Performed

Tested locally, and verified that images are pulled correctly.

### DIFF
--- a/integration-tests/suites/config/images.go
+++ b/integration-tests/suites/config/images.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
 	"io/ioutil"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


integration-tests/suites/config/images.go
```diff
@@ -62,7 +63,16 @@ func loadImageStore(location string) (*ImageStore, error) {
 // tag as is otherwise.
 func getQATag(base_tag string) string {
 	if collectorQATag == "" {
```
Molter73, line 65: [ultra-nit] Do we need lazy loading of this variable? If we only read it once or twice, can't we just read it always and make the function slightly simpler? 

Or maybe since it's used as part of the `ImageStore` struct we could create a `func (i *ImageStore) New()` that will initialize this value internally? I could even argue the existing `loadImageStore` function could be easily refactored into `New`.

JoukoVirtanen (approved):
I tested this and it works. You can merge once you get approval from Mauro.

integration-tests/suites/config/images.go
Stringy: Yeah it's read a lot over the course of a test run so it's worth caching the result, but your point about `loadImageStore` is good so I'll move it there

Molter73 (approved):
LGTM!

PR #1333, opened 2023-09-16
Jv build all integration test containers

JoukoVirtanen:
## Description

The integration tests depend on several images. Each of these images have their own directory in integration-tests/container. Each of those directories has a Makefile that can be used to build and push the images. CI builds and pushes these images using the Makefiles. To make this more convenient when running the integration tests locally a Makefile is added to the integration-tests/container to build and push all of the images, so that the Makefile targets don't have to be run individually for each directory.

All of the Makefiles are very similar except for perf. Seems like this is a violation of DRY and they should be consolidated into the Makefile added in this PR.

## Checklist

**Automated testing**
~~ - [ ] Added unit tests ~~
~~ - [ ] Added integration tests ~~
~~ - [ ] Added regression tests ~~

This just adds a Makefile so no tests are needed for it.

## Testing Performed

`make build-all`

### DIFF
--- a/integration-tests/container/Makefile
+++ b/integration-tests/container/Makefile
@@ -1,0 +1,20 @@
+BASE_PATH = .
+include Makefile-constants.mk
+
+SHELL := /bin/bash
+
+.PHONY: build-all
+build-all:
+	@set -eou pipefail; \
+	mapfile -t container_dirs < <(ls -d */); \
+	for container_dir in "$${container_dirs[@]}"; do \
+		make -C "$$container_dir" build; \
+	done
+
+.PHONY: build-and-push-all
+build-and-push-all:
+	@set -eou pipefail; \
+	mapfile -t container_dirs < <(ls -d */); \
+	for container_dir in "$${container_dirs[@]}"; do \
+		make -C "$$container_dir" build-and-push; \
+	done
--- a/integration-tests/container/Makefile
+++ b/integration-tests/container/Makefile
@@ -3,18 +3,11 @@ include Makefile-constants.mk
 
 SHELL := /bin/bash
 
-.PHONY: build-all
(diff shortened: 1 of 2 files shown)


integration-tests/container/Makefile
```diff
@@ -0,0 +1,20 @@
+BASE_PATH = .
+include Makefile-constants.mk
+
+SHELL := /bin/bash
+
+.PHONY: build-all
+build-all:
+	@set -eou pipefail; \
+	mapfile -t container_dirs < <(ls -d */); \
```
Stringy, line 9: I think mapfile is linux specific, so this may break for those of us who are using MacOS. I think this can also be achieved in pure (gnu) make

```
DIRS := $(sort $(dirs $(wildcard ./*))
```

And possibly the following, though I've not tested it

```
TARGETS := build build-and-push
DIRS := $(sort $(dirs $(wildcard ./*))

$(TARGETS): $(DIRS)
$(DIRS):
    $MAKE -C $@ $(MAKECMDGOALS)

.PHONY: $(TARGETS) $(DIRS)
```

JoukoVirtanen: Your suggestion didn't work, but I changed it a bit and now it does.

integration-tests/container/README.md
```diff
@@ -23,6 +23,18 @@ The intention of this version number is to ensure consistency across master
 testing and release testing, whereby releases are pegged to a specific QA image
 version.
 
+To build all images run
+
+`make build-all`
```
Stringy, line 28: I think these targets are now `build` and `build-and-push` with the above makefile changes

Stringy (approved):
One minor documentation suggestion, but otherwise LGTM!

PR #1335, opened 2023-09-18
Remove dependency on curl's URL API

Molter73:
## Description

Curl's URL API was added on version 7.62, but RHEL 8 only support 7.61. We could've added libcurl to the builder as a dependency, but given we are still trying to sort builder versioning, the image already takes a pretty long time to build and we only really require a very small subset of the URL API, I opted for adding a small subclass to FileDownloader that will handle parsing of URLs for our use case. The URL API also returned C strings that were copies of what was stored internally and needed to be copied into proper std::strings and free'd, so this approach is a bit cleaner IMO and requires less memory manipulation.

## Checklist

## Testing Performed

The unit tests added by #1326 should provide coverage for our use case.

### DIFF
--- a/collector/lib/FileDownloader.cpp
+++ b/collector/lib/FileDownloader.cpp
@@ -141,8 +141,8 @@ FileDownloader::~FileDownloader() {
 }
 
 bool FileDownloader::SetURL(const std::string& url) {
-  std::string_view file_path{url};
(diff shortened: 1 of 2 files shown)


ovalenti (approved):
That should make the trick :fist_oncoming:

PR #1346, opened 2023-09-25
Remove all mentions of Sysdig and Falco in our codebase

Molter73:
## Description

This is a long due change, all mentions of Sysdig and Falco have been replaced by a more generic system_inspector name. No functional change has been made, the PR is purely organizational.

## Checklist

## Testing Performed

CI should be enough, since things are just being shuffled around.

### DIFF
--- a/collector/lib/CMakeLists.txt
+++ b/collector/lib/CMakeLists.txt
@@ -1,43 +1,8 @@
-set(COLLECTOR_LIB_SRC_FILES
-    NetworkConnection.cpp
-    CollectorConfig.cpp
-    CollectorArgs.cpp
-    SignalServiceClient.cpp
-    RateLimit.cpp
-    Process.cpp
-    ProtoUtil.cpp
-    Logging.cpp
-    HostInfo.cpp
-    ProfilerHandler.cpp
-    GetStatus.cpp
-    FileDownloader.cpp
-    AbortHandler.cpp
-    ProcfsScraper.cpp
-    EventNames.cpp
-    CollectorService.cpp
-    NetworkSignalHandler.cpp
-    CollectionMethod.cpp
-    GRPCUtil.cpp
-    NetworkStatusNotifier.cpp
-    NRadix.cpp
-    GRPC.cpp
-    GetKernelObject.cpp
-    HostHeuristics.cpp
-    DriverCandidates.cpp
-    CollectorStatsExporter.cpp
-    SelfCheckHandler.cpp
-    CollectorStats.cpp
-    ConnTracker.cpp
-    StoppableThread.cpp
-    Utility.cpp
-    ProcessSignalHandler.cpp
-    ProcessSignalFormatter.cpp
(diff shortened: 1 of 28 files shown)


collector/test/SystemInspectorServiceTest.cpp
```diff
@@ -1,10 +1,9 @@
-#include "SysdigService.h"
-#include "gmock/gmock.h"
```
erthalion, line 2: Interesting, it wasn't used here?

docs/troubleshooting.md
```diff
@@ -315,10 +315,10 @@ Note that the `[syscall]` suffix in a metric name means that it is instanciated
 
 Note that if ProcfsScraper is unable to open /proc it is not able to open any of the subdirectories, but only procfs_could_not_open_proc_dir will be incremented in that case.
 
-### Falco timers per syscall
+### system_inspector timers per syscall
```
erthalion, line 318: It's probably worth mentioning somewhere before this point, that system_inspector in the current incarnation is actually nothing more than Falco inspector.

collector/lib/CollectorService.h
```diff
@@ -7,11 +7,11 @@
 #include "CollectorStats.h"
 #include "Control.h"
 #include "DriverCandidates.h"
-#include "SysdigService.h"
+#include "system-inspector/Service.h"
 
 namespace collector {
 
-class SysdigService;
+class Service;
```
erthalion, line 14: A nitpick, but can this be referred to by the fully qualified name, with the namespace? Otherwise the name "Service" is quite generic, and if it could be consistently used everywhere with the "system_inspector" (as everywhere else in this PR), that would be great.

collector/lib/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
+set(COLLECTOR_LIB_SRC_FILES
```
erthalion, line 1: Are you sure about switching from `*.cpp` to the exact list of files? Might be coming as a surprise for unprepared developers.

erthalion (commented):
Thanks, looks good and makes sense. Few comment below.

collector/test/SystemInspectorServiceTest.cpp
Molter73: It wasn't, the tests use full fledged Falco objects instead of mocks, they just doesn't instruct the inspector to start collecting data.

docs/troubleshooting.md
Molter73: Makes sense, I'll find a place for it.

collector/lib/CMakeLists.txt
Molter73: I just prefer the explicit way of listing source files with cmake, (I believe this is also the recommended way to do things, since globbing `*.cpp` can cause things like new files not being added to already configured builds), but you are probably right that this is not the place to make this change, I'll roll it back and we can circle back to it at a later point.

erthalion (approved):

PR #1348, opened 2023-09-29
ROX-19898: refactor collector configuration in tests

Stringy:
## Description

To avoid having to provide (1) a hard-coded json string and (2) duplicated default options across all tests, this PR introduces `CollectorStartupOptions` structure which allows tests to set specific options/environment variables/mount points before starting collector. The collector manager itself is responsible for serializing this information as necessary and tests only need to modify the options that they need to.

Options provided by a test _always_ supersede those defined by default in the collector manager.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Tested and built locally against Garden Linux (of all things) CI will catch everything else. This should have no impact on the outcome of the tests.

### DIFF
--- a/integration-tests/suites/base.go
+++ b/integration-tests/suites/base.go
@@ -60,13 +60,13 @@ type PerformanceResult struct {
 
 // StartCollector will start the collector container and optionally
 // start the MockSensor, if disableGRPC is false.
-func (s *IntegrationTestSuiteBase) StartCollector(disableGRPC bool) {
+func (s *IntegrationTestSuiteBase) StartCollector(disableGRPC bool, options *common.CollectorStartupOptions) {
 	if !disableGRPC {
 		s.Sensor().Start()
 	}
 
 	s.Require().NoError(s.SetSelinuxPermissiveIfNeeded())
-	s.Require().NoError(s.Collector().Setup())
+	s.Require().NoError(s.Collector().Setup(options))
 	s.Require().NoError(s.Collector().Launch())
 
 	// wait for self-check process to guarantee collector is started
--- a/integration-tests/suites/common/collector_manager.go
+++ b/integration-tests/suites/common/collector_manager.go
@@ -1,48 +1,63 @@
 package common
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"golang.org/x/exp/maps"
+
 	"github.com/hashicorp/go-multierror"
 	"github.com/stackrox/collector/integration-tests/suites/config"
 )
 
+type CollectorStartupOptions struct {
+	Mounts        map[string]string
+	Env           map[string]string
+	Config        map[string]any
+	BootstrapOnly bool
+}
+
 type CollectorManager struct {
-	executor        Executor
-	Mounts          map[string]string
-	Env             map[string]string
+	executor      Executor
+	mounts        map[string]string
+	env           map[string]string
+	config        map[string]any
+	bootstrapOnly bool
+	coreDumpPath  string
+	testName      string
+
 	CollectorOutput string
-	CollectorImage  string
-	BootstrapOnly   bool
-	TestName        string
-	CoreDumpFile    string
-	VmConfig        string
 	ContainerID     string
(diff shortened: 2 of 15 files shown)


integration-tests/suites/base.go
```diff
@@ -390,24 +390,13 @@ func (s *IntegrationTestSuiteBase) waitForFileToBeDeleted(file string) error {
...
 	}
-
-	return nil
 }
 
 // isSelinuxPermissiveNeeded returns whether or not a given VM requires
 // SELinux permissive mode. e.g. rhel, or fedora-coreos
 func (s *IntegrationTestSuiteBase) isSelinuxPermissiveNeeded() bool {
 	vmType := config.VMInfo().Config
-	if strings.Contains(vmType, "coreos") || strings.Contains(vmType, "rhcos") {
```
Stringy, line 401: Not sure how this change made it in, I think I hadn't pushed the base branch when I created this branch. I'll juggle commits at some point, or we can just merge this change as part of this PR (I don't think it strictly affects either of them)

integration-tests/suites/repeated_network_flow.go
```diff
@@ -36,11 +37,17 @@ type RepeatedNetworkFlowTestSuite struct {
 func (s *RepeatedNetworkFlowTestSuite) SetupSuite() {
 	s.StartContainerStats()
 
-	s.Collector().Env["COLLECTOR_CONFIG"] = `{"logLevel":"debug","turnOffScrape":true,"scrapeInterval":` + strconv.Itoa(s.ScrapeInterval) + `}`
-	s.Collector().Env["ROX_AFTERGLOW_PERIOD"] = strconv.Itoa(s.AfterglowPeriod)
-	s.Collector().Env["ROX_ENABLE_AFTERGLOW"] = strconv.FormatBool(s.EnableAfterglow)
+	collectorOptions := common.CollectorStartupOptions{
+		Config: map[string]any{
+			"scrapeInterval": s.ScrapeInterval,
```
Stringy, line 42: @JoukoVirtanen is this correct? it seems in this test we're specifically setting a scrape interval, but always turn scraping off.

integration-tests/suites/common/collector_manager.go
```diff
@@ -138,25 +168,32 @@ func (c *CollectorManager) launchCollector() error {
...
+
+	for k, v := range c.env {
 		cmd = append(cmd, "--env", k+"="+v)
 	}
 
-	cmd = append(cmd, c.CollectorImage)
+	configJson, err := json.Marshal(c.config)
+	if err != nil {
+		return err
+	}
```
Molter73, line 191: I'd pay for go to implement something like the `?` from rust so we could do `configJson := json.Marshal(c.config)?` and be done...

integration-tests/suites/base.go
Molter73: I think it's fine, we can merge it as part of this PR.

Molter73 (approved):
This is great! I think the new way of organizing configuration for collector is a lot cleaner to read and I love we are using proper JSON serialization (although I'd prefer we didn't have that JSON env variable, but that will be dealt with at some other point in time).

LGTM! But I'd still like to hear from @JoukoVirtanen on your comment.

integration-tests/suites/repeated_network_flow.go
Stringy: I've tested this locally and the test still passes with `turnScrapeOff=false`. I think this probably better represents what is intended in this test, but could do with some confirmation in case there is some nuance here that I'm missing

JoukoVirtanen: The scrapeInterval is also the reporting interval. Even if scrape is off the scrapeInterval will be used to determine when connections and endpoint are reported. It is a bit confusing and should be changed. In this case it shouldn't matter if scraping is done or not. In that case network connections will come from NetworkSignalHandler.

Stringy: Thanks! I'll turn off scraping again and fix the conflicts and hopefully we can merge this today

PR #1351, opened 2023-10-03
Makes expect logging clearer when more events are reveived than expected

Stringy:
## Description

Improves the logging output for clarity in scenarios where number of events received exceeds the expected number.

Before:

`only retrieved 4 processes, expected 3` 

After

`found 4 processes (expected 3)`


Molter73 (approved):
Thanks for this! LGTM!!

PR #1352, opened 2023-10-03
Clean up of perf results accesses and storage

Stringy:
## Description

Removes some boilerplate from the tests around the printing and storage of various perf results after the tests have executed. Now, perf results are all handled within the integration tests base struct, and the test simply has to call `WritePerfResults` in `TearDownSuite`

This also fixes an issue where container stats weren't initialized properly, resulting in invalid perf reporting. 

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Tested locally against RHEL 8 and RHEL 9

### DIFF
--- a/integration-tests/scripts/baseline/main.py
+++ b/integration-tests/scripts/baseline/main.py
@@ -150,8 +150,8 @@ def verify_data(data):
         # NOTE: We rely only on baseline/collector hackbench at the moment. As
         # soon as data for more tests will be collected, this section has to be
         # extended accordingly.
-        no_overhead_count = counter.get("baseline_benchmark", 0)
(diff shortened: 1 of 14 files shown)


ovalenti (approved):
The code looks good to me ! I just noticed that the perf comment in the PR is empty. This is not related, right ?

Stringy:
>  I just noticed that the perf comment in the PR is empty. This is not related, right ?

No, I think this bug has existed for a little while, though it might be affected by the change in names in the perf result (`benchmark` -> `TestBenchmarkCollector`) I'll investigate before merging

PR #1358, opened 2023-10-06
Implement builder versioning

Molter73:
## Description

Versioning for the builder has been changed so that tags and release branches build their own versions without conflicting with master and PRs. The builder has also undergone a slight refactor so instead of using the entire repository as its context, just the builder directory is used.

We also removed luajit from the list of dependencies, this hasn't been needed since the removal of chisels on #1290.

Supersedes #1315.

## Checklist

## Testing Performed

- [x] Check collector is properly built with the changes to the builder on all archs. [CI run](https://github.com/stackrox/collector/actions/runs/6430399727?pr=1358)
- [x] Run the changes on a test tag. [CI run](https://github.com/stackrox/collector/actions/runs/6455211309)
- [x] Run the changes on a test release- branch. [CI run](https://github.com/stackrox/collector/actions/runs/6455167343)

### DIFF
--- a/.github/workflows/collector-builder.yml
+++ b/.github/workflows/collector-builder.yml
@@ -15,7 +15,7 @@ on:
 
 env:
   COLLECTOR_TAG: ${{ inputs.collector-tag }}
-  DEFAULT_BUILDER_TAG: cache
+  DEFAULT_BUILDER_TAG: master
 
 jobs:
   builder-needs-rebuilding:
@@ -32,8 +32,9 @@ jobs:
         with:
           filters: |
             builder-changed:
-              - builder/**
-              - third_party/**
+              - builder/install/**
+              - builder/third_party/**
+              - builder/Dockerfile
 
   build-builder-image:
     name: Build the builder image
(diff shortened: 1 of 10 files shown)


.github/workflows/collector-builder.yml
```diff
@@ -71,7 +74,9 @@ jobs:
         id: builder-tag
         run: |
           COLLECTOR_BUILDER_TAG="${DEFAULT_BUILDER_TAG}"
-          if [[ "${{ github.event_name }}" == 'pull_request' ]]; then
+          if [[ "${{ github.event_name }}" == 'pull_request' || \
+                "${{ github.ref_type }}" == 'tag' || \
+                "${{ github.ref_name }}" =~ ^release- ]]; then
```
erthalion, line 79: Just out of curiosity, what would happen for a commit with both a tag and a release branch, it will build the same image twice?

Molter73: It will, but they will be tagged as different images. That was the case for these two runs:
- Tag: https://github.com/stackrox/collector/actions/runs/6455211309
- Release branch: https://github.com/stackrox/collector/actions/runs/6455167343

erthalion: Yeah, I see, at the end it boils down to ref_name vs make tag.

erthalion (approved):

PR #1359, opened 2023-10-06
Build ARM64 images

Molter73:
## Description

[GCP added ARM VMs last year](https://cloud.google.com/blog/products/compute/tau-t2a-is-first-compute-engine-vm-on-an-arm-chip). With the effort to support Power and Z architectures that has been ongoing for about the past year as well, it is relatively easy to enable collector builds for ARM and add integration tests for it using CO-RE BPF thanks to Falco supporting the architecture already. So, this PR adds ARM image builds for collector and tests them on GCP.

## Checklist

## Testing Performed
CI should be enough.


.github/workflows/collector-full.yml
```diff
@@ -138,15 +138,64 @@ jobs:
...
+      !contains(github.event.pull_request.labels.*.name, 'skip-multiarch-builds')
+    env:
+      COLLECTOR_IMAGE: quay.io/stackrox-io/collector:${{ inputs.collector-tag }}
+      RHACS_ENG_IMAGE: quay.io/rhacs-eng/collector:${{ inputs.collector-tag }}
+    steps:
+      - name: Pull image to retag
+        run: |
+          docker pull ${{ env.COLLECTOR_IMAGE }}-arm64-slim
+
+      - name: Retag and push stackrox-io -slim
```
robbycochran, line 157: The tags for each section `name` are not consistent between stackrox-io and rhacs-eng.

robbycochran:
Nice! Tests passing on arm64!

.github/workflows/collector-full.yml
```diff
@@ -24,20 +24,37 @@ on:
...
+      - name: archs string to JSON array
+        id: archs-json
+        shell: python
+        run: |
+          import json
+          import os
+
+          archs="${{ inputs.archs }}".split()
+          with open(os.environ['GITHUB_OUTPUT'], 'a') as f:
+            f.write(f'archs-json={json.dumps(archs)}\n')
```
Stringy, line 48: Compared to the bash equivalent, this is great!

.github/workflows/integration-tests.yml
```diff
@@ -63,6 +63,27 @@ jobs:
       job-tag: ${{ inputs.job-tag }}
     secrets: inherit
 
+  arm-integration-tests:
```
Stringy, line 66: it might be worth renaming `all-integration-tests` and `required-integration-tests` to something like `amd64-required-tests` and `amd64-additional-tests` to be consistent with this. I think it'll make things a lot clearer, especially with s390x and ppc64le added in for good measure (I'd advocate for renaming the multiarch test job as well, though there'll be more duplicated boilerplate for this, given there's only a single vm_type for each)

.github/workflows/integration-tests.yml
```diff
@@ -63,6 +63,27 @@ jobs:
...
+          - fcarm
+
+    with:
+      vm_type: ${{ matrix.vm_type }}
+      collector-tag: ${{ inputs.collector-tag }}
+      collector-qa-tag: ${{ inputs.collector-qa-tag }}
+      job-tag: ${{ inputs.job-tag }}
+    secrets: inherit
+
   notify:
```
Stringy, line 87: I think we'll need to add arm to the needs for this

.github/workflows/collector-full.yml
```diff
@@ -221,50 +237,3 @@ jobs:
           dst-image: quay.io/rhacs-eng/collector:${{ inputs.collector-tag }}-latest
           username: ${{ secrets.QUAY_RHACS_ENG_RW_USERNAME }}
           password: ${{ secrets.QUAY_RHACS_ENG_RW_PASSWORD }}
-
-  retag-collector-slim:
```
Stringy, line 225: I long for a day where retagging is unneeded (in a post-ebpf world?) because I find it painfully difficult to visualise. Moving this to ansible is a great step though, and simplifies things nicely.

Stringy (approved):
LGTM! This is excellent, and demonstrates all the work we put into multi-arch was worth the effort!

.github/workflows/integration-tests.yml
Molter73: Makes sense, I'll rename the tests.

.github/workflows/integration-tests.yml
Molter73: Yeah, I missed this one, great catch!

.github/workflows/collector-full.yml
Molter73: You and me both buddy, we have way too many edge cases attached to the full collector image. For instance, I just realized master fails to build s390x and ppc64le on master when we bump the driver version because the drivers have not been built downstream and pushed back up for us to use.

PR #1362, opened 2023-10-09
Fixes test cleanup, to ensure cleanup regardless of exit reason

Stringy:
## Description

The initial attempt at cleaning up after panics only covered a single possible exit condition for the tests (panicking). If a test/setup/teardown instead calls `assert.FailNow` it will ultimately call `runtime.Goexit` which according to the docs - "recover() calls will return nil"

The new mechanism will register a function with `testing.T.Cleanup` which will run when the tests and subtests end for any reason (panic, fails, and success.)

The success possibility is taken into account - we cannot fail or panic if resources that are being cleaned up no longer exist.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Tested locally by forcing a FailNow and a panic and observing that the cleanup function is called correctly. Also verified that the clean up works correctly when the tests pass.

### DIFF
--- a/integration-tests/suites/async_connections.go
+++ b/integration-tests/suites/async_connections.go
@@ -31,7 +31,7 @@ type AsyncConnectionTestSuite struct {
  *     a dummy address.
  */
 func (s *AsyncConnectionTestSuite) SetupSuite() {
-	defer s.RecoverSetup("server", "client")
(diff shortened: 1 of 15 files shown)


Molter73 (approved):
LGTM!

PR #1363, opened 2023-10-09
Add perf_event_open test

erthalion:
## Checklist

**Automated testing**
~- [ ] Added unit tests~
  - [x] Added integration tests

## Testing Performed


integration-tests/container/perf_event_open/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:jammy
+
+COPY sched_process_exit.c /sched_process_exit.c
+COPY entrypoint.sh /entrypoint.sh
+
+RUN apt update -y && apt install gcc -y
+RUN gcc sched_process_exit.c -o sched_process_exit
```
Molter73, line 7: [nit] Run commands should be grouped together to reduce the number of layers in the images. Of course, this is not strictly required in this image.
```suggestion
RUN apt update -y && apt install gcc -y && \
    gcc sched_process_exit.c -o sched_process_exit
```

integration-tests/container/perf_event_open/sched_process_exit.c
```diff
@@ -0,0 +1,62 @@
...
+  // How long to wait for events, in seconds
+  int wait_interval = 10;
+
+  // /sys/kernel/debug/tracing/events/sched/sched_process_exit/id
+  int tracepoint_code = 310;
+
+  if (argc == 3) {
+    wait_interval = atoi(argv[1]);
+    tracepoint_code = atoi(argv[2]);
+  }
```
Molter73, line 34: Looks like we are always using the arguments passed in, so we might want to fail if the number of arguments is wrong instead.
```suggestion
  if (argc != 3) {
    fprintf(stderr, "Expected 2 arguments, got: %d", argc - 1);
    exit(EXIT_FAILURE);
  }
  // How long to wait for events, in seconds
  int wait_interval = atoi(argv[1]);

  // /sys/kernel/debug/tracing/events/sched/sched_process_exit/id
  int tracepoint_code = atoi(argv[2]);

```

integration-tests/container/perf_event_open/sched_process_exit.c
```diff
@@ -0,0 +1,62 @@
...
+                int cpu, int group_fd, unsigned long flags) {
+  int ret;
+
+  ret = syscall(__NR_perf_event_open, hw_event, pid, cpu,
+                group_fd, flags);
+  return ret;
+}
+
+int main(int argc, char** argv) {
+  struct perf_event_attr pe;
```
Molter73, line 21: I see `pe` is initialized with a memset on line 36, we can probably omit that and initialize it here instead.
```suggestion
  struct perf_event_attr pe = {0};
```

Molter73: We could also initialize most of its fields, but I leave that up to you if you want to.

integration-tests/suites/base.go
```diff
@@ -234,7 +234,7 @@ func (s *IntegrationTestSuiteBase) GetLogLines(containerName string) []string {
 }
 
 func (s *IntegrationTestSuiteBase) launchContainer(name string, args ...string) (string, error) {
-	cmd := []string{common.RuntimeCommand, "run", "-d", "--name", name}
+	cmd := []string{common.RuntimeCommand, "run", "-d", "--name", name, "--privileged"}
```
Molter73, line 237: This will launch all containers as privileged, you probably want to add the flag specifically to the `perf-event-open` container via the `args` parameter.

Molter73: Either that or we want a separate `launchPrivilegedContainer` method that adds the flag.

kernel-modules/MODULE_VERSION
```diff
@@ -1 +1 @@
-2.6.0
+2.7.0-rc1
```
Molter73, line 1: Since we are pretty close to the release, we probably want to go directly to `2.7.0`

Molter73 (commented):
Thanks for putting this together @erthalion! I left a few comments but nothing major, think we can go ahead and merge stackrox/falcosecurity-libs#57 so we can directly update the falco submodule here, WDYT?

erthalion:
> Thanks for putting this together @erthalion! I left a few comments but nothing major, think we can go ahead and merge [stackrox/falcosecurity-libs#57](https://github.com/stackrox/falcosecurity-libs/pull/57) so we can directly update the falco submodule here, WDYT?

Yep, makes sense to me.

integration-tests/container/perf_event_open/Dockerfile
erthalion: I would say, logically related commands have to be grouped together (you probably don't want to mix together a package installation and compilation phases). But as you said, for this image it's not important.

integration-tests/suites/base.go
erthalion: Oh, this one is definitely not what I intended to do, probably a left over after some code shuffling.

Molter73 (approved):
LGTM!

PR #1364, opened 2023-10-11
Missing qa containers

Molter73:
## Description

While working on #1359, I found some containers that have been forcefully pushed to our registries with no clear way to rebuild them or add new architectures to them. These changes are not exclusive to the PR mentioned and I think it would be a good idea to add them to our repo so we can have CI take care of them in future work.

## Checklist

## Testing Performed

CI should be enough.


Stringy (approved):
LGTM! It'll be so nice to regain full understanding of our test image provenance!

PR #1365, opened 2023-10-11
Prevent QA tag collisions on PRs

Molter73:
## Description

When multiple PRs change the `QA_TAG` file to the same value, they can collide and cause images to be overwritten, leading to frustration and time wasted. This change makes it so PRs always use a unique tag for QA containers when needed, falling back to the value in `QA_TAG` and images built from the main branch otherwise.

## Checklist

## Testing Performed
- [ ] Ensure QA containers use a unique tag when running in a PR.
- [ ] Ensure running integration tests manually falls back to use the value in `QA_TAG`


.github/workflows/init.yml
```diff
@@ -6,6 +6,9 @@ on:
       collector-tag:
         description: The tag used when building collector
         value: ${{ jobs.common-variables.outputs.collector-tag }}
+      collector-qa-tag:
```
Stringy, line 9: I think this is currently unused (at least from this source) since it's set by the `integration-test-containers.yml` workflow 

We can probably use either source to setup the necessary qa tag for master vs. pull requests, but it's worth converging on a single source. I sort of think that `init.yml` is the right place for this kind of setting, but I'm not overly attached to that opinion so use whichever you think is best.

Stringy (approved):
One minor comment about where the QA tag is set, but overall LGTM! Thanks for putting this together, it will make working on the tests a lot easier

.github/workflows/init.yml
Molter73: I initially meant for this to be a backup in case the `integration-test-containers.yml` workflow was skipped, but I guess that changed at some point and forgot to unify.

If I move it to be here I'll have to also add a duplicated step checking if `QA_TAG` was modified, but that should be fine (I can actually check it here and set another output for the other step now that I think about it).

Stringy (approved):
LGTM!

PR #1366, opened 2023-10-13
Automate full image layer calculations

Stringy:
## Description

Sometimes CI fails because the number of drivers we have changes compared to the number of layers they are split into when building the full image. This PR automates that calculation so that only the right layers are used.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)

### DIFF
--- a/kernel-modules/container/Dockerfile
+++ b/kernel-modules/container/Dockerfile
@@ -4,7 +4,7 @@ ARG base_image=${collector_repo}:${collector_version}-slim
 
 FROM python:3 as kernel-modules
 ARG max_layer_size=xxx
-ARG max_layer_depth=xxx
+ARG required_layers=xxx
 
 # Increase the value in this check if additional hardcoded layers are added below.
 RUN test "${max_layer_depth}" -le 9
@@ -15,7 +15,7 @@ COPY partition-probes.py /
 WORKDIR /kernel-modules
 
 # Split probes into directories containing at most max_layer_size bytes.
-RUN /partition-probes.py "${max_layer_depth}" "${max_layer_size}" "." "/layers" && \
+RUN /partition-probes.py "${required_layers}" "${max_layer_size}" "." "/layers" && \
   for i in "/layers"/*; do mkdir -p "$(basename "$i")"; xargs -a "$i" mv -t "$(basename "$i")" ; done
 
 # Probes are split across multiple layers to limit blob size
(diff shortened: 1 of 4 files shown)


kernel-modules/container/Dockerfile
```diff
@@ -4,18 +4,20 @@ ARG base_image=${collector_repo}:${collector_version}-slim
 
 FROM python:3 as kernel-modules
 ARG max_layer_size=xxx
-ARG max_layer_depth=xxx
+ARG required_layers=xxx
 
 # Increase the value in this check if additional hardcoded layers are added below.
-RUN test "${max_layer_depth}" -le 9
+RUN test "${required_layers}" -le 9
```
Molter73, line 10: I don't really see the point in this check, I guess we were supposed to bring it back down when we dropped kernel-modules for it to make sense. I think we should either adjust the check so it warns when we go over the limit or just accept the number of layers can increase as a fact of life and get rid of the step altogether. Since we will be getting rid of more drivers in favor of CO-RE BPF, I'd get rid of the check.

kernel-modules/container/required-layers.py
```diff
@@ -0,0 +1,35 @@
...
+
+    buckets = 0
+    remaining_bucket_bytes = bytes_per_layer
+    for size in probe_sizes:
+        if remaining_bucket_bytes - size < 0:
+            buckets += 1
+            remaining_bucket_bytes = bytes_per_layer
+        remaining_bucket_bytes -= size
+
+    print(buckets)
```
Molter73, line 35: Could we move this to a `main()` method? Just to make it more clear where argument parsing ends and the main logic begins.

kernel-modules/container/required-layers.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python3
+
+import os
+import argparse
+import re
+
+probe_name_re = re.compile(r'^collector.*.gz$')
+
+
+def get_probe_sizes(probe_dir):
```
Molter73, line 10: [nit] Type hinting would be nice here, quickly glancing through the code I initially thought this method returned the added size of all probes instead of the individual sizes in an array, (I know I should've paid more attention, but still).
```suggestion
def get_probe_sizes(probe_dir: str) -> list:
```

Molter73 (commented):
You can probably skip multiarch builds, since the layer problem always arises on x86 and the applied logic is the same for all archs.

kernel-modules/container/Dockerfile
Stringy: I left it in as a last check before we start copying from layers, but I suppose that's now controlled by the chosen target (and would probably give us a more useful error anyway.) I'll remove it.

Molter73 (approved):
LGTM!

PR #1373, opened 2023-10-17
Skip multiarch builds as a default

Molter73:
## Description

Multiarch builds take a pretty long time which slows down dev work. Making them opt-in for PRs should make dev a lot faster even if we might get broken builds on master every now and then.

## Checklist

## Testing Performed

All tests should force the builder image to be built.
- [x] Run CI without `run-multiarch-builds` should skip P/Z. [CI run](https://github.com/stackrox/collector/actions/runs/6548585384)
- [x] Run CI with `run-multiarch-builds` should run P/Z. [CI run](https://github.com/stackrox/collector/actions/runs/6559044529?pr=1373)

### DIFF
--- a/.github/workflows/collector-builder.yml
+++ b/.github/workflows/collector-builder.yml
@@ -97,7 +97,7 @@ jobs:
         if: |
           github.event_name == 'push' ||
           matrix.arch == 'amd64' ||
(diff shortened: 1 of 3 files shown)


Stringy (approved):
LGTM!

PR #1374, opened 2023-10-18
Network introspection endpoints

ovalenti:
## Description

Adds a pair of web endpoints to retrieve the internal network tables of Collector (normalized and deduplicated, same as what is sent to Sensor).

The endpoints are activated by an environment variable (~~`ROX_COLLECTOR_ENABLE_INTROSPECTION`~~ `DEBUG_ENABLE_INTROSPECTION`). The default disabled.

### DIFF
--- a/collector/lib/NetworkStatusInspector.cpp
+++ b/collector/lib/NetworkStatusInspector.cpp
@@ -60,45 +60,45 @@ static void SerializeEndpoint(const Endpoint& ep, Json::Value& node) {
   node["port"] = ep.port();
 }
 
-bool NetworkStatusInspector::handleGetEndpoints(struct mg_connection* conn, const QueryParams& queryParams) {
-  collector::AdvertisedEndpointMap endpointStates = conntracker_->FetchEndpointState(true, false);
-  Json::Value bodyRoot(Json::objectValue);
+bool NetworkStatusInspector::handleGetEndpoints(struct mg_connection* conn, const QueryParams& query_params) {
+  collector::AdvertisedEndpointMap endpoint_states = conntracker_->FetchEndpointState(true, false);
+  Json::Value body_root(Json::objectValue);
 
-  std::unordered_map<std::string, std::forward_list<collector::ContainerEndpointMap::value_type*>> byContainer;
+  std::unordered_map<std::string, std::forward_list<collector::ContainerEndpointMap::value_type*>> by_container;
 
-  std::optional<std::string> containerFilter = GetParameter(queryParams, kQueryParam_container);
+  std::optional<std::string> container_filter = GetParameter(query_params, kQueryParam_container);
 
-  for (auto& endpointState : endpointStates) {
-    if (!containerFilter || *containerFilter == endpointState.first.container()) {
-      byContainer[endpointState.first.container()].push_front(&endpointState);
+  for (auto& endpoint_state : endpoint_states) {
+    if (!container_filter || *container_filter == endpoint_state.first.container()) {
+      by_container[endpoint_state.first.container()].push_front(&endpoint_state);
     }
   }
 
-  for (auto& container : byContainer) {
-    const std::string& containerId = container.first;
+  for (auto& container : by_container) {
+    const std::string& container_id = container.first;
 
-    Json::Value containerArray(Json::arrayValue);
+    Json::Value container_array(Json::arrayValue);
 
-    for (auto endpointState : container.second) {
-      const collector::ContainerEndpoint& endpoint = endpointState->first;
-      const collector::ConnStatus& status = endpointState->second;
+    for (auto endpoint_state : container.second) {
+      const collector::ContainerEndpoint& endpoint = endpoint_state->first;
+      const collector::ConnStatus& status = endpoint_state->second;
 
-      Json::Value endpointNode(Json::objectValue);
+      Json::Value endpoint_node(Json::objectValue);
 
-      endpointNode["active"] = status.IsActive();
+      endpoint_node["active"] = status.IsActive();
 
-      endpointNode["l4proto"] = Proto2str(endpoint.l4proto());
+      endpoint_node["l4proto"] = Proto2str(endpoint.l4proto());
 
-      endpointNode["endpoint"] = Json::objectValue;
-      SerializeEndpoint(endpoint.endpoint(), endpointNode["endpoint"]);
+      endpoint_node["endpoint"] = Json::objectValue;
+      SerializeEndpoint(endpoint.endpoint(), endpoint_node["endpoint"]);
 
-      containerArray.append(endpointNode);
+      container_array.append(endpoint_node);
     }
 
-    bodyRoot[containerId] = containerArray;
+    body_root[container_id] = container_array;
   }
 
-  std::string buffer = Json::writeString(jsonStreamWriterBuilder_, bodyRoot);
+  std::string buffer = Json::writeString(json_stream_writer_builder_, body_root);
 
   mg_send_http_ok(conn, "application/json", buffer.size());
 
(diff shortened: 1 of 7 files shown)


vikin91:
I <3 this PR, it helps so much! Could we have it merged?

erthalion (approved):
Looks good, thanks. I guess the use case is a manual inspection of the networking state, right?

Could you pls add a few line to the documentation, troubleshooting section? As well as an integration test to fetch those endpoints, making sure they didn't get broken (the purpose is troubleshooting, so we're going to exercise them not so frequently, increasing chances something will break unnoticed).

erthalion (approved):

ovalenti:
I rebased on top of @Molter73 's ContainerInfo introspection endpoint.

We now share a common base class.

The env var had already been merged, so removed it from this PR.

I merged the doc paragraphs, but kept the entry explaining `DEBUG_ENABLE_INTROSPECTION` references.md

collector/lib/NetworkStatusInspector.cpp
```diff
@@ -0,0 +1,177 @@
...
+  Json::Value bodyRoot(Json::objectValue);
+
+  std::unordered_map<std::string, std::forward_list<collector::ContainerEndpointMap::value_type*>> byContainer;
+
+  std::optional<std::string> containerFilter = GetParameter(queryParams, kQueryParam_container);
+
+  for (auto& endpointState : endpointStates) {
+    if (!containerFilter || *containerFilter == endpointState.first.container())
+      byContainer[endpointState.first.container()].push_front(&endpointState);
+  }
```
Molter73, line 74: Oh how I wish we could do something like `.filter(...).map(...)` in C++!

collector/lib/NetworkStatusInspector.cpp
```diff
@@ -0,0 +1,177 @@
...
+  collector::AdvertisedEndpointMap endpointStates = conntracker_->FetchEndpointState(true, false);
+  Json::Value bodyRoot(Json::objectValue);
+
+  std::unordered_map<std::string, std::forward_list<collector::ContainerEndpointMap::value_type*>> byContainer;
+
+  std::optional<std::string> containerFilter = GetParameter(queryParams, kQueryParam_container);
+
+  for (auto& endpointState : endpointStates) {
+    if (!containerFilter || *containerFilter == endpointState.first.container())
+      byContainer[endpointState.first.container()].push_front(&endpointState);
```
Molter73, line 73: Can we put braces around this if statement? Not having them makes me nervous.
```suggestion
    if (!containerFilter || *containerFilter == endpointState.first.container()) {
      byContainer[endpointState.first.container()].push_front(&endpointState);
    }
```

collector/lib/NetworkStatusInspector.cpp
```diff
@@ -0,0 +1,177 @@
...
+      endpointNode["endpoint"] = Json::objectValue;
+      SerializeEndpoint(endpoint.endpoint(), endpointNode["endpoint"]);
+
+      containerArray.append(endpointNode);
+    }
+
+    bodyRoot[containerId] = containerArray;
+  }
+
+  std::string buffer = Json::writeString(jsonStreamWriterBuilder_, bodyRoot);
```
Molter73, line 100: Haven't had time to test this, do you know if the output of `Json::writeString` is formatted? I prefer it unformatted, since I pipe it into `jq` most of the times anyways, but wither way works I guess.

collector/lib/NetworkStatusInspector.cpp
```diff
@@ -0,0 +1,177 @@
...
+  collector::ConnMap connectionStates = conntracker_->FetchConnState(true, false);
+  Json::Value bodyRoot(Json::objectValue);
+
+  std::unordered_map<std::string, std::forward_list<collector::ConnMap::value_type*>> byContainer;
+
+  std::optional<std::string> containerFilter = GetParameter(queryParams, kQueryParam_container);
+
+  for (auto& connectionState : connectionStates) {
+    if (!containerFilter || *containerFilter == connectionState.first.container())
+      byContainer[connectionState.first.container()].push_front(&connectionState);
```
Molter73, line 119: Braces here? Maybe?
```suggestion
    if (!containerFilter || *containerFilter == connectionState.first.container()) {
      byContainer[connectionState.first.container()].push_front(&connectionState);
    }
```

collector/lib/NetworkStatusInspector.cpp
```diff
@@ -0,0 +1,177 @@
...
+  auto parameters = ParseParameters(req_info->query_string);
+  if (uri == kEndpointRoute) {
+    return handleGetEndpoints(conn, parameters);
+  } else if (uri == kConnectionRoute) {
+    return handleGetConnections(conn, parameters);
+  }
+  return ClientError(conn, "unknown route");
+}
+
+}  // namespace collector
```
Molter73, line 177: clang-format is not picking up this missing newline? Weird 🤔

collector/lib/NetworkStatusInspector.h
```diff
@@ -0,0 +1,38 @@
...
+  static const std::string kConnectionRoute;
+
+  NetworkStatusInspector(const std::shared_ptr<ConnectionTracker> conntracker);
+
+  // implementation of CivetHandler
+  bool handleGet(CivetServer* server, struct mg_connection* conn) override;
+
+ private:
+  const std::shared_ptr<ConnectionTracker> conntracker_;
+  Json::StreamWriterBuilder jsonStreamWriterBuilder_;
```
Molter73, line 29: Would it make sense to turn this into a `Json::StreamWriter` and use that for printing? Right now we are creating and destroying one on every API call, or is a race condition on multiple calls a concern and I'm just being dumb?

integration-tests/suites/http_endpoint_availability.go
```diff
@@ -0,0 +1,47 @@
...
+	Endpoints        []string
+}
+
+func (s *HttpEndpointAvailabilityTestSuite) SetupSuite() {
+	s.StartCollector(false, &s.CollectorOptions)
+}
+
+func (s *HttpEndpointAvailabilityTestSuite) TestAvailability() {
+	for _, endpoint := range s.Endpoints {
+		url := fmt.Sprintf("http://localhost:%d%s", s.Port, endpoint)
```
Molter73, line 25: At some point we might want to add a way to get the IP collector can be found to the `collector.Manager` interface, or even a method for getting the full URL to the introspection port, for now this is fine.

integration-tests/suites/http_endpoint_availability.go
```diff
@@ -0,0 +1,47 @@
...
+
+		fmt.Println(url)
+
+		resp, err := http.Get(url)
+		s.Require().NoError(err)
+		s.Require().NotNil(resp)
+		if resp != nil {
```
Molter73, line 32: Checking for nil twice seems a bit overkill.

integration-tests/suites/http_endpoint_availability.go
```diff
@@ -0,0 +1,47 @@
...
+			s.Require().Equal(resp.StatusCode, http.StatusOK, "HTTP status code")
+
+			var buf bytes.Buffer
+			defer resp.Body.Close()
+			_, err = buf.ReadFrom(resp.Body)
+			s.Require().NoError(err)
+			s.Require().True(json.Valid(buf.Bytes()))
```
Molter73, line 39: Any way we could do some validation on the actual content of the JSON response? With this test the endpoint could be returning an empty object incorrectly and this wouldn't catch it.

integration-tests/suites/http_endpoint_availability.go
ovalenti: I think that the testing framework tries to continue execution as far as possible and reports a failure only after returning.

collector/lib/NetworkStatusInspector.cpp
ovalenti: We could do:
```c++
std::copy_if(connectionStates.begin(),
             connectionStates.end(),
             std::back_inserter(byContainer),
             [&containerFilter](collector::ConnMap::iterator::value_type& connectionState) -> bool {
               return !containerFilter || *containerFilter == connectionState.first.container();
             });
```

But I find it less readable.

integration-tests/suites/http_endpoint_availability.go
Molter73: I'm sure it doesn't, if you see that execution continues it is most likely due to the teardown section gets called even if an error occurred, but execution is stopped immediately.

collector/lib/NetworkStatusInspector.cpp
Molter73: Yeah, I got the same response from googling and I don't think it's a good alternative.

integration-tests/suites/http_endpoint_availability.go
ovalenti: My bad, Require() stops immediately and Assert() does not.

collector/lib/NetworkStatusInspector.h
ovalenti: The FastWriter seems just right for that purpose, I replaced with a single instance of it.

integration-tests/suites/http_endpoint_availability.go
ovalenti: The test results from a request of @erthalion, and is very minimally testing that we don't break the introspection API without noticing it. Testing the content is more work and is covered by the MockSensor integration tests (not through the introspection API though, but this is not a customer feature).

Molter73 (approved):
LGTM!

PR #1377, opened 2023-10-19
Remove dependency on test_platform naming

Molter73:
## Description

Variables that used to depend on the contents of the test_platform now instead depend on the contents of the associated virtual_machines.

## Checklist

## Testing Performed

CI should be enough.

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -5,11 +5,11 @@ ibmcloud_ssh_key_file: "{{ lookup('env', 'IBMCLOUD_SSH_KEY_FILE', default='~/.ss
 
 current_user: "{{ lookup('env', 'USER') }}"
 test_platform: "{{ lookup('env', 'VM_TYPE', default='rhel') }}"
-arch_info: "{{ 's390x' if 's390x' in test_platform else 'ppc64le' if 'ppc64le' in test_platform else 'amd64'}}"
+arch_info: "{{ virtual_machines[test_platform].arch | default('amd64') }}"
 collector_root: "{{ playbook_dir }}/.."
 collector_repo: "quay.io/stackrox-io"
 
-ansible_ssh_private_key_file: "{{ ibmcloud_ssh_key_file if arch_info != 'amd64' else gcp_ssh_key_file }}"
+ansible_ssh_private_key_file: "{{ virtual_machines[test_platform].ssh_key_file | default(gcp_ssh_key_file) }}"
 
 # These control the container runtime used on the VMs.
 # They can be overridden in platform-specific group_vars files
@@ -36,6 +36,8 @@ virtual_machines:
 
   rhel-s390x:
     project: rhel-s390x-cloud
+    arch: s390x
+    ssh_key_file: ibmcloud_ssh_key_file
     families:
       - rhel-8-6-s390x
 
@@ -43,6 +45,8 @@ virtual_machines:
   # in ibm cloud (56 chars max)
   rhel-ppc64le:
     project: rhel-ppc64le-cloud
+    arch: ppc64le
+    ssh_key_file: ibmcloud_ssh_key_file
     families:
       - p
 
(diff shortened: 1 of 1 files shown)


ansible/group_vars/all.yml
```diff
@@ -65,7 +69,7 @@ virtual_machines:
         users:
           - name: core
             sshAuthorizedKeys:
-              - "{{ lookup('file', gcp_ssh_key_file + '.pub') }}"
+              - "{{ lookup('file', gcp_ssh_key_file + '.pub', errors='ignore') }}"
```
Molter73, line 72: This is needed for a somewhat convoluted reason. When building the collector images, `virtual_machines` was not evaluated because it wasn't used and so this line was ignored, now `virtual_machines` is needed for both `arch_info` and `ansible_ssh_private_key_file`, so this line is evaluated and fails. That file is only really needed for RHCOS testing though, so ignoring the missing file should be fine.

Stringy:
Looks good! 

I wonder if we can also remove `arch_info` entirely, in favour of the new arch field in the VM structure. We'd need to add a field `vm_arch` [here](https://github.com/stackrox/collector/blob/master/ansible/roles/create-all-vms/tasks/by-family.yml) and [here](https://github.com/stackrox/collector/blob/master/ansible/roles/create-all-vms/tasks/by-family.yml) and also modify [these](https://github.com/stackrox/collector/blob/master/ansible/roles/create-vm/tasks/main.yml#L39-L45) checks.

It might also make sense to add a label to at least the GCP VMs, which we can use to group by architecture in the inventory

Molter73:
> I wonder if we can also remove `arch_info` entirely, in favour of the new arch field in the VM structure. We'd need to add a field `vm_arch` [here](https://github.com/stackrox/collector/blob/master/ansible/roles/create-all-vms/tasks/by-family.yml) and [here](https://github.com/stackrox/collector/blob/master/ansible/roles/create-all-vms/tasks/by-family.yml) and also modify [these](https://github.com/stackrox/collector/blob/master/ansible/roles/create-vm/tasks/main.yml#L39-L45) checks.

Problem is `arch_info` is used in https://github.com/stackrox/collector/blob/master/ansible/roles/destroy-vm/tasks/main.yml and I'm not exactly sure how to replace it there. Also, I might group all `vm_` vars in a single `vm` dictionary, because I started refactoring and now I can't really stop.

> It might also make sense to add a label to at least the GCP VMs, which we can use to group by architecture in the inventory

I have no idea how to do this but I'll look into it.

Stringy:
For GCP labeling add to [here](https://github.com/stackrox/collector/blob/master/ansible/roles/create-vm/tasks/main.yml#L28):

```
- set_fact:
    gcp_extra_labels:
      platform: "{{ vm_platform }}"
      job_id: "{{ job_id }}"
      vm_config: "{{ vm_config }}"
      vm_image_family: "{{ vm_family }}"
      vm_collection_method: "{{ vm_collection_method }}"
      vm_arch: "{{ vm_arch }}"
```
(presumes that vm_arch is set already in by_image.yml and by_family.yml)

and then we can add an architecture grouping in `{dev,ci}/inventory_gcp.yml` under the `keyed_groups` section.

For s390x and ppc64le, we'll need to modify the inventory template in `roles/create-vms/templates/ibm-cloud-inventory.j2` to include the new `vm_arch` variable, and pass the variable in [here](https://github.com/stackrox/collector/blob/master/ansible/roles/create-vm/tasks/create-s390x-vm.yml#L62) and [here](https://github.com/stackrox/collector/blob/master/ansible/roles/create-vm/tasks/create-ppc64le-vm.yml#L149)

With that variable for s390x and ppc64le, we can remove the arch_info requirement from destroy-vms because all VMs will have `vm_arch` set for their lifetime.

Stringy (approved):
This is fantastic, and makes the whole VM management process loads clearer and more intuitive. LGTM!

PR #1380, opened 2023-10-20
Enable s390x always in CPaaS workflows

Stringy:
## Description

Enables s390x testing in CPaaS workflow.

There is an ongoing workflow to add it to PRs as well as on the CPaaS schedule, but this will give us some idea of what's affecting s390x more frequently than currently.

## Checklist

## Testing Performed

CI is sufficient

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -36,7 +36,7 @@ virtual_machines:
   rhel-s390x:
     project: rhel-s390x-cloud
     arch: s390x
-    ssh_key_file: ibmcloud_ssh_key_file
+    ssh_key_file: "{{ ibmcloud_ssh_key_file }}"
     families:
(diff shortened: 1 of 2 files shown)


ansible/group_vars/all.yml
```diff
@@ -36,7 +36,7 @@ virtual_machines:
   rhel-s390x:
     project: rhel-s390x-cloud
     arch: s390x
-    ssh_key_file: ibmcloud_ssh_key_file
+    ssh_key_file: "{{ ibmcloud_ssh_key_file }}"
```
Molter73, line 39: Dang! My bad

Stringy: and my bad for missing it in the review 😄

Molter73 (approved):
LGTM! Provided CI passes

PR #1383, opened 2023-10-20
ROX-19778: Load metrics for connections

ovalenti:
## Description

We add a set of new Prometheus metrics reporting the total amount of connections known by collector and their creation rate over a reporting-period (scrape-interval).

These metrics are split per connection direction (inbound, outbound) and the location of the peer address (private or remote IP).

## Checklist

**Automated testing**
  - [x] Added unit tests
 
Excerpt on an idle single-node k8s:
```
# HELP rox_connections_total Amount of stored connections over time
# TYPE rox_connections_total summary
rox_connections_total_count{dir="out",peer="public"} 36
rox_connections_total_sum{dir="out",peer="public"} 18
rox_connections_total{dir="out",peer="public",quantile="0.5"} 0
rox_connections_total{dir="out",peer="public",quantile="0.9"} 1
rox_connections_total{dir="out",peer="public",quantile="0.95"} 3
rox_connections_total_count{dir="out",peer="private"} 36
rox_connections_total_sum{dir="out",peer="private"} 59537
rox_connections_total{dir="out",peer="private",quantile="0.5"} 1558
rox_connections_total{dir="out",peer="private",quantile="0.9"} 2067
rox_connections_total{dir="out",peer="private",quantile="0.95"} 2119
rox_connections_total_count{dir="in",peer="public"} 36
rox_connections_total_sum{dir="in",peer="public"} 0
rox_connections_total{dir="in",peer="public",quantile="0.5"} 0
rox_connections_total{dir="in",peer="public",quantile="0.9"} 0
rox_connections_total{dir="in",peer="public",quantile="0.95"} 0
rox_connections_total_count{dir="in",peer="private"} 36
rox_connections_total_sum{dir="in",peer="private"} 5009
rox_connections_total{dir="in",peer="private",quantile="0.5"} 101
rox_connections_total{dir="in",peer="private",quantile="0.9"} 179
rox_connections_total{dir="in",peer="private",quantile="0.95"} 180
# HELP rox_connections_rate Rate of connections over time
# TYPE rox_connections_rate summary
rox_connections_rate_count{dir="out",peer="public"} 35
rox_connections_rate_sum{dir="out",peer="public"} 0.06666667014360428
rox_connections_rate{dir="out",peer="public",quantile="0.5"} 0
rox_connections_rate{dir="out",peer="public",quantile="0.9"} 0
rox_connections_rate{dir="out",peer="public",quantile="0.95"} 0
rox_connections_rate_count{dir="out",peer="private"} 35
rox_connections_rate_sum{dir="out",peer="private"} 1947.28048324585
rox_connections_rate{dir="out",peer="private",quantile="0.5"} 51.43333435058594
rox_connections_rate{dir="out",peer="private",quantile="0.9"} 67.80000305175781
rox_connections_rate{dir="out",peer="private",quantile="0.95"} 69.53333282470703
rox_connections_rate_count{dir="in",peer="public"} 35
rox_connections_rate_sum{dir="in",peer="public"} 0
rox_connections_rate{dir="in",peer="public",quantile="0.5"} 0
rox_connections_rate{dir="in",peer="public",quantile="0.9"} 0
rox_connections_rate{dir="in",peer="public",quantile="0.95"} 0
rox_connections_rate_count{dir="in",peer="private"} 35
rox_connections_rate_sum{dir="in",peer="private"} 119.9425313472748
rox_connections_rate{dir="in",peer="private",quantile="0.5"} 2.17241382598877
rox_connections_rate{dir="in",peer="private",quantile="0.9"} 4.800000190734863
rox_connections_rate{dir="in",peer="private",quantile="0.95"} 4.833333492279053
```

## Testing performed

- setting `ROX_COLLECTOR_ENABLE_CONNECTION_STATS` to _false_ shows all the metrics value as NaN or 0.
- the values of `ROX_COLLECTOR_CONNECTION_STATS_QUANTILES`, `ROX_COLLECTOR_CONNECTION_STATS_ERROR` and `ROX_COLLECTOR_CONNECTION_STATS_WINDOW` are parsed and appear in the logs.

### DIFF
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -40,6 +40,8 @@ BoolEnvVar collect_connection_status("ROX_COLLECT_CONNECTION_STATUS", true);
 
 BoolEnvVar enable_external_ips("ROX_ENABLE_EXTERNAL_IPS", false);
 
+BoolEnvVar enable_connection_stats("ROX_COLLECTOR_ENABLE_CONNECTION_STATS", true);
+
 }  // namespace
 
 constexpr bool CollectorConfig::kTurnOffScrape;
@@ -61,6 +63,7 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
   import_users_ = set_import_users.value();
   collect_connection_status_ = collect_connection_status.value();
   enable_external_ips_ = enable_external_ips.value();
+  enable_connection_stats_ = enable_connection_stats.value();
 
   for (const auto& syscall : kSyscalls) {
     syscalls_.push_back(syscall);
@@ -160,6 +163,48 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
   HandleAfterglowEnvVars();
 
   host_config_ = ProcessHostHeuristics(*this);
+
+  const char* envvar;
+
+  connection_stats_quantiles_ = {0.50, 0.90, 0.95};
+
+  if ((envvar = std::getenv("ROX_COLLECTOR_CONNECTION_STATS_QUANTILES")) != NULL) {
+    connection_stats_quantiles_.clear();
+    std::stringstream quantiles(envvar);
+    while (quantiles.good()) {
+      std::string quantile;
+      std::getline(quantiles, quantile, ',');
+      try {
+        double v = std::stod(quantile);
+        connection_stats_quantiles_.push_back(v);
+        CLOG(INFO) << "Connection statistics quantile: " << v;
+      } catch (...) {
+        CLOG(ERROR) << "Invalid quantile value: '" << quantile << "'";
+      }
+    }
+  }
+
+  connection_stats_error_ = 0.01;
+
+  if ((envvar = std::getenv("ROX_COLLECTOR_CONNECTION_STATS_ERROR")) != NULL) {
+    try {
+      connection_stats_error_ = std::stod(envvar);
+      CLOG(INFO) << "Connection statistics error value: " << connection_stats_error_;
+    } catch (...) {
+      CLOG(ERROR) << "Invalid quantile error value: '" << envvar << "'";
+    }
+  }
+
+  connection_stats_window_ = 60;
+
+  if ((envvar = std::getenv("ROX_COLLECTOR_CONNECTION_STATS_WINDOW")) != NULL) {
+    try {
+      connection_stats_window_ = std::stoi(envvar);
+      CLOG(INFO) << "Connection statistics window: " << connection_stats_window_;
+    } catch (...) {
+      CLOG(ERROR) << "Invalid window length value: '" << envvar << "'";
+    }
+  }
 }
 
 void CollectorConfig::HandleAfterglowEnvVars() {
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -161,9 +161,41 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
   }
 
   HandleAfterglowEnvVars();
+  HandleConnectionStatsEnvVars();
 
   host_config_ = ProcessHostHeuristics(*this);
+}
 
+void CollectorConfig::HandleAfterglowEnvVars() {
+  if (!set_enable_afterglow) {
+    enable_afterglow_ = false;
+  }
+
+  if (const char* afterglow_period = std::getenv("ROX_AFTERGLOW_PERIOD")) {
+    afterglow_period_micros_ = static_cast<int64_t>(atof(afterglow_period) * 1000000);
+  }
+
+  if (enable_afterglow_ && afterglow_period_micros_ > 0) {
+    CLOG(INFO) << "Afterglow is enabled";
+    return;
+  }
+
+  if (!enable_afterglow_) {
+    CLOG(INFO) << "Afterglow is disabled";
+    return;
+  }
+
+  if (afterglow_period_micros_ < 0) {
+    CLOG(WARNING) << "Invalid afterglow period " << afterglow_period_micros_ / 1000000 << ". ROX_AFTERGLOW_PERIOD must be positive.";
+  } else {
+    CLOG(WARNING) << "Afterglow period set to 0";
+  }
+
+  enable_afterglow_ = false;
+  CLOG(INFO) << "Disabling afterglow";
+}
+
+void CollectorConfig::HandleConnectionStatsEnvVars() {
   const char* envvar;
 
   connection_stats_quantiles_ = {0.50, 0.90, 0.95};
@@ -205,35 +237,6 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
       CLOG(ERROR) << "Invalid window length value: '" << envvar << "'";
     }
   }
-}
-
-void CollectorConfig::HandleAfterglowEnvVars() {
-  if (!set_enable_afterglow) {
-    enable_afterglow_ = false;
-  }
-
-  if (const char* afterglow_period = std::getenv("ROX_AFTERGLOW_PERIOD")) {
-    afterglow_period_micros_ = static_cast<int64_t>(atof(afterglow_period) * 1000000);
-  }
-
-  if (enable_afterglow_ && afterglow_period_micros_ > 0) {
-    CLOG(INFO) << "Afterglow is enabled";
-    return;
-  }
-
-  if (!enable_afterglow_) {
-    CLOG(INFO) << "Afterglow is disabled";
-    return;
-  }
-
-  if (afterglow_period_micros_ < 0) {
-    CLOG(WARNING) << "Invalid afterglow period " << afterglow_period_micros_ / 1000000 << ". ROX_AFTERGLOW_PERIOD must be positive.";
-  } else {
-    CLOG(WARNING) << "Afterglow period set to 0";
-  }
-
-  enable_afterglow_ = false;
-  CLOG(INFO) << "Disabling afterglow";
 }
 
 bool CollectorConfig::TurnOffScrape() const {
--- a/collector/lib/ConnTracker.cpp
+++ b/collector/lib/ConnTracker.cpp
@@ -145,7 +145,7 @@ bool EmplaceOrUpdate(UnorderedMap<T, ConnStatus>* m, const T& obj, ConnStatus st
 void ConnectionTracker::EmplaceOrUpdateNoLock(const Connection& conn, ConnStatus status) {
   COUNTER_INC(CollectorStats::net_conn_updates);
   if (EmplaceOrUpdate(&conn_state_, conn, status)) {
-    ClassifyConnectionStats(conn, inserted_connections_counters_);
+    IncrementConnectionStats(conn, inserted_connections_counters_);
   }
 }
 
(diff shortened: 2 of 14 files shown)


JoukoVirtanen:
Can you get the metrics and copy and paste the relevant ones here?

ovalenti:
> Can you get the metrics and copy and paste the relevant ones here?

I added a sample to the PR description :+1:

JoukoVirtanen:
There is a README for metrics at docs/troubleshooting.md. I think these metrics should be explained there.

collector/test/ConnTrackerTest.cpp
```diff
@@ -1525,6 +1525,69 @@ TEST(ConnTrackerTest, TestAdvertisedEndpointEquality) {
       ContainerEndpoint("container", a, L4Proto::TCP, processWithDifferentArgs)));
 }
 
+TEST(ConnTrackerTest, TestConnectionStats) {
```
JoukoVirtanen, line 1528: Thanks for adding unit tests

collector/lib/CollectorConfig.cpp
```diff
@@ -160,6 +163,48 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
...
+    connection_stats_quantiles_.clear();
+    std::stringstream quantiles(envvar);
+    while (quantiles.good()) {
+      std::string quantile;
+      std::getline(quantiles, quantile, ',');
+      try {
+        double v = std::stod(quantile);
+        connection_stats_quantiles_.push_back(v);
+        CLOG(INFO) << "Connection statistics quantile: " << v;
+      } catch (...) {
```
JoukoVirtanen, line 181: [nit] Consider catching more specific errors.

collector/lib/CollectorConfig.cpp
```diff
@@ -160,6 +163,48 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
...
+        CLOG(ERROR) << "Invalid quantile value: '" << quantile << "'";
+      }
+    }
+  }
```
Molter73, line 185: Probably a good idea to move this to a dedicated `GetConnectionStatsQuantiles()` function in `Utility.cpp` in order to reduce clutter.

collector/lib/CollectorConfig.cpp
```diff
@@ -160,6 +163,48 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
...
+  const char* envvar;
+
+  connection_stats_quantiles_ = {0.50, 0.90, 0.95};
+
+  if ((envvar = std::getenv("ROX_COLLECTOR_CONNECTION_STATS_QUANTILES")) != NULL) {
+    connection_stats_quantiles_.clear();
+    std::stringstream quantiles(envvar);
+    while (quantiles.good()) {
+      std::string quantile;
+      std::getline(quantiles, quantile, ',');
```
Molter73, line 176: Wait, `ROX_COLLECTOR_CONNECTION_STATS_QUANTILES` is a new line separated environment variable? That doesn't sound very ergonomic, why not use commas, colons, semicolons or some other printable character?

collector/lib/CollectorConfig.cpp
```diff
@@ -160,6 +163,48 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
...
+  connection_stats_error_ = 0.01;
+
+  if ((envvar = std::getenv("ROX_COLLECTOR_CONNECTION_STATS_ERROR")) != NULL) {
+    try {
+      connection_stats_error_ = std::stod(envvar);
+      CLOG(INFO) << "Connection statistics error value: " << connection_stats_error_;
+    } catch (...) {
+      CLOG(ERROR) << "Invalid quantile error value: '" << envvar << "'";
+    }
+  }
```
Molter73, line 196: This could also be moved to its own function elsewhere to reduce clutter.

collector/lib/CollectorConfig.cpp
```diff
@@ -160,6 +163,48 @@ CollectorConfig::CollectorConfig(CollectorArgs* args) {
...
+  connection_stats_window_ = 60;
+
+  if ((envvar = std::getenv("ROX_COLLECTOR_CONNECTION_STATS_WINDOW")) != NULL) {
+    try {
+      connection_stats_window_ = std::stoi(envvar);
+      CLOG(INFO) << "Connection statistics window: " << connection_stats_window_;
+    } catch (...) {
+      CLOG(ERROR) << "Invalid window length value: '" << envvar << "'";
+    }
+  }
```
Molter73, line 207: This could also be moved to its own function elsewhere to reduce clutter.

collector/lib/ConnTracker.cpp
```diff
@@ -319,4 +323,43 @@ void ConnectionTracker::UpdateIgnoredL4ProtoPortPairs(UnorderedSet<L4ProtoPortPa
     }
   }
 }
+
+// Increment the stat counter matching the connection's characteristics
+inline void ConnectionTracker::ClassifyConnectionStats(Connection conn, ConnectionTracker::Stats& stats) const {
```
erthalion, line 328: The name could be misleading, one could get an impression only classification is happening without actually updating the  stats argument. Probably `<Update/Increment>NetworkStats(stats, conn)` would be better, what do you think?

collector/test/ConnTrackerTest.cpp
```diff
@@ -1525,6 +1525,69 @@ TEST(ConnTrackerTest, TestAdvertisedEndpointEquality) {
...
+  tracker.UpdateConnection(conn6, 0, true);  // inserted
+  tracker.UpdateConnection(conn1, 0, true);  // already known
+
+  stats = tracker.GetConnectionStats_StoredConnections();
+  EXPECT_EQ(stats.inbound.private_, 1);
+  EXPECT_EQ(stats.inbound.public_, 2);
+  EXPECT_EQ(stats.outbound.private_, 1);
+  EXPECT_EQ(stats.outbound.public_, 2);
+
+  stats = tracker.GetConnectionStats_NewConnectionCounters();
```
erthalion, line 1569: If I get it correctly, it would be also great to see a test where new conn counters are different from stored conn counters.

collector/lib/ConnTracker.h
```diff
@@ -138,6 +138,20 @@ class ConnectionTracker {
...
+  //
+  // Statistics on the number of stored connections and their rate creation.
+  //
+  struct Stats {
+    struct {
+      unsigned int public_;
+      unsigned int private_;
+    } inbound, outbound;
+  };
+  // Retrieve the number of connections currently known to ConnTracker, indexed by in/out and public/private nature.
```
erthalion, line 150: Would be great to add a few words about how calculation of those counters differ. E.g. that the former is getting calculated right in there, where the latter is getting updated immediately after a connection is getting added.

erthalion (commented):
Overall looks good, a few notes below.

collector/lib/CollectorConfig.cpp
ovalenti: This is a coma separated string. `getline()` is unfortunately badly named.

collector/lib/CollectorConfig.cpp
```diff
@@ -40,6 +40,8 @@ BoolEnvVar collect_connection_status("ROX_COLLECT_CONNECTION_STATUS", true);
 
 BoolEnvVar enable_external_ips("ROX_ENABLE_EXTERNAL_IPS", false);
 
+BoolEnvVar enable_connection_stats("ROX_COLLECTOR_ENABLE_CONNECTION_STATS", true);
```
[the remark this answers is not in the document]
ovalenti: Shoud this be added to the reference.md page along with the other env vars ? or to the troubleshooting.md (actually I just pushed a small description here) ?

collector/lib/CollectorConfig.cpp
ovalenti: I wanted to keep the error-handling codepath to a minimum, since the INFO log gives the values that have been parsed, I think that a malformed value should be rather easy to spot.

collector/lib/CollectorConfig.cpp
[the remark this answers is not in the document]
ovalenti: Maybe I should rather move those to the reference page....
https://github.com/stackrox/collector/blob/ovalenti/external_networks_metrics/docs/troubleshooting.md#connection-statistics

collector/lib/NetworkStatusNotifier.h
```diff
@@ -16,8 +16,19 @@ class NetworkStatusNotifier : protected ProtoAllocator<sensor::NetworkConnection
...
+      : conn_scraper_(conn_scraper),
+        scrape_interval_(scrape_interval),
+        turn_off_scraping_(turn_off_scrape),
+        scrape_listen_endpoints_(scrape_listen_endpoints),
+        conn_tracker_(std::move(conn_tracker)),
+        afterglow_period_micros_(afterglow_period_micros),
+        enable_afterglow_(use_afterglow),
+        comm_(comm),
+        connections_total_reporter_(connections_total_reporter),
+        connections_rate_reporter_(connections_rate_reporter) {
```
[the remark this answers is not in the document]
ovalenti: Updated as suggested, that's lighter :+1: 

The slight drawback is that now NetworkStatusNotifier is aware of the Config object. But this one is already mocked in the tests, so it should be fine.

collector/lib/CollectorConfig.cpp
ovalenti: Moved all the connection-stats specific to a dedicated method, in the same spirit as Afterglow parsing.

Does this look ok this way ?

collector/lib/CollectorConfig.cpp
Molter73: What do you mean? You can move all of this to a function, including the error handling and return the resulting vector wrapped in a `std::optional` in case the variable is not set.

Molter73: Anyways, you already moved handling of env variables to a separate method, I still think this it quite cluttered, but meh, good enough.

collector/lib/CollectorConfig.cpp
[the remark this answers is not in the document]
erthalion: > Shoud this be added to the reference.md page along with the other env vars ? or to the troubleshooting.md

I think both, just under the different context. The first one should contain simply all configurations with minimal explanation what it is for. The second one should describe how to use them for troubleshooting, in this case probably we can simply mention "in case of network load issues, turn this off/on".

collector/test/ConnTrackerTest.cpp
ovalenti: I added a step to the test, and now the numbers diverge :+1:

docs/troubleshooting.md
```diff
@@ -342,3 +342,85 @@ Units: bytes
...
+
+#### Total number of known connections
+
+```
+Component: ConnectionTracker
+Prometheus names: rox_connections_total
+Units: count
+```
+
+This is the number of connections known to the ConnectionTracker during a reporting interval.
```
JoukoVirtanen, line 364: The number of connections is recorded for each scrape interval within a time window. Quantile="0.5" for example means that in 50% of the scrape intervals x or greater number of connections were reported.

JoukoVirtanen (approved):
LGTM but ideally metrics would be for connections after afterglow.

PR #1384, opened 2023-10-23
Bump disk size for s390x VM when running in dev mode

Molter73:
## Description

IBM cloud VMs are created with 10GB disks by default, this is enough for our tests but it's also a heavy limitation when trying to use one of these VMs for dev work. This patch bumps the size of the created disk to 20GB, which is not huge, but enough for creating a few images and running some tests.

## Checklist

## Testing Performed

- [x] Create a VM manually and check its disk size is 20GB.


Stringy (approved):
LGTM!

PR #1390, opened 2023-10-25
Remove CORE_BPF available message on Power machines

Molter73:
## Description

When running on Power machines, we are wrongly printing a message stating that CORE_BPF is available on the platform, but we don't support it in code, so we shouldn't be printing the message. Additionally, setting CORE_BPF as the collection method should hard fail since it is unsupported.

## Checklist

## Testing Performed

Will try to test this on a power VM, but the change is small enough that should be good enough to be merged as is IMO.

### DIFF
--- a/collector/lib/HostHeuristics.cpp
+++ b/collector/lib/HostHeuristics.cpp
@@ -23,9 +23,15 @@ class CollectionHeuristic : public Heuristic {
                   << " does not support eBPF, which is a requirement for Collector.";
     }
 
+    auto kernel = host.GetKernelVersion();
     // If we're configured to use eBPF with BTF, we try to be conservative
     // and fail instead of falling-back to ebpf.
     if (config.GetCollectionMethod() == CollectionMethod::CORE_BPF) {
+      if (kernel.machine == "ppc64le") {
+        CLOG(FATAL) << "CORE_BPF collection method is not supported on ppc64le. "
+                    << "Please set collector.collectionMethod=EBPF.";
+      }
+
       if (!host.HasBTFSymbols()) {
         CLOG(FATAL) << "Missing BTF symbols, core_bpf is not available. "
                     << "They can be provided by the kernel when configured with DEBUG_INFO_BTF, "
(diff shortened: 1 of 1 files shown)


ovalenti (dismissed):
Looks good to me !

A pity that we have to get into such exception, but this definitely makes it clearer for the user.

collector/lib/HostHeuristics.cpp
```diff
@@ -23,9 +23,15 @@ class CollectionHeuristic : public Heuristic {
                   << " does not support eBPF, which is a requirement for Collector.";
     }
 
+    auto kernel = host.GetKernelVersion();
     // If we're configured to use eBPF with BTF, we try to be conservative
     // and fail instead of falling-back to ebpf.
     if (config.GetCollectionMethod() == CollectionMethod::CORE_BPF) {
+      if (kernel.machine == "ppc64le") {
+        CLOG(FATAL) << "CORE_BPF collection method is not supported on ppc64le. "
+                    << "Please set collector.collectionMethod=EBPF.";
```
erthalion, line 32: Maybe reword it for consistency with other similar messages?
```
"HINT: You may alternatively want to use eBPF based collection "
"with collector.collectionMethod=EBPF."
```

Molter73: My only objection to this is the proposed message makes it seem like this is an option, when in reality collector will not work with `CORE_BPF`. WDYT about:
```cpp
"HINT: Change collection method to EBPF with collector.collectionMethod=EBPF."
```

Molter73:
> Afterthough, we probably should use a build-time check instead

This would be fine, but compile time checks are usually not as nice to read. If this was in a hot path or compilation absolutely depended on the change I'd 100% use a build-time check, but since this is not the case, I think an explicit runtime check is better.

collector/lib/HostHeuristics.cpp
erthalion: Why does it sound like an option to you? It clearly says, CORE_BPF is not supported here, but you could use some other collection method. The other cases down below are the same, something is not supported, and the error message says exactly that.

Molter73: Because `You may alternatively want` is a very passive way of stating something you HAVE to do.

Molter73: > The other cases down below are the same, something is not supported, and the error message says exactly that.

I also don't agree with the other message being so passive, but I'm not changing those in this PR.

Molter73: Changed wording on 2 of the messages, the one that is left with `You may alternatively want` makes sense to keep that wording because providing BTF symbols would also be an option.

erthalion (approved):

Molter73:
Merging now since the integration tests will not be affected by this change and I'd like to start cutting our release.

PR #1393, opened 2023-10-27
Add gperftools as a submodule

Molter73:
## Description

gperftools were historically used for profiling back when the collector image was still based on ubuntu. Unfortunately, UBI repositories don't provide a precompiled version of gperftools, but nothing prevents us from building it ourselves.

## Checklist

## Testing Performed

Manually tested the gperftools API endpoints and the generated dumps.


collector/CMakeLists.txt
```diff
@@ -68,8 +68,8 @@ target_link_libraries(collector_lib z ssl crypto CURL::libcurl bpf)
 add_executable(collector collector.cpp)
 target_link_libraries(collector collector_lib)
 
-if(NOT DEFINED ENV{DISABLE_PROFILING})
-	target_link_libraries(collector profiler tcmalloc)
+if(NOT DEFINED ENV{DISABLE_PROFILING} AND (NOT DEFINED ENV{USE_VALGRIND} OR NOT "$ENV{USE_VALGRIND}"))
```
Molter73, line 71: Looks like we can set the env variable `RUNNING_ON_VALGRIND=1` to prevent gperftools from changing the malloc implementation, allowing valgrind to work correctly, we might want to remove everything after the `AND` in this condition.

Molter73: Ended up with a different approach, basically disabling profiling if we are compiling collector with asan or with the `USE_VALGRIND` flag (though this last one might be unnecessary, since we can just disable profiling manually).

robbycochran:
It doesn't look like P/Z are supported by gperftools, is there any risk as is? should we disable on those archs?

Molter73:
> It doesn't look like P/Z are supported by gperftools, is there any risk as is? should we disable on those archs?

Good point, I think we can disable them until we have some time to test them.

erthalion (approved):

erthalion:
Doing a lightweight perf evaluation shows that gperftools idle presence at
runtime doesn't change things much, so we can assume it has small resources
footprint when not actively profiling.

Here is the graph showing CPU usage under a moderate berserker load (arrival:
10, departure: 200, random: true) with without gperf at all, with gperf idling,
and with gperf actively profiling the heap. The raw data is shown with
transparency, and the rolling median is plotted over it to show the smoother
trend. While without gperf & with idling gperf CPU consumption is essentially
the same, of course when profiling we've got visible overhead.

gperf

The next one is an event propagation latency (i.e. time between a syscall was
fired in the system, and a corresponding event was received by Collector) under
the same load, with and without gperf. Except one outlier, they're again
essentially the same.

latency

To summarize, there seems to be no reasons to think gperftools enabling can
have a significant impact on the performance, unless actively profiling.

PR #1395, opened 2023-10-30
ROX-20541: Temporarily exclude kernels 6.6 from build

ovalenti:
## Description

A patch was introduced to solve build issues on earlier RC versions of 6.6, which is now causing
build issues. It has been mostly reverted upstream.

Block temporarily those kernels while we work on pulling upstream.

## Testing

This does not touch the Collector code, we just want to make the drivers build happy.

### DIFF
--- a/kernel-modules/BLOCKLIST
+++ b/kernel-modules/BLOCKLIST
@@ -72,7 +72,7 @@
 6.2.* 2.2.0 *
 6.2.* 2.1.0 *
 6.2.* 2.0.1 *
-# TODO(ROX-16705) - 6.3/6.4 kernel compilation errors
+# ROX-16705 - 6.3/6.4 kernel compilation errors
 ~6\.[3-9]\.\d+.* ~2\.[2-5]\.\d+(?:-rc\d+)?
 # TODO(ROX-20541) - 6.6 kernel compilation errors
 6.6.* *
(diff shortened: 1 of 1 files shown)


ovalenti:
6.6 is blocked : https://github.com/stackrox/collector/actions/runs/6695747229/job/18192081362?pr=1395#step:7:23983

kernel-modules/BLOCKLIST
```diff
@@ -72,8 +72,10 @@
 6.2.* 2.2.0 *
 6.2.* 2.1.0 *
 6.2.* 2.0.1 *
-# TODO(ROX-16705) - 6.3/6.4 kernel compilation errors
+# ROX-16705 - 6.3/6.4 kernel compilation errors
 ~6\.[3-9]\.\d+.* ~2\.[2-5]\.\d+(?:-rc\d+)?
+# TODO(ROX-20541) - 6.6 kernel compilation errors
+6.6.* *
```
Molter73, line 78: You don't need the second parameter as a `*`, since the script that processes this file assumes all missing parts are `*`. https://github.com/stackrox/collector/blob/24b43e3f04594741fc92957662a4eae41bc53fdb/kernel-modules/build/apply-blocklist.py#L41C8-L41C8

Molter73 (approved):
LGTM!

kernel-modules/BLOCKLIST
ovalenti: And it is explained in the comment header... Removed it !

PR #1400, opened 2023-11-01
Jv processes and endpoints integration test

JoukoVirtanen:
## Description

Adds a new test suite. The main purpose of the test suite is to test cases in which one command is run in one container. The processes and endpoints reported by collector are checked against expected processes and endpoints. Actual tests for the suite are being worked on here https://github.com/stackrox/collector/pull/1340

## Checklist

**Automated testing**
  ~~- [ ] Added unit tests~~
  - [x] Added integration tests

The actual integration tests are being added here https://github.com/stackrox/collector/pull/1340

## Testing Performed

Ran tests locally and in CI here  https://github.com/stackrox/collector/pull/1340

This is still being worked on, but the suite itself is working as intended.

### DIFF
--- a/integration-tests/suites/common/utils.go
+++ b/integration-tests/suites/common/utils.go
@@ -24,3 +24,12 @@ func QuoteArgs(args []string) []string {
 	}
 	return quotedArgs
 }
+
+// Returns the min of two integers.
+// Strangely there is no such built in or function in math
+func Min(x int, y int) int {
+	if x < y {
+		return x
+	}
+	return y
+}
--- a/integration-tests/suites/connections_and_endpoints.go
+++ b/integration-tests/suites/connections_and_endpoints.go
@@ -2,7 +2,6 @@ package suites
 
 import (
 	"fmt"
-	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -133,10 +132,8 @@ func (s *ConnectionsAndEndpointsTestSuite) TestConnectionsAndEndpoints() {
 	if s.Server.ExpectedEndpoints != nil {
 		assert.Equal(s.T(), len(s.Server.ExpectedEndpoints), len(serverEndpoints))
 
-		sort.Slice(s.Server.ExpectedEndpoints, func(i, j int) bool {
-			return endpointComparison(s.Server.ExpectedEndpoints[i], s.Server.ExpectedEndpoints[j])
-		})
-		sort.Slice(serverEndpoints, func(i, j int) bool { return endpointComparison(serverEndpoints[i], serverEndpoints[j]) })
+		types.SortEndpoints(s.Server.ExpectedEndpoints)
+		types.SortEndpoints(serverEndpoints)
 
 		for idx := range serverEndpoints {
 			assert.Equal(s.T(), s.Server.ExpectedEndpoints[idx].Protocol, serverEndpoints[idx].Protocol)
@@ -146,37 +143,4 @@ func (s *ConnectionsAndEndpointsTestSuite) TestConnectionsAndEndpoints() {
 		assert.Equal(s.T(), 0, len(serverEndpoints))
 	}
 
-}
-
-func endpointComparison(endpoint1 types.EndpointInfo, endpoint2 types.EndpointInfo) bool {
-	addr1, addr2 := endpoint1.Address, endpoint2.Address
-
-	if addr1 == nil {
-		return false
-	}
-	if addr2 == nil {
-		return true
-	}
-
-	if addr1.AddressData < addr2.AddressData {
-		return true
-	}
-
-	if addr1.AddressData > addr2.AddressData {
-		return false
-	}
-
-	if addr1.Port < addr2.Port {
-		return true
-	}
-
-	if addr1.Port > addr2.Port {
-		return false
-	}
-
-	if endpoint1.Protocol < endpoint2.Protocol {
-		return true
-	}
-
-	return false
 }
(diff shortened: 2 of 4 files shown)


integration-tests/suites/connections_and_endpoints.go
```diff
@@ -147,36 +144,3 @@ func (s *ConnectionsAndEndpointsTestSuite) TestConnectionsAndEndpoints() {
 	}
 
 }
-
-func endpointComparison(endpoint1 types.EndpointInfo, endpoint2 types.EndpointInfo) bool {
```
JoukoVirtanen, line 151: Makes more sense to move this to somewhere that can be called by multiple test suites.

integration-tests/suites/processes_and_endpoints.go
```diff
@@ -0,0 +1,83 @@
...
+	for idx := 0; idx < minEndpoints; idx++ {
+		assert.Equal(s.T(), s.ExpectedEndpoints[idx].Protocol, endpoints[idx].Protocol)
+		assert.Equal(s.T(), s.ExpectedEndpoints[idx].Address.AddressData, endpoints[idx].Address.AddressData)
+		assert.Equal(s.T(), s.ExpectedEndpoints[idx].Address.Port, endpoints[idx].Address.Port)
+		assert.Equal(s.T(), s.ExpectedEndpoints[idx].Address.IpNetwork, endpoints[idx].Address.IpNetwork)
+		assert.Equal(s.T(), s.ExpectedEndpoints[idx].IsActive(), endpoints[idx].IsActive())
+		assert.Equal(s.T(), s.ExpectedEndpoints[idx].Originator.ProcessName, endpoints[idx].Originator.ProcessName)
+		assert.Equal(s.T(), s.ExpectedEndpoints[idx].Originator.ProcessExecFilePath, endpoints[idx].Originator.ProcessExecFilePath)
+		assert.Equal(s.T(), s.ExpectedEndpoints[idx].Originator.ProcessArgs, endpoints[idx].Originator.ProcessArgs)
+	}
```
Stringy, line 74: We might be able to use `assert.ElementsMatch(s.T(), s.ExpectedEndpoints, endpoints)` here (sorting may be required)

integration-tests/suites/processes_and_endpoints.go
```diff
@@ -0,0 +1,83 @@
...
+
+	minProcesses := common.Min(len(s.ExpectedProcesses), len(processes))
+
+	for idx := 0; idx < minProcesses; idx++ {
+		assert.Equal(s.T(), s.ExpectedProcesses[idx].Name, processes[idx].Name)
+		assert.Equal(s.T(), s.ExpectedProcesses[idx].ExePath, processes[idx].ExePath)
+		assert.Equal(s.T(), s.ExpectedProcesses[idx].Args, processes[idx].Args)
+	}
```
Stringy, line 82: As above; we might be able to use `assert.ElementsMatch(s.T(), s.ExpectedProcesses, processes)`

integration-tests/suites/processes_and_endpoints.go
JoukoVirtanen: I don't think that will work. The problem is this line
```
assert.Equal(s.T(), s.ExpectedEndpoints[idx].IsActive(), endpoints[idx].IsActive())
```
It doesn't make sense to have an expected close timestamp. We can only expect the close timestamp to be null or not null.

integration-tests/suites/processes_and_endpoints.go
JoukoVirtanen: Done

integration-tests/suites/processes_and_endpoints.go
```diff
@@ -0,0 +1,83 @@
...
+	s.cleanupContainers(s.ContainerName)
+	s.WritePerfResults()
+}
+
+func (s *ProcessesAndEndpointsTestSuite) TestProcessesAndEndpoints() {
+	processes := s.Sensor().Processes(s.container)
+	endpoints := s.Sensor().Endpoints(s.container)
+
+	assert.Equal(s.T(), len(s.ExpectedEndpoints), len(endpoints))
+	assert.Equal(s.T(), len(s.ExpectedProcesses), len(processes))
```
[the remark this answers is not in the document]
JoukoVirtanen: Done. Though I changed it to
```
        processes := s.Sensor().ExpectProcessesN(s.T(), s.container, 20 * time.Second, len(s.ExpectedProcesses))
        endpoints := s.Sensor().ExpectEndpointsN(s.T(), s.container, 20 * time.Second, len(s.ExpectedEndpoints))
```

integration-tests/suites/common/utils.go
```diff
@@ -24,3 +24,12 @@ func QuoteArgs(args []string) []string {
 	}
 	return quotedArgs
 }
+
+// Returns the min of two integers.
+// Strangely there is no such built in or function in math
```
Molter73, line 29: Actually, min and max were added to the 1.21 version of golang: https://tip.golang.org/doc/go1.21

I believe they didn't exist because of the lack of templates until that version. Could we bump the required version for our tests and use the builtin instead?

Stringy (approved):

integration-tests/suites/common/utils.go
JoukoVirtanen: I think we should probably try to use the version of Go used by stackrox/stackrox, which is currently 1.20.10. That is the version that I have locally. Since the tests are containerized, I could try just running them that way with different versions of Go on my host and the container. However, that is currently not working for me. I will merge as is and try to come back to this issue.

PR #1401, opened 2023-11-02
Build test container on CPaaS workflow

Molter73:
## Description

Since the integration tests have been containerized, the CPaaS workflow needs to build the image in order for it to be able to run the tests.

## Checklist

## Testing Performed

- [ ] Run CI with `run-cpaas-steps` label.


Stringy (approved):
LGTM! Thanks for fixing this, and apologies for causing it in the first place!

PR #1406, opened 2023-11-06
Miscellaneous build improvements

Molter73:
## Description

This patch does quite a bit of things, mainly:
- Reuse the builder container instead of spinning a new container on each command.
- Refactor CMakeLists files so they depend on less env variables.
- Separate unit tests to their own cmake script.
- Compile unit tests as separate binaries and run them via ctest.
- Add unit tests running under ASAN and UBSAN in CI.

This is meant to be step one of a larger refactor, aiming to be able to build collector on any system and not just inside the builder. Next step is to turn all third party dependencies into cmake modules, giving the option to use system installed dependencies or compiling the dependencies alongside the collector build step.

## Checklist

## Testing Performed

CI should be enough.


Stringy (approved):
LGTM!

robbycochran:
Nice changes!

collector/test/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+enable_testing()
+
+# Unit Tests
+file(GLOB TEST_SRC_FILES ${PROJECT_SOURCE_DIR}/test/*.cpp)
+foreach(test_file ${TEST_SRC_FILES})
+    get_filename_component(test_name ${test_file} NAME_WE)
+    add_executable("${test_name}" "${test_file}")
+
+    if(NOT DISABLE_PROFILING AND NOT USE_VALGRIND)
+        target_link_libraries(${test_name} profiler tcmalloc)
```
erthalion, line 11: Not sure I get it with `NOT USE_VALGRIND`: so, if we don't use Valgrind, we still link with tcmalloc? I think originally it was `NOT DEFINED ENV{USE_VALGRIND}`?

collector/test/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+    add_test(${test_name} ${test_name})
+
+    if(USE_VALGRIND)
+        # TODO: This test has a deadlock when running on valgrind. Further investigation needed.
+        string(COMPARE EQUAL ${test_name} NetworkStatusNotifierTest res)
+        if(res)
+            continue()
+        endif()
+
+        add_test(NAME memcheck_${test_name} COMMAND valgrind -q --leak-check=full $<TARGET_FILE:${test_name}>)
```
erthalion, line 31: Just a quick check, there are no processes spawn from the tests, right? Otherwise we might add `--trace-children=yes` as well.

collector/Makefile
```diff
@@ -17,19 +17,21 @@ COLLECTOR_BUILD_DEPS := $(HDRS) $(SRCS) $(shell find falcosecurity-libs -name '*
...
-		-e TRACE_SINSP_EVENTS=$(TRACE_SINSP_EVENTS) \
-		-e ADDRESS_SANITIZER=$(ADDRESS_SANITIZER) \
-		-e COLLECTOR_APPEND_CID=$(COLLECTOR_APPEND_CID) \
-		-e DISABLE_PROFILING="true" \
-		quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG) "$(SRC_MOUNT_DIR)/builder/build/build-collector.sh"
+cmake-configure/collector:
+	docker exec $(COLLECTOR_BUILDER_NAME) \
+		cmake -S $(BASE_PATH) -B $(CMAKE_DIR) \
+			-DCMAKE_BUILD_TYPE=$(CMAKE_BUILD_TYPE) \
+			-DDISABLE_PROFILING=ON \
```
erthalion, line 24: Wouldn't it be better to have it configurable via an env var as well?

collector/Makefile
```diff
@@ -17,19 +17,21 @@ COLLECTOR_BUILD_DEPS := $(HDRS) $(SRCS) $(shell find falcosecurity-libs -name '*
...
+			-DUSE_VALGRIND=$(USE_VALGRIND) \
+			-DADDRESS_SANITIZER=$(ADDRESS_SANITIZER) \
```
erthalion, line 26: Originally there were some default values specified for this variables in the build-collector script. Are they defined somewhere else now?

.github/workflows/main.yml
```diff
@@ -97,6 +97,34 @@ jobs:
...
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/stackrox-io/collector-builder:${{ needs.build-builder-image.outputs.collector-builder-tag }}
+    needs:
+      - build-builder-image
+    strategy:
+      fail-fast: false
+      matrix:
+        config-args:
+        - -DDISABLE_PROFILING=ON -DADDRESS_SANITIZER=ON
```
erthalion, line 110: Does `DISABLE_PROFILING=ON` means profiling is actually enabled?

Makefile
```diff
@@ -108,20 +108,25 @@ ifeq (,$(wildcard $(DEV_SSH_SERVER_KEY)))
...
+		-v $(CURDIR):$(CURDIR) \
+		-w $(CURDIR) \
+		--cap-add sys_ptrace \
 		quay.io/stackrox-io/collector-builder:$(COLLECTOR_BUILDER_TAG)
 
-.PHONY: teardown-dev
-teardown-dev:
-	-docker rm -fv collector_remote_dev
+.PHONY: check-builder
+check-builder:
```
erthalion, line 121: Would it be handy to actually start the builder image if it's not there, to not require this manual step from a developer? On the CI it's being done for us, as far as I see.

erthalion (approved):
Looks good, thanks. Couple of commentaries below, but they're mostly from "nice to have" category, or clarifications.

One thing that often bothers me is that currently the build process could not be interrupted with SIGINT. Would this be different with this PR?

collector/test/CMakeLists.txt
Molter73: Yeah, this condition is quite confusing. The end goal is that if we ARE using valgrind we don't want to link with gperftools because tcmalloc messes with it. I think what we really want is if `USE_VALGRIND` is `true` then we need to ensure `DISABLE_PROFILING` is also `true`, I have a similar change on a separate PR (too much duplicated efforts, sorry!).

collector/test/CMakeLists.txt
Molter73: AFAIK, our unit tests are single threaded, but there's no harm in adding the flag to be sure.

collector/Makefile
Molter73: Yeah, I actually added this in #1393, but makes sense to have it here.

collector/Makefile
Molter73: Always have been: https://github.com/stackrox/collector/blob/6c763b76bcee5c66ae7bc92b475d9171378f33c3/Makefile-constants.mk#L14-L19

.github/workflows/main.yml
Molter73: No, it means there's no profiling, (I hate double negative variables).

Makefile
Molter73: I can probably add it to the `check-builder.sh` script, I didn't see it as a big deal because I already spin up my own devcontainer for it anyways, but there's no harm in adding it I think.

Molter73:
> One thing that often bothers me is that currently the build process could not be interrupted with SIGINT. Would this be different with this PR?

Just tested it and SIGINT absolutely stops compilation on `make image` with this change.

PR #1407, opened 2023-11-06
WIP: RHTAP Onboarding

tommartensen:
## Description

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


rhtap.Dockerfile
```diff
@@ -0,0 +1,135 @@
...
+ENV SRC_ROOT_DIR=${SRC_ROOT_DIR}
+
+ARG CMAKE_BUILD_DIR
+ENV CMAKE_BUILD_DIR=${CMAKE_BUILD_DIR}
+
+# TODO: CMAKE_BUILD_TYPE should probably not be Release for PR, normal branch builds
+ARG CMAKE_BUILD_TYPE=Release
+ENV CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+
+# TODO: what does this do?
```
Molter73, line 56: This is used on PRs, it appends an argument to the driver download URL that is used for filtering alerts on missing kernels. We haven't really been using it lately, but the TL;DR is that this needs to be true on PRs only.

tommartensen: Added comment, resolution is captured under https://issues.redhat.com/browse/ROX-20240.

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,151 @@
...
+    version=${COLLECTOR_VERSION} \
+    collector_version=${COLLECTOR_VERSION}
+    # TODO: These two labels only exist on upstream, keep or remove?
+    # io.stackrox.collector.module-version="${MODULE_VERSION}" \
+    # io.stackrox.collector.version="${COLLECTOR_VERSION}"
+
+WORKDIR /
+
+COPY collector/container/${BUILD_TYPE}/install.sh /
+RUN ./install.sh && rm -f install.sh
```
msugakov, line 128: Let's please use this as an opportunity to also improve this `install.sh`:

Doing `microdnf -y upgrade --nobest` instead of just `microdnf -y upgrade` saves us from occasional failures when RPM repo isn't in a consistent state.

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,151 @@
...
+
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.2
+
+ARG BUILD_TYPE=rhel
+ARG ROOT_DIR=.
+# TODO(ROX-20236): configure injection of dynamic version value when it becomes possible.
+ARG COLLECTOR_VERSION=0.0.1-todo
+ARG MODULE_VERSION=0.0.1-todo
+
+ENV ROOT_DIR=$ROOT_DIR
```
msugakov, line 102: I don't see `$ROOT_DIR` set in the downstream dockerfile. Are you sure it's needed?

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,151 @@
...
+# TODO(ROX-20651): Set ENV WITH_RHEL_RPMS=true when RHEL RPMs can be installed to enable hermetic builds.
+RUN ./builder/install/install-dependencies.sh \
+    && cmake -DDISABLE_PROFILING=ON -S "${SRC_ROOT_DIR}" -B "${CMAKE_BUILD_DIR}" \
+    && cmake --build "${CMAKE_BUILD_DIR}" --target all -- -j "$(nproc)" \
+    && (cd ${CMAKE_BUILD_DIR} && ctest -V) \
+    && strip --strip-unneeded "${CMAKE_BUILD_DIR}/collector/collector" "${CMAKE_BUILD_DIR}/collector/EXCLUDE_FROM_DEFAULT_BUILD/libsinsp/libsinsp-wrapper.so"
+
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.2
+
+ARG BUILD_TYPE=rhel
```
msugakov, line 96: [nit] Seems like `BUILD_TYPE` is used in only one spot. Maybe it's better to inline it. WDYT?

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,151 @@
...
+    # io.stackrox.collector.version="${COLLECTOR_VERSION}"
+
+WORKDIR /
+
+COPY collector/container/${BUILD_TYPE}/install.sh /
+RUN ./install.sh && rm -f install.sh
+
+COPY collector/container/scripts/collector-wrapper.sh /usr/local/bin
+COPY collector/container/scripts/bootstrap.sh /
+COPY collector/LICENSE-kernel-modules.txt /kernel-modules/LICENSE
```
msugakov, line 132: I can't spot `LICENSE-kernel-modules.txt` being touched downstream, neither the file present in the resulting image. Are you sure it's needed?

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,151 @@
...
+# TODO(ROX-20312): we can't pin image tag or digest because currently there's no mechanism to auto-update that.
+# TODO(ROX-20651): Use RHEL/ubi base image when entitlement is solved.
+# RPMs requiring entitlement: bpftool, cmake-3.18.2-9.el8, elfutils-libelf-devel, tbb-devel, c-ares-devel, jq-devel
+# FROM registry.access.redhat.com/ubi8/ubi:latest AS builder
+FROM quay.io/centos/centos:stream9 AS builder
+
+COPY . .
+
+# TODO(ROX-20234): use hermetic builds when installing/updating RPMs becomes hermetic.
+RUN dnf -y update \
```
msugakov, line 16: TIL that `dnf update` is the same as `dnf upgrade`.

```suggestion
RUN dnf upgrade -y --nobest \
```

See related https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/merge_requests/2263

msugakov (commented):
Posting some intermediate review while I'm going through the code. Most things are cosmetic things that stick out when comparing with downstream dockerfiles.

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,151 @@
+ARG BUILD_DIR=/build
+ARG SRC_ROOT_DIR=${BUILD_DIR}
+ARG CMAKE_BUILD_DIR=${SRC_ROOT_DIR}/cmake-build
+ARG USE_VALGRIND=false
+ARG ADDRESS_SANITIZER=false
```
msugakov, line 5: These ARGs set here don't have any effect. IIUC, they are intended to be used in cmake configure step that is in the Makefile upstream 
https://github.com/stackrox/collector/blob/6d3c3822446be80f372bf3bc40e9cdabce7a6f61/collector/Makefile#L20-L28
(obviously the step cannot be reused here because there's no `docker` to exec).
These variables are introduced in a special file upstream - https://github.com/stackrox/collector/blob/6d3c3822446be80f372bf3bc40e9cdabce7a6f61/Makefile-constants.mk

The problem is that these ARGs don't become ENVs and so remain not visible to the actual command that's supposed to see them. Probably these specific don't cause issues because default values are also `false`.

I know the downstream dockerfile does this too and I know that it's not necessarily the part of this RHTAP onboarding to solve this problem, but...

The problem is that we have multiple differing patterns how these variables are set and used.
1. First is the upstream (see links above).
2. Downstream dockerfile that sets variables and [delegates](https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/blob/rhacs-1.0-rhel-8/distgit/containers/rhacs-collector-slim/Dockerfile.in?ref_type=heads#L64) the rest to `build-collector.sh` file that [no longer exists](https://redhat-internal.slack.com/archives/CFMQ5C2TT/p1700843996467709).
3. This RHTAP dockerfile where you call the cmake command directly (`cmake -DDISABLE_PROFILING=ON -S "${SRC_ROOT_DIR}" -B "${CMAKE_BUILD_DIR}"`). No variables except for `DISABLE_PROFILING` are set. I'm not sure if the assumption that `cmake` respects environment variables is true. The main point is: this is different from the downstream dockerfile and from the upstream approach.

I find this to be a mess which needs fixing. Doing the same thing in three different ways is just too taxing on time and mental capacity. The situation with the downstream dockerfile after `build-collector.sh` removal needs to be addressed anyway. It would be good to have such a solution that will fit RHTAP dockerfile without much adaptation. I suggest getting in touch with the Collector team folks to find a solution.

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,151 @@
...
+    && cp -a /builder/third_party third_party \
+    && cp -a /kernel-modules/MODULE_VERSION kernel-modules/MODULE_VERSION \
+    && cp -a /CMakeLists.txt CMakeLists.txt
+
+# WITH_RHEL_RPMS controls for dependency installation, ie if they were already installed as RPMs.
+# Setting the value to empty will cause dependencies to be downloaded from repositories or accessed in submodules and compiled.
+# TODO(ROX-20651): Set ENV WITH_RHEL_RPMS=true when RHEL RPMs can be installed to enable hermetic builds.
+RUN ./builder/install/install-dependencies.sh \
+    && cmake -DDISABLE_PROFILING=ON -S "${SRC_ROOT_DIR}" -B "${CMAKE_BUILD_DIR}" \
+    && cmake --build "${CMAKE_BUILD_DIR}" --target all -- -j "$(nproc)" \
```
msugakov, line 90: The problem is that `nproc` command is not aware of cgroups and hence does not know how many CPUs were actually allocated to the container. It just reports the number of CPUs seen by the kernel, this will be the number of cores of a node which are shared between all containers.

As we saw in the discussions related to RHTAP performance, CPU limits are still there and node CPUs are (currently) getting much oversubscribed. It may not cause a huge context switching overhead to let the command work with `nproc` result which is `8`, but we're certainly not getting what we asked for.

I suggest to replace this misleading `nproc` with just a fixed value (e.g. via a named variable) which we could manually tune if/when needed.

`build-collector.sh` used to set it to `2`, by the way.
https://github.com/stackrox/collector/blob/acf241f5fd43b5e9c995bbb253a866347d064bc1/builder/build/build-collector.sh#L31

msugakov (commented):
Completed the first pass. I could have certainly missed some things therefore can comment on code I've seen in later iterations.

.tekton/collector-slim-pull-request.yaml
```diff
@@ -0,0 +1,396 @@
...
+        - name: kind
+          value: task
+        resolver: bundles
+
+    - name: clone-repository
+      params:
+      - name: url
+        value: $(params.git-url)
+      - name: revision
+        value: $(params.revision)
```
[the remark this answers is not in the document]
tommartensen: As far as I could tell, the version is not embedded in the binary. 
There is a `COLLECTOR_VERSION` based on `git describe` (https://github.com/stackrox/collector/blob/master/Makefile-constants.mk#L6-L12) or injected. That is passed to the container build as an argument (https://github.com/stackrox/collector/blob/master/Makefile#L63-L68) and set as an env variable on upstream [here](https://github.com/stackrox/collector/blob/master/collector/container/Dockerfile#L9C1-L9C45), but not on downstream: https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/blob/rhacs-4.3-rhel-8/distgit/containers/rhacs-collector-slim/Dockerfile.in?ref_type=heads#L81-82. 

This env variable would be used to print the collector version: https://github.com/stackrox/collector/blob/master/collector/container/scripts/bootstrap.sh#L78.

I confirmed with

- quay.io/rhacs-eng/collector-slim:4.3.0
- registry.access.redhat.com/advanced-cluster-security/rhacs-collector-slim-rhel8@sha256:6afa0e1268e028a0afa7ad71531ed2f4bf12b5582984a513163840aedd6925d2

that it behaves like I described.

Maybe something to discuss with the @stackrox/collector-team colleagues - should this be set as an env variable on downstream as well? 
For RHTAP, we might have to find a different approach, to set the image label and env variable, e.g. move the `git describe` into the container.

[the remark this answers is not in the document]
tommartensen: In this PR, the bootstrap script would output `0.0.1-todo` for the collector version based on the ARG.

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,151 @@
...
+RUN ./install.sh && rm -f install.sh
+
+COPY collector/container/scripts/collector-wrapper.sh /usr/local/bin
+COPY collector/container/scripts/bootstrap.sh /
+COPY collector/LICENSE-kernel-modules.txt /kernel-modules/LICENSE
+COPY kernel-modules/MODULE_VERSION /kernel-modules/MODULE_VERSION.txt
+COPY --from=builder /build/cmake-build/collector/collector /usr/local/bin/collector
+COPY --from=builder /build/cmake-build/collector/self-checks /usr/local/bin/self-checks
+COPY --from=builder /THIRD_PARTY_NOTICES/ /THIRD_PARTY_NOTICES/
+COPY --from=builder /collector/NOTICE-sysdig.txt /THIRD_PARTY_NOTICES/sysdig
```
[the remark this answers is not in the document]
tommartensen: `THIRD_PARTY_NOTICES` do not seem to be part of the downstream image. I will check with the team.

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,151 @@
...
+# WITH_RHEL_RPMS controls for dependency installation, ie if they were already installed as RPMs.
+# Setting the value to empty will cause dependencies to be downloaded from repositories or accessed in submodules and compiled.
+# TODO(ROX-20651): Set ENV WITH_RHEL_RPMS=true when RHEL RPMs can be installed to enable hermetic builds.
+RUN ./builder/install/install-dependencies.sh \
+    && cmake -DDISABLE_PROFILING=ON -S "${SRC_ROOT_DIR}" -B "${CMAKE_BUILD_DIR}" \
+    && cmake --build "${CMAKE_BUILD_DIR}" --target all -- -j "$(nproc)" \
+    && (cd ${CMAKE_BUILD_DIR} && ctest -V) \
+    && strip --strip-unneeded "${CMAKE_BUILD_DIR}/collector/collector" "${CMAKE_BUILD_DIR}/collector/EXCLUDE_FROM_DEFAULT_BUILD/libsinsp/libsinsp-wrapper.so"
+
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.2
```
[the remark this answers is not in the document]
tommartensen: CentOS Stream *8* also worked, so applied that as builder image.

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,151 @@
...
+    io.k8s.display-name="collector-slim" \
+    io.openshift.tags="rhacs,collector,stackrox" \
+    maintainer="Red Hat, Inc." \
+    name="rhacs-collector-slim-rhel8" \
+    source-location="https://github.com/stackrox/collector" \
+    summary="Runtime data collection for the StackRox Kubernetes Security Platform" \
+    url="https://catalog.redhat.com/software/container-stacks/detail/60eefc88ee05ae7c5b8f041c" \
+    version=${COLLECTOR_VERSION} \
+    collector_version=${COLLECTOR_VERSION}
+    # TODO: These two labels only exist on upstream, keep or remove?
```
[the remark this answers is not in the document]
tommartensen: Done

collector/container/rhtap-slim.Dockerfile
tommartensen: Done

collector/container/rhtap-slim.Dockerfile
tommartensen: Updated PR with fixes from https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/merge_requests/2278. 
ARGs are now set in the correct stages.

collector/container/rhtap-slim.Dockerfile
tommartensen: Adapted to what was configured for downstream in https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/merge_requests/2278, implementing the suggestion of a named variable with a default.

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,151 @@
...
+    && cp -a /falcosecurity-libs falcosecurity-libs \
+    && cp -a /builder/third_party third_party \
+    && cp -a /kernel-modules/MODULE_VERSION kernel-modules/MODULE_VERSION \
+    && cp -a /CMakeLists.txt CMakeLists.txt
+
+# WITH_RHEL_RPMS controls for dependency installation, ie if they were already installed as RPMs.
+# Setting the value to empty will cause dependencies to be downloaded from repositories or accessed in submodules and compiled.
+# TODO(ROX-20651): Set ENV WITH_RHEL_RPMS=true when RHEL RPMs can be installed to enable hermetic builds.
+RUN ./builder/install/install-dependencies.sh \
+    && cmake -DDISABLE_PROFILING=ON -S "${SRC_ROOT_DIR}" -B "${CMAKE_BUILD_DIR}" \
```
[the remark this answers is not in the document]
tommartensen: Fixed with updates from https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/merge_requests/2278.

collector/container/rhtap-slim.Dockerfile
tommartensen: It is present in the upstream image: https://github.com/stackrox/collector/blob/master/collector/container/Dockerfile#L28-L29 🤔

collector/container/rhtap-slim.Dockerfile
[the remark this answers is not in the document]
tommartensen: It is present in the upstream image: https://github.com/stackrox/collector/blob/master/collector/container/Dockerfile#L28-L29 🤔

tommartensen:
@stackrox/collector-team 

Please have a look at this PR. It adds a new Dockerfile for builds on RHTAP. You should be able to build the same image with something like `docker build . -f collector/container/rhtap-slim.Dockerfile`. 

Example image: `quay.io/redhat-user-workloads/rh-acs-tenant/acs/collector-slim:on-pr-d44cd270c5de75adcaf81a7d72a6140cb4f472ce`. 

Biggest open questions we have:
- https://github.com/stackrox/collector/pull/1407#discussion_r1404230895 and https://github.com/stackrox/collector/pull/1407#discussion_r1404290865
- https://github.com/stackrox/collector/pull/1407#discussion_r1406122884 -> do we need to embed the version into the collector binary or is this just for the label?

.tekton/collector-slim-pull-request.yaml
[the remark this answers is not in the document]
tommartensen: Response from team on Slack about embedding Collector version:

> Nope, the version is not embedded anywhere into the Collector binary, and the closest to that is we use it as an image label.

https://redhat-internal.slack.com/archives/CFMQ5C2TT/p1702565504592349?thread_ts=1702393437.684719&cid=CFMQ5C2TT

.snyk
```diff
@@ -0,0 +1,16 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.25.0
+ignore: {}
+patch: {}
+exclude:
+  global:
+    - '.github/actions'
+    - 'collector/lib/'
```
erthalion, line 8: I'm not familiar with this tool, but isn't it to broad exclusion pattern?

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,152 @@
+ARG BUILD_DIR=/build
+ARG CMAKE_BUILD_DIR=${BUILD_DIR}/cmake-build
+
+# TODO(ROX-20312): we can't pin image tag or digest because currently there's no mechanism to auto-update that.
+# TODO(ROX-20651): Use RHEL/ubi base image when entitlement is solved.
+# RPMs requiring entitlement: bpftool, cmake-3.18.2-9.el8, elfutils-libelf-devel, tbb-devel, c-ares-devel, jq-devel
```
erthalion, line 6: Just out of curiosity, why this particular cmake version is called out? As far as I recall 3.20 is actually what's being used currently.

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,152 @@
...
+        valgrind \
+        wget \
+        which \
+        bpftool \
+    && dnf clean all
+
+ARG BUILD_DIR
+ARG SRC_ROOT_DIR=${BUILD_DIR}
+ARG CMAKE_BUILD_DIR
+# TODO(ROX-20240): CMAKE_BUILD_TYPE should probably not be Release for PR, normal branch builds
```
erthalion, line 57: This one is trickier than it looks like. We plan to change it to Debug everywhere for normal PRs, but for now it should be Release.

collector/container/rhtap-slim.Dockerfile
```diff
@@ -0,0 +1,152 @@
...
+ARG CMAKE_BUILD_TYPE=Release
+# Appends an argument to the driver download URL that is used for filtering alerts on missing kernels.
```
erthalion, line 59: I'm not entirely sure this alerts exists anymore, would probably makes sense to verify.

.snyk
tommartensen: [snyk.io](https://snyk.io/) is a required check in RHTAP. 
To pass the pipeline initially, we decided to exclude all violations for now and come back later to address them.

collector/container/rhtap-slim.Dockerfile
tommartensen: From https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/blob/rhacs-1.0-rhel-8/distgit/containers/rhacs-collector-slim/Dockerfile.in?ref_type=heads#L17

collector/container/rhtap-slim.Dockerfile
tommartensen: I added your comment to the issue: https://issues.redhat.com/browse/ROX-20240?focusedId=23767694&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23767694.

erthalion (approved):

PR #1413, opened 2023-11-09
Switches ansible dev inventory to acs-team-sandbox

Stringy:
## Description

Now that the stackrox-dev project is being deleted, we need to shift the development inventory to the new project. There may be a follow up in the coming weeks to do the same for stackrox-ci

## Testing Performed

Tested locally. This PR does not affect anything beyond local testing.


Molter73 (approved):
LGTM!

PR #1424, opened 2023-11-15
Avoid kernel modules for rhel 8.9

erthalion:
## Description

RHEL 8.9 with latest kernels contains backports for the once fixed issue for kernel modules [[1]]. Kernel modules are going to be deprecated soon anyway, so simply block their build for those versions with module version 2.3.0 & 2.4.0 (the last yet in build). 

[1]: https://github.com/falcosecurity/libs/issues/918


Stringy (approved):

ovalenti (approved):
This could be checked by adding the label to build legacy probes. But it should be fine.

PR #1441, opened 2023-11-21
ROX-20235: build collector-full on Konflux

tommartensen:
## Description

This PR adds the collector-full builds to Konflux.
The difference to `collector-slim` builds is addition of kernel modules downloaded from support packages and from the `driver-builds` image. 

Planned follow-ups: 
* Plan for ensuring collector support packages and drivers can be made available for fast stream (https://issues.redhat.com/browse/ROX-22244).
* Use build stages (https://issues.redhat.com/browse/ROX-22394) to reduce duplication between collector-slim and collector-full Dockerfiles.
 
## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

I tested the new image in the collector integration tests.
StackRox E2E tests were validated in https://github.com/stackrox/stackrox/pull/9614.


collector/container/rhtap.Dockerfile
```diff
@@ -0,0 +1,153 @@
+ARG BUILD_DIR=/build
```
tommartensen, line 1: This Dockerfile is very similar to `rhtap-slim.Dockerfile`:

```bash
$ diff collector/container/rhtap-slim.Dockerfile collector/container/rhtap.Dockerfile
100a101,104
> # Do NOT use follow_tag here, as we do not need or want collector to be rebuilt
> # with each drivers build (which may become very frequent)
> FROM brew.registry.redhat.io/rh-osbs/rhacs-drivers-build-rhel8:0.1.0 AS drivers-build
> # FROM registry-proxy.engineering.redhat.com/rh-osbs/rhacs-drivers-build-rhel8:0.1.0 AS drivers-build
132a137
> COPY --from=drivers-build /kernel-modules /kernel-modules
```

Stringy: Is it possible to parameterize the builds to reduce the dockerfile duplication?

tommartensen: This is to be addressed in https://issues.redhat.com/browse/ROX-22394.

collector/container/konflux.Dockerfile
```diff
@@ -0,0 +1,198 @@
...
+           -DDISABLE_PROFILING=${DISABLE_PROFILING} \
+           -DCOLLECTOR_APPEND_CID=${COLLECTOR_APPEND_CID} \
+           -DUSE_VALGRIND=${USE_VALGRIND} \
+           -DADDRESS_SANITIZER=${ADDRESS_SANITIZER} \
+           -DTRACE_SINSP_EVENTS=${TRACE_SINSP_EVENTS} && \
+    cmake --build ${CMAKE_BUILD_DIR} --target all -- -j "${NPROCS:-2}" && \
+    ctest -V --test-dir ${CMAKE_BUILD_DIR} && \
+    strip -v --strip-unneeded "${CMAKE_BUILD_DIR}/collector/collector" && \
+    if [[ -f "${CMAKE_BUILD_DIR}/collector/EXCLUDE_FROM_DEFAULT_BUILD/libsinsp/libsinsp-wrapper.so" ]]; then \
+        strip -v --strip-unneeded "${CMAKE_BUILD_DIR}/collector/EXCLUDE_FROM_DEFAULT_BUILD/libsinsp/libsinsp-wrapper.so"; fi
```
Stringy, line 99: I think this is unneeded now, since we are statically building collector.

Stringy (approved):
One minor comment, but otherwise the dockerfile and scripts LGTM!

As we've discussed previously, there are some big changes coming particularly for drivers and full images in collector and this will form the basis for a bunch of that work, so thanks for your efforts Tom!

I'll leave the pipeline code to those more in the know :)

collector/container/konflux.Dockerfile
tommartensen: (removed)

collector/container/konflux.Dockerfile
```diff
@@ -0,0 +1,196 @@
...
+    strip -v --strip-unneeded "${CMAKE_BUILD_DIR}/collector/collector"
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS support-packages-downloader
+
+WORKDIR /staging
+
+COPY kernel-modules/MODULE_VERSION kernel-modules/MODULE_VERSION
+COPY collector/container/scripts/download-support-package.sh download-support-package.sh
+
+RUN ./download-support-package.sh
```
msugakov, line 106: I think, the amount of effort to make this script run in a Tekton stage (see example [here](https://github.com/stackrox/scanner/blob/b29219d34d7ccff433aadfe7807e71b79b0830f8/.tekton/scanner-pull-request.yaml#L236-L254)) is the same as in Dockerfile.

The advantage of the Tekton stage is that we don't need to modify this part of Dockerfile when switching on hermetic builds. Since, I hope, the amount of effort to change isn't large, I suggest to make it as part of this PR.

collector/container/konflux.Dockerfile
```diff
@@ -0,0 +1,196 @@
...
+
+# Do NOT use follow_tag here, as we do not need or want collector to be rebuilt
+# with each drivers build (which may become very frequent)
+FROM brew.registry.redhat.io/rh-osbs/rhacs-drivers-build-rhel8:0.1.0 AS drivers-build
+# FROM registry-proxy.engineering.redhat.com/rh-osbs/rhacs-drivers-build-rhel8:0.1.0 AS drivers-build
```
msugakov, line 111: This line is redundant
```suggestion
```

collector/container/konflux.Dockerfile
```diff
@@ -0,0 +1,196 @@
...
+WORKDIR /staging
+
+COPY kernel-modules/MODULE_VERSION kernel-modules/MODULE_VERSION
+COPY collector/container/scripts/download-support-package.sh download-support-package.sh
+
+RUN ./download-support-package.sh
+
+# Do NOT use follow_tag here, as we do not need or want collector to be rebuilt
+# with each drivers build (which may become very frequent)
+FROM brew.registry.redhat.io/rh-osbs/rhacs-drivers-build-rhel8:0.1.0 AS drivers-build
```
msugakov, line 110: The comment is valid in CPaaS world where `follow_tag` is a thing. However, Konflux does not know it. Let's reword it a bit while having the main message the same.

```suggestion
# 0.1.0 is a floating tag and it's used intentionally to pick up the most recent downstream drivers build without
# having to routinely and frequently bump tags here.
FROM brew.registry.redhat.io/rh-osbs/rhacs-drivers-build-rhel8:0.1.0 AS drivers-build
```

collector/container/konflux.Dockerfile
```diff
@@ -0,0 +1,196 @@
...
+
+# TODO(ROX-20312): we can't pin image tag or digest because currently there's no mechanism to auto-update that.
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS unpacker
+
+RUN microdnf install -y unzip findutils
+WORKDIR /staging
+
+COPY --from=support-packages-downloader /staging/support-pkg.zip /staging/
+COPY kernel-modules/MODULE_VERSION MODULE_VERSION.txt
+# Creating this directory ensures the scratch build with dummy support-pkg.zip will not fail.
```
msugakov, line 121: The comment is meaningless in Konflux which does not have a notion of a "scratch build" (scratch build is executed during `check-patch`, in GitLab MR). Although I'm fine to keep the following `RUN mkdir -p ...`
```suggestion

```

collector/container/konflux.Dockerfile
```diff
@@ -0,0 +1,196 @@
...
+RUN mkdir -p "/staging/kernel-modules/$(cat MODULE_VERSION.txt)"
+# First, unpack upstream support package, only on x86_64
+RUN if [[ "$(uname -m)" == x86_64 ]]; then unzip support-pkg.zip ; fi
+# Fail non-scratch build if there were no drivers matching the module version.
+RUN if [[ "$(uname -m)" == x86_64 && "$(ls -A /staging/kernel-modules/$(cat MODULE_VERSION.txt))" == "" && "$(unzip -Z1 support-pkg.zip)" != "dummy-support-pkg" ]] ; then \
```
msugakov, line 126: Let's remove "scratch" build things here too.

```suggestion

RUN if [[ "$(uname -m)" == x86_64 && "$(ls -A /staging/kernel-modules/$(cat MODULE_VERSION.txt))" == "" ]] ; then \
```

collector/container/konflux.Dockerfile
```diff
@@ -0,0 +1,196 @@
...
+      >&2 echo "Did not find any kernel drivers for the module version $(cat MODULE_VERSION.txt) in the support package"; \
+      exit 1; \
+    fi
+
+# Next, import modules from downstream build, which take priority over upstream, on non-x86 architectures
+# TODO(ROX-13563): find a way to not have to separately pull in the support package and downstream-built drivers.
```
msugakov, line 132: https://redhat-internal.slack.com/archives/CFMQ5C2TT/p1707834313011919

collector/container/konflux.Dockerfile
```diff
@@ -0,0 +1,196 @@
...
+      cp -r /staging/downstream/. /staging/kernel-modules/ ; \
+    fi
+
+# Create destination for drivers.
+RUN mkdir /kernel-modules
+# Move files for the current version to /kernel-modules
+RUN find "/staging/kernel-modules/$(cat MODULE_VERSION.txt)/" -type f -exec mv -t /kernel-modules {} +
+# Fail the build if at the end there were no drivers matching the module version.
+RUN if [[ "$(ls -A /kernel-modules)" == "" && \
+        !("$(uname -m)" == x86_64 && "$(unzip -Z1 support-pkg.zip)" == "dummy-support-pkg") ]]; then \
```
msugakov, line 144: Scratch builds don't exist in Konflux so this check becomes
```suggestion
RUN if [[ "$(ls -A /kernel-modules)" == "" ]]; then \
```

Note that now it partially duplicates the check done above for x86_64.

The difference is that the above check verifies whether the _support package_ arrived without the matching drivers version. In case of x86_64, RHEL drivers (i.e. excluding non-Red Hat OS-es) should also come from `brew.registry.redhat.io/rh-osbs/rhacs-drivers-build-rhel8:0.1.0` which can mask the problem of mismatching (or empty) support package if the above check was removed.

This check here is most meaningful for ppc64le and s390x where drivers come _only_ from `rhacs-drivers-build-rhel8`.

Both checks are needed but there could probably be more clear and concise way to express them.

.tekton/collector-pull-request.yaml
```diff
@@ -0,0 +1,391 @@
...
+        resources:
+          requests:
+            storage: 5Gi
+      status: { }
+  - name: git-auth
+    secret:
+      secretName: '{{ git_auth_secret }}'
+
+  timeouts:
+    pipeline: 1h30m0s
```
msugakov, line 62: 1. This setting is absent from `collector-push.yaml`. Please add it there if it's meaningful.
2. If the timeout is deliberate, please add a short comment explaining it (e.g. we increase it because the default one is not enough to account for downloading the support package). Otherwise, if you introduced it as part of testing and the default one is also ok, I suggest removing it.

.tekton/collector-pull-request.yaml
```diff
@@ -0,0 +1,391 @@
...
+  - name: path-context
+    value: .
+  - name: revision
+    value: '{{revision}}'
+  - name: rebuild
+    value: 'true'
+  # TODO(ROX-20234): Enable hermetic builds
+  # - name: hermetic
+  #   value: "true"
+  # No language dependencies are required for central-db image.
```
msugakov, line 41: [nit] It's not `central-db`.
I believe, collector relies on git submodules. Do we get them checked out as part of `clone-repository` without any extra configuration on our side?

.tekton/collector-pull-request.yaml
```diff
@@ -0,0 +1,391 @@
+apiVersion: tekton.dev/v1beta1
```
msugakov, line 1: Please use `v1`
```suggestion
apiVersion: tekton.dev/v1
```

.tekton/collector-pull-request.yaml
```diff
@@ -0,0 +1,391 @@
...
+kind: PipelineRun
```
msugakov, line 2: [not about this line]

Regarding

> We need to follow up on building the driver-builds in Konflux as well eventually (<https://issues.redhat.com/browse/ROX-22244>).

in the PR description.

I meant ROX-22244 not for onboarding driver builds to Konflux. ROX-22244 was about a different problem. I expanded the ticket's description providing more context. Please have a look and let's follow discussions with the Collector team to jointly discover their plans and be in sync how this could match ACS Konflux onboarding.

msugakov (commented):
Made the first pass but could have missed something. Will check again after you iterate on comments.

collector/container/konflux.Dockerfile
```diff
@@ -0,0 +1,196 @@
...
+    cmake -S ${SRC_ROOT_DIR} -B ${CMAKE_BUILD_DIR} \
+           -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} \
+           -DDISABLE_PROFILING=${DISABLE_PROFILING} \
+           -DCOLLECTOR_APPEND_CID=${COLLECTOR_APPEND_CID} \
+           -DUSE_VALGRIND=${USE_VALGRIND} \
+           -DADDRESS_SANITIZER=${ADDRESS_SANITIZER} \
+           -DTRACE_SINSP_EVENTS=${TRACE_SINSP_EVENTS} && \
+    cmake --build ${CMAKE_BUILD_DIR} --target all -- -j "${NPROCS:-2}" && \
+    ctest -V --test-dir ${CMAKE_BUILD_DIR} && \
+    strip -v --strip-unneeded "${CMAKE_BUILD_DIR}/collector/collector"
```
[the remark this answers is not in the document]
tommartensen: According to https://github.com/stackrox/collector/pull/1441#discussion_r1484450706, the libinsp wrapper handling is not required anymore. 
I have prepared another PR for slim to remove it and rename from RHTAP -> Konflux: https://github.com/stackrox/collector/pull/1549

.tekton/collector-pull-request.yaml
tommartensen: Removed. There was a time where the pipeline regularly took more than 1 hour, but that has not happened in the last 7 days.

.tekton/collector-pull-request.yaml
tommartensen: We do check out the submodules by default: https://github.com/redhat-appstudio/build-definitions/blob/main/task/git-clone/0.1/git-clone.yaml#L28-L31

collector/container/konflux.Dockerfile
tommartensen: Applied your suggestion and simplified in https://github.com/stackrox/collector/pull/1441/commits/90535d5917c5ef6ee3c4ad4ea9acbfa945b33985 .

collector/container/konflux.Dockerfile
```diff
@@ -96,34 +96,24 @@ RUN ./builder/install/install-dependencies.sh && \
...
-# FROM registry-proxy.engineering.redhat.com/rh-osbs/rhacs-drivers-build-rhel8:0.1.0 AS drivers-build
 
 # TODO(ROX-20312): we can't pin image tag or digest because currently there's no mechanism to auto-update that.
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS unpacker
 
 RUN microdnf install -y unzip findutils
 WORKDIR /staging
 
-COPY --from=support-packages-downloader /staging/support-pkg.zip /staging/
+COPY staging/support-pkg.zip /staging/
```
msugakov, line 109: Would it be more straightforward to just copy `staging` (with everything inside) to `/` (so that the resulting paths are `/staging/...`)?

Asking because this `COPY staging/support-pkg.zip /staging/` looks a bit odd but I don't really know if my suggestion would look better.

.tekton/collector-pull-request.yaml
msugakov: This makes sense, thanks!

I'm a bit torn though if we should provide explicit `submodules`=`true` in `clone-repository` params
- to make it visible for readers that we request checking out submodules,
- to prevent from a case when the default `submodules` value suddenly becomes `false` and our builds break,
- to catch early a case when `submodules` gets renamed which may come with other behavior change.

WDYT?

tommartensen: 1) visibility -> agree
2+3) new parameter or default values would require a new task version according to https://github.com/redhat-appstudio/build-definitions/?tab=readme-ov-file#versioning, with the automated bot linking to the migration notes.

Specifying the parameter would only help for visibility. Let's try it.

collector/container/scripts/download-support-package.sh
```diff
@@ -0,0 +1,52 @@
...
+    # and I do not want to risk downloading kernel drivers from untrusted location.
+    MODULE_VERSION=$(cat kernel-modules/MODULE_VERSION)
+    SUPPORT_PKG_VERSION="20240209164910"
+    support_pkg="https://cdn.stackrox.io/collector/support-packages/x86_64/${MODULE_VERSION}/support-pkg-${MODULE_VERSION}-${SUPPORT_PKG_VERSION}.zip"
+
+    zip_file="$(basename "${support_pkg}")"
+
+    # This downloads the support package with all collector kernel drivers (probes) built upstream.
+    # Eventually this needs to go away and we should build kernel drivers downstream.
+    # See https://stack-rox.atlassian.net/browse/RS-230
```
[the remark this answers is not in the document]
tommartensen: https://github.com/stackrox/collector/pull/1441/commits/e90c5360582c0df2f36b13bec46b35f8455480b8

(RS-230 -> ROX-11373)

collector/container/konflux.Dockerfile
tommartensen: Even though it creates a coupling between the download script and the Dockerfile based on the filename, I prefer to be explicit here to avoid copying other contents from `staging`. There is none today, but to future proof the script.

collector/container/konflux.Dockerfile
tommartensen: Removed comment following https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/merge_requests/2600. 

Since collector team will be taking care of further midstream & Konflux changes with the removal of eBPF, I wouldn't want to introduce any more changes here.

.tekton/collector-pull-request.yaml
tommartensen: updated PR description. The additional context makes it clear what ROX-22244 should achieve and we can start with this after MVP of MVP

.tekton/collector-pull-request.yaml
```diff
@@ -224,6 +227,22 @@ spec:
       - name: source
         workspace: workspace
 
+    - name: fetch-support-package
```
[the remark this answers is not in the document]
tommartensen: https://github.com/stackrox/collector/pull/1441/commits/6f3179481e5f06b73b414a989c9fe33dd0f5e981

collector/container/konflux.Dockerfile
msugakov: That's fine.

.tekton/collector-pull-request.yaml
```diff
@@ -238,13 +245,6 @@ spec:
...
               #!/usr/bin/env bash
 
-              set -x
-
-              pwd
-              ls -lisa
-
-              ls -lisa "$(workspaces.source.path)"
-              ls -lisa "$(workspaces.source.path)/source"
               cd "$(workspaces.source.path)/source"
```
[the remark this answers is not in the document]
tommartensen: That was from local testing, good catch. fixed in https://github.com/stackrox/collector/pull/1441/commits/53e797a6c490a1bedbd38410ded8d552b3e1122b

collector/container/scripts/download-support-package.sh
```diff
@@ -30,6 +30,8 @@ main() {
 
     zip_file="$(basename "${support_pkg}")"
 
+    cd "${TARGET_DIR}"
```
msugakov, line 33: WDYT if the `TARGET_DIR` **is** `$(workspaces.source.path)/source/staging` and you both create and enter it here?

```suggestion
    mkdir -p "${TARGET_DIR}"
    cd "${TARGET_DIR}"
```

Then, at the end of the script, you just rename the zip file.

There's no need to worry about interrupted partial downloads because (a) sha256sum check would not let partial file go through and besides (b) no-zero exit code from `curl` or anywhere of this script will abort the Tekton pipline with error.

I bring this up because it feels _a bit arbitrary_ that files are first downloaded to `$(workspaces.source.path)/source` to be later moved to `$(workspaces.source.path)/source/staging`.
From the perspective of this script's code it will be just cleaner if the caller provides it the ultimate directory path where everything happens and the script does not need to worry that at some point `/staging/` should be appended to the path.

collector/container/scripts/download-support-package.sh
```diff
@@ -45,10 +47,10 @@ main() {
     sha256sum -c "${zip_file}.sha256"
 
     # Rename the support package so the docker build can find it in the same place every build.
-    mkdir -p "${TARGET_DIR}"
-    mv "${zip_file}" "${TARGET_DIR}/support-pkg.zip"
+    mkdir -p "staging"
+    mv "${zip_file}" "staging/support-pkg.zip"
 
-    echo "Saved support package as ${TARGET_DIR}/support-pkg.zip"
+    echo "Saved support package as ${TARGET_DIR}/staging/support-pkg.zip"
```
msugakov, line 53: Here, it would be
```suggestion
    mv "${zip_file}" "support-pkg.zip"

    echo "Saved support package as ${TARGET_DIR}/support-pkg.zip"
```

.tekton/collector-pull-request.yaml
```diff
@@ -245,9 +245,8 @@ spec:
             script: |
               #!/usr/bin/env bash
 
-              cd "$(workspaces.source.path)/source"
               collector/container/scripts/download-support-package.sh \
-                "$(workspaces.source.path)/source/staging"
+                "$(workspaces.source.path)/source"
```
msugakov, line 249: Would be:
```suggestion
                "$(workspaces.source.path)/source/staging"
```

.tekton/collector-push.yaml
```diff
@@ -245,9 +245,8 @@ spec:
             script: |
               #!/usr/bin/env bash
 
-              cd "$(workspaces.source.path)/source"
               collector/container/scripts/download-support-package.sh \
-                "$(workspaces.source.path)/source/staging"
+                "$(workspaces.source.path)/source"
```
msugakov, line 249: Also
```suggestion
                "$(workspaces.source.path)/source/staging"
```

.tekton/collector-pull-request.yaml
```diff
@@ -245,8 +245,9 @@ spec:
             script: |
               #!/usr/bin/env bash
 
-              collector/container/scripts/download-support-package.sh \
-                "$(workspaces.source.path)/source"
+              "$(workspaces.source.path)/source/collector/container/scripts/download-support-package.sh" \
+                "$(workspaces.source.path)/source/staging" \
+                "$(cat $(workspaces.source.path)/source/kernel-modules/MODULE_VERSION)"
```
msugakov, line 250: Cosmetics:
```suggestion
                "$(cat "$(workspaces.source.path)/source/kernel-modules/MODULE_VERSION")"
```

.tekton/collector-push.yaml
```diff
@@ -245,8 +245,9 @@ spec:
             script: |
               #!/usr/bin/env bash
 
-              collector/container/scripts/download-support-package.sh \
-                "$(workspaces.source.path)/source"
+              "$(workspaces.source.path)/source/collector/container/scripts/download-support-package.sh" \
+                "$(workspaces.source.path)/source/staging" \
+                "$(cat $(workspaces.source.path)/source/kernel-modules/MODULE_VERSION)"
```
msugakov, line 250:
```suggestion
                "$(cat "$(workspaces.source.path)/source/kernel-modules/MODULE_VERSION")"
```

msugakov (approved):
Ah, sorry, I did not realize `MODULE_VERSION` file is also a dependency of the script. LGTM.

PR #1448, opened 2023-11-29
ppc64le core-bpf testing and PR testing

Stringy:
## Description

Should enable core-bpf testing for ppc64le. Once it is working/failing, I'll rebase it over the vanilla branch. 

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


.github/workflows/collector-builder.yml
```diff
@@ -11,11 +11,12 @@ on:
     outputs:
       collector-builder-tag:
         description: The builder tag used by the build
-        value: ${{ jobs.build-builder-image.outputs.collector-builder-tag || 'master' }}
+        value: ${{ jobs.build-builder-image.outputs.collector-builder-tag || '3.16.x-195-g8f32e71fad' }}
```
erthalion, line 14: What was this image exactly? And would it be dropped before merging?

.github/workflows/collector-builder.yml
```diff
@@ -110,6 +143,24 @@ jobs:
             -e @'${{ github.workspace }}/ansible/secrets.yml' \
             ansible/ci-build-builder.yml
 
+      - name: Build s390x images
```
erthalion, line 146: This feels a bit weird, is it easier to build s390x images in a separate step, not as part of the previous one?

.github/workflows/collector-slim.yml
```diff
@@ -42,50 +43,117 @@ jobs:
...
+            -i ansible/ci \
+            -e arch='${{ matrix.arch }}' \
+            -e build_hosts='job_id_${{ env.JOB_ID }}' \
             -e @'${{ github.workspace }}/ansible/secrets.yml' \
             ansible/ci-build-collector.yml
         env:
-          ANSIBLE_CONFIG: ansible/ansible.cfg
+          VM_TYPE: rhel-s390x
+
+      - name: Save CMake cache
```
erthalion, line 146: Will this cache be used automatically, or does it have to be explicitly included? It's probably a good idea to add a condition to allow running without cache if specified via a label.

ansible/ci-build-collector-full.yml
```diff
@@ -19,7 +19,7 @@
     - name: Build full image
       when:
         - build_full_image
-        - arch != 'arm64'
+        - arch != 'arm64' and arch != 'ppc64le' and arch != 's390x'
```
erthalion, line 22: This is the part where we temporarily disable full builds for multi-arch, right?

ansible/ci-build-builder.yml
```diff
@@ -1,16 +1,28 @@
...
 
   tasks:
+    - name: Clone repository
+      ansible.builtin.git:
+        repo: https://github.com/stackrox/collector
+        dest: "{{ collector_root }}"
+        version: "{{ collector_git_sha }}"
+        refspec: "+{{ collector_git_ref | replace('refs/', '') }}"
+        recursive: true
+      when: arch == "s390x"
```
erthalion, line 21: Not sure I follow, why to clone only for s390x?

erthalion (commented):
Thanks, looks good. I've left few commentaries, but suggest to follow up on them in a separate PR. Otherwise not sure what's your plan to proceed -- s390x tests seems to pass, but few others are failing, is that expected?

.github/workflows/collector-builder.yml
Stringy: This was taken from mauro's original vanilla PR and will be dropped before merging.

.github/workflows/collector-builder.yml
Stringy: For now, yeah it's a little simpler, just because we can force the previous step to use localhost only, but this step will use the entire inventory (which is only one VM at the moment)

When I implement full multi-arch native builds I'll hopefully combine these steps

.github/workflows/collector-slim.yml
Stringy: This was just for debugging the build, I'll remove it.

ansible/ci-build-collector-full.yml
Stringy: yeah, it is. I'll remove before merging.

ansible/ci-build-builder.yml
Stringy: Because it's the only platform (currently) that is building on a separate machine to the GHA runner. All the other platforms can build from the existing checked out source

Stringy:
> s390x tests seems to pass, but few others are failing, is that expected?

Seems to have been an SSH key mix up following my ansible changes; I've fixed that now so this run should pass 🤞

erthalion (approved):

PR #1454, opened 2023-12-04
Enable debug stdout callback in ansible when GHA is in debug mode

Stringy:
## Description

With the move to the selective callback in ansible, we sometimes lose some valuable information about certain error conditions. This PR will switch the callback to `debug` when we re-run the workflow with debug logging in GHA.


erthalion (approved):
Looks good, thanks!

PR #1465, opened 2023-12-11
Integration test for gperftool api

erthalion:
## Description

Make sure gperftool api is covered with an integration test. Note, that it's going to be performed only on supported platforms.

## Checklist

**Automated testing**
 ~- [ ] Added unit tests~
  - [x] Added integration tests

## Testing Performed

Local testing

### DIFF
--- a/integration-tests/suites/gperftools.go
+++ b/integration-tests/suites/gperftools.go
@@ -1,31 +1,19 @@
 package suites
 
 import (
-	"bytes"
-	"fmt"
 	"net/http"
 	"strings"
 	"time"
 
-	"golang.org/x/sys/unix"
+	"github.com/stackrox/collector/integration-tests/suites/common"
 )
 
 type GperftoolsTestSuite struct {
 	IntegrationTestSuiteBase
 }
 
-// IsArchSupported: at the moment only x86_64 builds have gperftools
-func (s *GperftoolsTestSuite) IsArchSupported() (bool, string) {
-	u := unix.Utsname{}
-	unix.Uname(&u)
-
-	// Exclude null bytes from comparison
-	arch := string(bytes.Trim(u.Machine[:], "\x00"))
-	return arch == "x86_64", arch
-}
-
 func (s *GperftoolsTestSuite) SetupSuite() {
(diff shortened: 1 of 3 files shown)


integration-tests/suites/gperftools.go
```diff
@@ -0,0 +1,78 @@
...
+// * fetch the profile
+//
+// NOTE: The test will only be performed on supported architectures, as defined
+// by IsArchSupported.
+func (s *GperftoolsTestSuite) TestFetchHeapProfile() {
+	heap_api_url := "http://localhost:8080/profile/heap"
+	data_type := "application/x-www-form-urlencoded"
+
+	if ok, arch := s.IsArchSupported(); !ok {
+		fmt.Printf("[WARNING]: Skip GperftoolsTestSuite on %s\n", arch)
```
Stringy, line 55: You can do `s.T().Skip("[WARNING]: skip GperftoolsTestSuite on ", arch)`

integration-tests/suites/gperftools.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// IsArchSupported: at the moment only x86_64 builds have gperftools
+func (s *GperftoolsTestSuite) IsArchSupported() (bool, string) {
+	u := unix.Utsname{}
+	unix.Uname(&u)
+
+	// Exclude null bytes from comparison
+	arch := string(bytes.Trim(u.Machine[:], "\x00"))
+	return arch == "x86_64", arch
```
Stringy, line 24: I wonder if we can make this a bit more general purpose for use elsewhere in the tests

```golang

// suites/common/utils.go
import "slices"

func ArchSupported(supported ...string) (bool, string) {
        u := unix.Utsname{}
        unix.Uname(&u)

        arch := string(bytes.Trim(u.Machine[:], "\x00"))
        return slices.Contains(supported, arch), arch
}
```

integration-tests/suites/gperftools.go
erthalion: Haven't found anything immediately about how to do this in setup/teardown.

Stringy (approved):

PR #1473, opened 2023-12-15
Make tests verbose about startup failures

erthalion:
## Description

In the past we've got situations when the CI integration tests reports were not extensive enough to clarify what's going on, and we were launching a full-blown debugging process just to find out that the probes were missing.

To make initialization issues more visible, make integration tests verbose about them:

* Introduce a HEALTHCHECK for the Collector docker image, based on the /ready Civet API. This command would be ignored on K8S, but we could use it on CI for containers filtering.

* Wait for Collector container to become healthy first before proceeding with the rest of tests. This will make it immediately clear if the test is failing due to missing probes.

* Make self-checks verification a hard failure. Currently, we wait for the self-checks, but do not report if they're missing. This change will make it clear if something is badly broken to the extent that we don't receive any events.

## Checklist

## Testing Performed

Local testing, but CI is sufficient.


Stringy (approved):
LGTM! This is a great idea, and should help a lot

PR #1476, opened 2023-12-19
Configurable sinsp buffer

erthalion:
## Description

Make sinsp event buffer related parameters (how many buffers to allocate and of what size) configurable via Collector config. The default values are the same as before, but it makes memory utilization more flexible. Taking into account the difference between on-demand allocation for ebpf and right-away allocation for core_bpf, this can help to address situations, when users are unhappy about Collector memory consumption, e.g. one can lower it down significantly by reducing number of buffers or their size -- the trade-off being of course higher latencies due to congestion on buffers.

## Checklist

## Testing Performed

Local testing.


ovalenti (approved):
Looks good to me !

PR #1480, opened 2023-12-21
Skip health check for CPaaS workflow

erthalion:
## Description

Currently, we use health check to verify Collector has started properly, but not all the pipelines have the health check enabled in the image. It's fine, because we want to use it only for testing, so make the integration tests smarter to handle situations where there is no health check in the image.

## Checklist

## Testing Performed

Local testing. CI is sufficient.

### DIFF
--- a/integration-tests/suites/base.go
+++ b/integration-tests/suites/base.go
@@ -89,18 +89,20 @@ func (s *IntegrationTestSuiteBase) StartCollector(disableGRPC bool, options *com
 	} else {
 		fmt.Println("No HealthCheck found, do not wait for collector to become healthy")
 
-		// No way to figure out when all the services up and running, so give
-		// Collector a bit of time to spin up GRPC connection.
-		time.Sleep(10 * time.Second)
+		// No way to figure out when all the services up and running, skip this
+		// phase.
 	}
 
-	// Self-check process is not going to be sent via GRPC, instead create at
-	// least one canary process to make sure everything is fine.
-	_, err = s.execContainer("collector", []string{"echo"})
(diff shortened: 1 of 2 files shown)


integration-tests/suites/base.go
```diff
@@ -69,13 +69,26 @@ func (s *IntegrationTestSuiteBase) StartCollector(disableGRPC bool, options *com
...
+			"collector",
+			s.Collector().ContainerID,
+			defaultWaitTickSeconds, 1*time.Minute)
+		s.Require().NoError(err)
+	} else {
+		fmt.Println("No HealthCheck found, do not wait for collector to become healthy")
+	}
 
 	// wait for self-check process to guarantee collector is started
 	selfCheckOk := s.Sensor().WaitProcessesN(s.Collector().ContainerID, 30*time.Second, 1)
```
ovalenti, line 94: Now that we check for the success of this phase, it fails quite a significant amount of times. It may require a longer timeout.

ovalenti (dismissed):
The code changes in this PR look fine to me, but the previous changes seem to make the tests failures more likely.
Overall, this is a good thing, because you might have spotted a source for a lot of flakes (including the "I did not expected that many processes" failure).

integration-tests/suites/base.go
ovalenti, line 94:
```suggestion
	selfCheckOk := s.Sensor().WaitProcessesN(s.Collector().ContainerID, 180*time.Second, 1)
```
I would be curious whether this fixes the issue.

Stringy (approved):
LGTM!

PR #1492, opened 2024-01-09
Build fixes for linux 6.7

ovalenti:
## Description

Some changes in the inode structure introduced in 6.7. We cherry-pick the fix, and block previous probe
versions (2.6&2.7) on 6.7 and up.

There is an issue with the size of some programs in the ebpf probe as well.

Those changes should end-up in the 3.18 branch, as we shall overwrite them in master with the next "big falco pull"

## Checklist


Stringy (dismissed):

Molter73 (approved):
LGTM!

PR #1498, opened 2024-01-11
Fixes BPF loading errors by using Host path for attachment

Stringy:
## Description

When attaching programs, falco uses the files in `/sys/kernel/debug/tracing/events`. On some platforms, this is only accessible via our own host mount `/host/sys/...`

This PR (and the associated PR in falco) make sure that this is taken into account when attaching tracepoint programs.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Tested on an OCP 4.11 (RHCOS, podman) cluster, with and without host path.


ovalenti (approved):
Well, that is a good catch :fist_oncoming:

PR #1506, opened 2024-01-17
Only create s390x build VMs when theyre needed

Stringy:
## Description

s390x build VMs were being created always, even when multi-arch builds were not requested as part of a PR. This simple makes sure that VM creation abides by the labels.


ovalenti (approved):
Looks good to me, thanks !

The CI failure is on rhel/amd64, so it does not seem related.

Maybe running the PR with the multiarch label would also check that the VM is successfully created.

PR #1507, opened 2024-01-17
Optional local builder access through SSH

ovalenti:
## Description

Make it easier to create a builder container that can be connected locally through SSH.

- Builder runs sshd as entrypoint.
- The port is exposed only when LOCAL_SSH_PORT is set.

This is intended to ease the use of the builder in dev environments where SSH is used
to connect to the builder container.


Stringy (approved):
LGTM! feels like a nice simplification with some added functionality 👌

PR #1508, opened 2024-01-18
Disable trusted exepath

erthalion:
## Description

Vanilla Falco has introduced more rigorous exepath logic which resolves
symlinks when needed. By itself it's a welcomed improvement, but turns
out it upsets ACS builtin policies that often rely on the original
exepath implementation. Temporary disable trusted exepath until we will
figure out how to migrate policies to use this information.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Local testing with an image triggering violations.


kernel-modules/MODULE_VERSION
```diff
@@ -1 +1 @@
-2.9.0-rc2
+2.9.0-rc4
```
Stringy, line 1: why do we need a module version bump for this?

builder/Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN dnf -y update \
         binutils-devel \
         bison \
         ca-certificates \
-        clang-15.0.7 \
+        clang-17.0.6 \
```
Stringy, line 18: since this is fixed in master now I think we need to rebase

kernel-modules/MODULE_VERSION
erthalion: I was under the impression it wont pickup Falco changes otherwise. You're saying it's not correct?

Stringy: a bump is only necessary if there is a change to the drivers, so since this is just userspace we don't need to :)

Stringy (approved):

PR #1509, opened 2024-01-19
Jv ebpf deprecation cleanup

JoukoVirtanen:
## Description

Sets the default collection method to CORE_BPF for collector and the integration tests. Fixed unit tests that broke as a result of changing the default collection method to CORE_BPF. Also removed the hardfail config option for CORE_BPF. The code now behaves as though it is always true.

## Checklist
- [x] Ran integration tests locally
~~- [ ] Updated documentation accordingly~~

**Automated testing**
  ~~- [ ] Added unit tests~~
  ~~- [ ] Added integration tests~~

This changes the default collection method so testing it is not needed.

## Testing Performed

TODO(replace-me)

### DIFF
--- a/collector/test/HostHeuristicsTest.cpp
+++ b/collector/test/HostHeuristicsTest.cpp
@@ -52,33 +52,18 @@ class MockHostInfoHeuristics : public HostInfo {
   MOCK_METHOD0(GetDistro, std::string&());
 };
 
-// Note that in every test below, ProcessHostHeuristics will be called first
-// with creation of a config mock, which is somewhat annoying, but doesn't
-// cause any serious issues.
-class MockCollectorConfig : public CollectorConfig {
- public:
-  MockCollectorConfig(CollectorArgs* collectorArgs)
-      : CollectorConfig(collectorArgs){};
-
-  MOCK_CONST_METHOD0(UseEbpf, bool());
-
-  void SetCollectionMethod(CollectionMethod cm) {
-    if (host_config_.HasCollectionMethod()) {
-      host_config_.SetCollectionMethod(cm);
-    }
-    collection_method_ = cm;
-  }
-};
-
 TEST(HostHeuristicsTest, TestS390XRHEL84) {
   MockS390xHeuristics s390xHeuristics;
   MockHostInfoHeuristics host;
   KernelVersion version = KernelVersion("4.18.0-305.88.1.el8_4.s390x", "", "s390x");
   CollectorArgs* args = CollectorArgs::getInstance();
-  MockCollectorConfig config(args);
+  int argc = 3;
+  const char* argv[] = {"collector", "--collection-method", "ebpf"};
+  int exitCode = 0;
+  args->parse(argc, const_cast<char**>(argv), exitCode);
+  CollectorConfig config(args);
   HostConfig hconfig;
 
-  config.SetCollectionMethod(CollectionMethod::EBPF);
   hconfig.SetCollectionMethod(CollectionMethod::EBPF);
   EXPECT_CALL(host, GetKernelVersion()).WillOnce(Return(version));
 
@@ -92,10 +77,13 @@ TEST(HostHeuristicsTest, TestARM64Heuristic) {
   MockHostInfoHeuristics host;
   KernelVersion version = KernelVersion("6.5.5-200.fc38.aarch64", "", "aarch64");
   CollectorArgs* args = CollectorArgs::getInstance();
-  MockCollectorConfig config(args);
+  int argc = 3;
+  const char* argv[] = {"collector", "--collection-method", "ebpf"};
+  int exitCode = 0;
+  args->parse(argc, const_cast<char**>(argv), exitCode);
+  CollectorConfig config(args);
   HostConfig hconfig;
 
-  config.SetCollectionMethod(CollectionMethod::EBPF);
   hconfig.SetCollectionMethod(CollectionMethod::EBPF);
   EXPECT_CALL(host, GetKernelVersion()).WillOnce(Return(version));
 
--- a/collector/test/HostHeuristicsTest.cpp
+++ b/collector/test/HostHeuristicsTest.cpp
@@ -52,18 +52,27 @@ class MockHostInfoHeuristics : public HostInfo {
   MOCK_METHOD0(GetDistro, std::string&());
 };
 
+class MockCollectorConfig : public CollectorConfig {
+ public:
+  MockCollectorConfig()
+      : CollectorConfig(){};
+
+  void SetCollectionMethod(CollectionMethod cm) {
+    if (host_config_.HasCollectionMethod()) {
+      host_config_.SetCollectionMethod(cm);
+    }
+    collection_method_ = cm;
+  }
+};
+
 TEST(HostHeuristicsTest, TestS390XRHEL84) {
   MockS390xHeuristics s390xHeuristics;
   MockHostInfoHeuristics host;
   KernelVersion version = KernelVersion("4.18.0-305.88.1.el8_4.s390x", "", "s390x");
-  CollectorArgs* args = CollectorArgs::getInstance();
-  int argc = 3;
-  const char* argv[] = {"collector", "--collection-method", "ebpf"};
-  int exitCode = 0;
-  args->parse(argc, const_cast<char**>(argv), exitCode);
-  CollectorConfig config(args);
+  MockCollectorConfig config;
   HostConfig hconfig;
 
+  config.SetCollectionMethod(CollectionMethod::EBPF);
   hconfig.SetCollectionMethod(CollectionMethod::EBPF);
   EXPECT_CALL(host, GetKernelVersion()).WillOnce(Return(version));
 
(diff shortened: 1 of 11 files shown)


collector/test/HostHeuristicsTest.cpp
```diff
@@ -52,33 +52,18 @@ class MockHostInfoHeuristics : public HostInfo {
...
-  MOCK_CONST_METHOD0(UseEbpf, bool());
-
-  void SetCollectionMethod(CollectionMethod cm) {
-    if (host_config_.HasCollectionMethod()) {
-      host_config_.SetCollectionMethod(cm);
-    }
-    collection_method_ = cm;
-  }
-};
-
```
JoukoVirtanen, line 72: MockCollectorConfig is not used anymore, so this is not needed.

collector/test/HostHeuristicsTest.cpp
```diff
@@ -52,33 +52,18 @@ class MockHostInfoHeuristics : public HostInfo {
   MOCK_METHOD0(GetDistro, std::string&());
 };
 
-// Note that in every test below, ProcessHostHeuristics will be called first
-// with creation of a config mock, which is somewhat annoying, but doesn't
-// cause any serious issues.
```
JoukoVirtanen, line 57: This did create a serious issue when the default collection method was changed to CORE_BPF. This is because ProcessHostHeuristics calls Process, which runs the following when the collection method is CORE_BPF

```
      if (!host.HasBPFRingBufferSupport()) {
        CLOG(FATAL) << "Missing RingBuffer support, core_bpf is not available. "
                    << "HINT: Change collection method to eBPF with collector.collectionMethod=EBPF.";
      }
```

collector/test/HostHeuristicsTest.cpp
```diff
@@ -52,33 +52,18 @@ class MockHostInfoHeuristics : public HostInfo {
...
-};
-
 TEST(HostHeuristicsTest, TestS390XRHEL84) {
   MockS390xHeuristics s390xHeuristics;
   MockHostInfoHeuristics host;
   KernelVersion version = KernelVersion("4.18.0-305.88.1.el8_4.s390x", "", "s390x");
   CollectorArgs* args = CollectorArgs::getInstance();
-  MockCollectorConfig config(args);
+  int argc = 3;
+  const char* argv[] = {"collector", "--collection-method", "ebpf"};
```
JoukoVirtanen, line 61: The collection method has to be set here, because `CollectorConfig config(args)` results in a call to `ProcessHostHeuristics` and in turn `Process`. If the collection method is set to the default `CORE_BPF` in  `Process` a fatal error will be triggered.

collector/test/HostHeuristicsTest.cpp
```diff
@@ -52,33 +52,18 @@ class MockHostInfoHeuristics : public HostInfo {
...
-    collection_method_ = cm;
-  }
-};
-
 TEST(HostHeuristicsTest, TestS390XRHEL84) {
   MockS390xHeuristics s390xHeuristics;
   MockHostInfoHeuristics host;
   KernelVersion version = KernelVersion("4.18.0-305.88.1.el8_4.s390x", "", "s390x");
   CollectorArgs* args = CollectorArgs::getInstance();
-  MockCollectorConfig config(args);
```
JoukoVirtanen, line 78: There is no need for `MockCollectorConifg` now that the collection method is set using arguments.

collector/test/HostHeuristicsTest.cpp
```diff
@@ -92,10 +77,13 @@ TEST(HostHeuristicsTest, TestARM64Heuristic) {
   MockHostInfoHeuristics host;
   KernelVersion version = KernelVersion("6.5.5-200.fc38.aarch64", "", "aarch64");
   CollectorArgs* args = CollectorArgs::getInstance();
-  MockCollectorConfig config(args);
+  int argc = 3;
+  const char* argv[] = {"collector", "--collection-method", "ebpf"};
+  int exitCode = 0;
+  args->parse(argc, const_cast<char**>(argv), exitCode);
+  CollectorConfig config(args);
```
JoukoVirtanen, line 84: Same as above.

collector/test/NetworkStatusNotifierTest.cpp
```diff
@@ -181,7 +173,12 @@ class NetworkConnectionInfoMessageParser {
 /* Simple validation that the service starts and sends at least one event */
 TEST(NetworkStatusNotifier, SimpleStartStop) {
   bool running = true;
-  CollectorConfig config_(0);
+  CollectorArgs* args = CollectorArgs::getInstance();
+  int argc = 3;
+  const char* argv[] = {"collector", "--collection-method", "ebpf"};
+  int exitCode = 0;
+  args->parse(argc, const_cast<char**>(argv), exitCode);
+  CollectorConfig config_(args);
```
JoukoVirtanen, line 181: Same as for the HostHeuristicsTests.

collector/test/NetworkStatusNotifierTest.cpp
```diff
@@ -64,15 +65,6 @@ class Semaphore {
...
 
-class MockCollectorConfig : public collector::CollectorConfig {
- public:
-  MockCollectorConfig() : collector::CollectorConfig(0) {}
-
-  void DisableAfterglow() {
-    enable_afterglow_ = false;
-  }
-};
-
```
JoukoVirtanen, line 75: This is not needed anymore as afterglow is disabled using an environment variable.

collector/test/NetworkStatusNotifierTest.cpp
```diff
@@ -241,7 +238,14 @@ TEST(NetworkStatusNotifier, SimpleStartStop) {
...
   bool running = true;
-  MockCollectorConfig config;
+  setenv("ROX_AFTERGLOW_PERIOD", "0", 1);
+  CollectorArgs* args = CollectorArgs::getInstance();
+  int argc = 3;
+  const char* argv[] = {"collector", "--collection-method", "ebpf"};
+  int exitCode = 0;
+  args->parse(argc, const_cast<char**>(argv), exitCode);
+  CollectorConfig config(args);
+  unsetenv("ROX_AFTERGLOW_PERIOD");
```
JoukoVirtanen, line 248: Same as the other cases. Afterglow is disabled using the ROX_AFTERGLOW_PERIOD environment variable instead of using a mock.

collector/test/HostHeuristicsTest.cpp
erthalion: Is it not possible to mock out `Process` as well? I'm not sure it's the best solution to go through config setup directly, because it makes unit tests less "unit", in the sense that they now less isolated and a bit dependent on the outside world (in the form of env vars).

collector/test/HostHeuristicsTest.cpp
erthalion: Even if it's better than using a mock, it should be refactored to be more reusable.

collector/test/HostHeuristicsTest.cpp
```diff
@@ -92,10 +77,13 @@ TEST(HostHeuristicsTest, TestARM64Heuristic) {
   MockHostInfoHeuristics host;
   KernelVersion version = KernelVersion("6.5.5-200.fc38.aarch64", "", "aarch64");
   CollectorArgs* args = CollectorArgs::getInstance();
-  MockCollectorConfig config(args);
+  int argc = 3;
+  const char* argv[] = {"collector", "--collection-method", "ebpf"};
```
erthalion, line 81: Why is it configured to use `ebpf` everywhere?

erthalion (commented):
Thanks for the PR, few comments below. Also, I've noticed there are couple of
outdated mentions of ebpf in our docs, in "how to start" and "references"
sections -- could you address them as well?

ovalenti (commented):
We probably want to default to core_bpf in the case where the collectionMethod provided by the user is not valid (or no more valid...).

https://github.com/stackrox/collector/pull/1509/files#diff-481ca574af93ac9fb58516ac8df6705b55eff51252c30680b711d4f453ab6f44L132-L135

collector/test/HostHeuristicsTest.cpp
JoukoVirtanen: I am using it again.

collector/test/HostHeuristicsTest.cpp
JoukoVirtanen: I ended up solving this by moving most of the CollectorConfig constructor into an InitCollectorConfig function, including ProcessHostHeuristics, which called Process and caused the fatal error. I feel like the constructor for CollectorConfig was doing too much and splitting it into two is an improvement.

I considered alternatives such as moving ProcessHostHeuristics into either the HostHeuristics or CollectorConfig classes and probably making it a static method. Neither option seemed very good. I also considered creating a new class for ProcessHostHeuristics.

collector/test/HostHeuristicsTest.cpp
JoukoVirtanen: I am now using the mock again.

collector/test/HostHeuristicsTest.cpp
JoukoVirtanen: I am using mocks again.

collector/test/NetworkStatusNotifierTest.cpp
JoukoVirtanen: Using mocks again.

collector/test/NetworkStatusNotifierTest.cpp
JoukoVirtanen: I am not using a mock instead of an environment variable, as it was originally.

JoukoVirtanen:
> We probably want to default to core_bpf in the case where the collectionMethod provided by the user is not valid (or no more valid...).
> 
> https://github.com/stackrox/collector/pull/1509/files#diff-481ca574af93ac9fb58516ac8df6705b55eff51252c30680b711d4f453ab6f44L132-L135

Done

collector/test/HostHeuristicsTest.cpp
JoukoVirtanen: These tests originally specified ebpf as the collection method. I have not changed that. When we remove ebpf these tests will be changed or removed.

JoukoVirtanen: > Even if it's better than using a mock, it should be refactored to be more reusable.

I am using mocks now, so this comment is outdated.

erthalion (approved):

PR #1513, opened 2024-01-23
Set default log level in tests to debug

Stringy:
## Description

During test containerization, the default log level was erroneously set to INFO. This switches it back to debug.


ovalenti:
@Stringy, this is probably part of the goal, but just to make sure : the size of the produced logs for core_bpf is very large. Do we have the right level applied here ?

Stringy:
Yes, I think it's unavoidable currently. The majority of logs that come from core-bpf are Falco's own debug logging in the modern bpf engine. We could perhaps change them to `TRACE` in a follow up (?)

ovalenti:
It is fine, I merely wanted to make sure that the size had been noticed and was expected :+1:

ovalenti (approved):
This will certainly make it easier to get to the root cause of issues when they happen.

PR #1520, opened 2024-01-26
Count the number of throttled log message and print it.

ovalenti:
## Description

We have recently increased significantly the delay for some throttled messages.

In order to get a better idea of the amount of messages, this PR adds a counter.

## Checklist
inspected CI test results for throttled messages:
- [x] Seen a message not throttled.
- [x] Seen a message which was throttled.

## Testing

With a container generating 100 zombie processes, the collector contains:
```
...
[INFO    2024/07/05 19:12:09] Successfully established GRPC stream for signals.
[INFO    2024/07/05 19:12:09] Found self-check process event.
[ERROR   2024/07/05 19:12:09] Could not determine network namespace: No such file or directory
[INFO    2024/07/05 19:12:10] Found self-check connection event.
[INFO    2024/07/05 19:12:19] self-check (pid=76) exited with status: 0
[INFO    2024/07/05 19:12:39] Flushing thread table
[INFO    2024/07/05 19:12:39] Flushing container table
[ERROR   2024/07/05 19:12:39] [Throttled 99 messages] Could not determine network namespace: No such file or directory
[INFO    2024/07/05 19:13:09] Flushing container table
[ERROR   2024/07/05 19:13:09] [Throttled 99 messages] Could not determine network namespace: No such file or directory
...
```

### DIFF
--- a/collector/lib/Logging.h
+++ b/collector/lib/Logging.h
@@ -122,16 +122,16 @@ class LogMessage {
 
 #define CLOG(lvl) CLOG_IF(true, lvl)
 
-#define CLOG_THROTTLED_IF(cond, lvl, interval)                                          \
-  static std::chrono::steady_clock::time_point _clog_lastlog_##__LINE__;                \
-  static unsigned long _clog_lastlog_##__LINE__##_times_ = 0;                           \
-  if (collector::logging::CheckLogLevel(collector::logging::LogLevel::lvl) && (cond)) { \
-    if (std::chrono::steady_clock::now() - _clog_lastlog_##__LINE__ < interval)         \
-      _clog_lastlog_##__LINE__##_times_++;                                              \
-  } else                                                                                \
-    _clog_lastlog_##__LINE__ = std::chrono::steady_clock::now(),                        \
-    _clog_lastlog_##__LINE__##_times_ = 0,                                              \
-    collector::logging::LogMessage(__FILE__, __LINE__, collector::logging::LogLevel::lvl, _clog_lastlog_##__LINE__##_times_)
+#define CLOG_THROTTLED_IF(cond, lvl, interval)                                        \
+  static std::chrono::steady_clock::time_point _clog_lastlog_##__LINE__;              \
+  static unsigned long _clog_lastlog_##__LINE__##_times_ = 0;                         \
+  if (collector::logging::CheckLogLevel(collector::logging::LogLevel::lvl) && (cond)) \
+    if (std::chrono::steady_clock::now() - _clog_lastlog_##__LINE__ < interval)       \
+      _clog_lastlog_##__LINE__##_times_++;                                            \
+    else                                                                              \
+      _clog_lastlog_##__LINE__ = std::chrono::steady_clock::now(),                    \
+      _clog_lastlog_##__LINE__##_times_ = 0,                                          \
+      collector::logging::LogMessage(__FILE__, __LINE__, collector::logging::LogLevel::lvl, _clog_lastlog_##__LINE__##_times_)
 
 #define CLOG_THROTTLED(lvl, interval) CLOG_THROTTLED_IF(true, lvl, interval)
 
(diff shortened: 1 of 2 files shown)


collector/lib/Logging.h
```diff
@@ -118,16 +123,20 @@ class LogMessage {
...
-  static std::chrono::steady_clock::time_point _clog_lastlog_##__LINE__;                \
-  if (collector::logging::CheckLogLevel(collector::logging::LogLevel::lvl) && (cond) && \
-      (std::chrono::steady_clock::now() - _clog_lastlog_##__LINE__ >= interval))        \
-  _clog_lastlog_##__LINE__ = std::chrono::steady_clock::now(),                          \
-  collector::logging::LogMessage(__FILE__, __LINE__, true, collector::logging::LogLevel::lvl)
+#define CLOG_THROTTLED_IF(cond, lvl, interval)                                                                                 \
+  static std::chrono::steady_clock::time_point _clog_lastlog_##__LINE__;                                                       \
+  static unsigned long _clog_throttle_times_##__LINE__ = 0;                                                                    \
+  std::chrono::duration _clog_elapsed_##__LINE__ = std::chrono::steady_clock::now() - _clog_lastlog_##__LINE__;                \
+  if (collector::logging::CheckLogLevel(collector::logging::LogLevel::lvl) && (cond) && _clog_elapsed_##__LINE__ < interval) { \
```
JoukoVirtanen, line 134: I have not been able to refactor this to use nested ifs. I think this is okay.

JoukoVirtanen (approved):

JoukoVirtanen:
I will approve after merge conflicts are fixed.

collector/lib/Logging.h
```diff
@@ -118,16 +123,20 @@ class LogMessage {
...
+    _clog_throttle_times_##__LINE__++;                                                                                         \
+  }                                                                                                                            \
+  if (collector::logging::CheckLogLevel(collector::logging::LogLevel::lvl) && (cond) && _clog_elapsed_##__LINE__ >= interval)  \
```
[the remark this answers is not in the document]
Stringy: Damn, that's very annoying!

JoukoVirtanen (approved):
LGTM!

collector/lib/Logging.h
```diff
@@ -118,16 +123,20 @@ class LogMessage {
...
 
 #define CLOG(lvl) CLOG_IF(true, lvl)
 
-#define CLOG_THROTTLED_IF(cond, lvl, interval)                                          \
-  static std::chrono::steady_clock::time_point _clog_lastlog_##__LINE__;                \
-  if (collector::logging::CheckLogLevel(collector::logging::LogLevel::lvl) && (cond) && \
-      (std::chrono::steady_clock::now() - _clog_lastlog_##__LINE__ >= interval))        \
-  _clog_lastlog_##__LINE__ = std::chrono::steady_clock::now(),                          \
-  collector::logging::LogMessage(__FILE__, __LINE__, true, collector::logging::LogLevel::lvl)
+#define CLOG_THROTTLED_IF(cond, lvl, interval)                                                                                 \
```
Molter73, line 130: I'm gonna be honest, I don't like macros. I think most of them are directly written by the devil himself, waiting for us to make a silly mistake in an expansion, triggering undefined behavior and bringing down entire programs in an instant.

This particular case was already pretty bad, but with the added changes it becomes borderline unreadable, I had to put a TON of effort into just trying to understand what was going on in it and, I have to assume, if we ever need to do any further changes or (God forbid) do a bug fix, it will take way too much effort to do so.

That said, I am also in favor of the added feature, I think counting the amount of times a log has been throttled adds value (throttling twice is not as bad as throttling 2000 times). Maybe we can come up with a better solution, there might be a way we can rewrite at least part of the macro to be a concrete method and pass things like `__LINE__` as an argument to it. Or maybe we have to rethink how our entire logging implementation works, the implementation itself is really clever, using the destructor of the class to do the actual print out, but that also means every time we log we are creating and destroying an object, maybe we want to change it to have a single instance of a logger object and this sort of throttling can be done with a map of sorts, grouping line number to the amount of triggers that have occurred.

I digress, I will not block the change from being merged because I like the idea behind it, but I cannot in good faith approve the PR because of the reasons stated before.

PR #1525, opened 2024-01-30
Add switch event

erthalion:
## Description

After temporary disabling trusted exepath [[1]] it turns out that original
implementation of exepath resolution wasn't reliable enough, producing flaky
test results. The usual failure looks like this:

* A process (usually runc) performs a chain of clones, potentially intertwined
  with other activity we watch, e.g. closing pipes (runc is talking to clones
  using pipes) or doing networking (as in some of our integration tests).

* As a last step this process does execve to perform some workload.

* The scap events we receive could be either our of order or even missing
  sometimes. Exepath resolution relies on the presence of an enter event, thus
  we got non-deterministic failures.

The reason for that is overlooked missing `sched_switch` event, which similar
to `procexit` became non-default one in [[2]]. This event makes conveying
process information more reliable when it comes to an overloaded system and
frequent switches between processes.

## Checklist

## Testing Performed

Local testing, Stackrox CI testing.

[1]: https://github.com/stackrox/collector/pull/1508
[2]: https://github.com/falcosecurity/libs/pull/1001

### DIFF
--- a/collector/lib/KernelDriver.h
+++ b/collector/lib/KernelDriver.h
@@ -51,6 +51,8 @@ class IKernelDriver {
      * procexit is essential for keeping threadinfo cache under control.
      */
     ppm_sc.insert((ppm_sc_code)PPM_SC_SCHED_PROCESS_EXIT);
+
+    ppm_sc.insert((ppm_sc_code)PPM_SC_SCHED_SWITCH);
     return ppm_sc;
   }
 };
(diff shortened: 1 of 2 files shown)


ovalenti (approved):
Glad that you found a way to solve this, thanks !

I was first concerned by the performance impact of adding a signal, but as you
explained, this used to be a default in the previous version; so it should not be
worse.

PR #1531, opened 2024-02-06
ROX-21403: Add fe80::/10 to the networks ignored by default.

ovalenti:
## Description

Link-local addresses are defined for IPv6 and was overlooked in #1511 

## Checklist


vikin91 (approved):

erthalion (approved):
Looks good, thanks. One small request, could you add somewhere a description where those defaults are coming from, just to give them some meaning.

PR #1537, opened 2024-02-07
Remove drivers builds from pipeline

Stringy:
## Description

This PR removes all drivers builds and workflows from the master GHA pipeline, including the construction of full images. Instead, the slim image is retagged. This will be undone later on when the notion of a full and slim image is completely deprecated from the product.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


Molter73 (commented):
Godspeed drivers! You will not be missed!

Side note, do we want to remove the `ansible/ci-build-collector-full.yml` playbook? I assume it won't be used by anyone if the `collector-full.yml` workflow is removed.

.github/workflows/collector.yml
```diff
@@ -171,6 +171,12 @@ jobs:
           username: ${{ secrets.QUAY_STACKROX_IO_RW_USERNAME }}
           password: ${{ secrets.QUAY_STACKROX_IO_RW_PASSWORD }}
 
+      - name: Create and push multiarch manifest for stackrox-io
+        uses: ./.github/actions/create-multiarch-manifest
+        with:
+          base-image: quay.io/stackrox-io/collector:${{ inputs.collector-tag }}
+          archs: ${{ env.ARCHS }}
+
       - name: Create and push multiarch manifest for stackrox-io -slim
```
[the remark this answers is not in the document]
Stringy: Hopefully I can start work on removing slim and base images as soon as the drivers have been shifted. I have been working on a plan for this as it affects the whole product.

.github/workflows/cpaas.yml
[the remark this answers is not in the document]
Stringy: I could certainly retain the full image build in the drivers branch to add to the testing. Would that be an appropriate middle ground do you think?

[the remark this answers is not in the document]
Stringy: actually having said that it probably makes more sense to keep it in master, since it's only used in cpaas steps and for constructing a test image. Adding that to the drivers branch complicates things a fair bit

.github/workflows/cpaas.yml
```diff
@@ -82,18 +82,13 @@ jobs:
...
-    - init
-    - build-collector-slim
-    - sync-drivers-and-support-packages
+      - name: Retag and push stackrox-io -amd64-slim
+        uses: stackrox/actions/images/retag-and-push@v1
+        with:
+          src-image: brew.registry.redhat.io/rh-osbs/rhacs-collector-slim-rhel8:0.1.0
+          dst-image: quay.io/rhacs-eng/collector:${{ needs.init.outputs.collector-tag }}-cpaas-${{ matrix.arch }}-slim
+          username: ${{ secrets.QUAY_RHACS_ENG_RW_USERNAME }}
+          password: ${{ secrets.QUAY_RHACS_ENG_RW_PASSWORD }}
```
Molter73, line 91: I noticed you had to add the `collector-test-tag` to some workflows because there is no multiarch manifest for this image, wonder if it'd be easier to retag and push without the `-slim` part, then simply use the `create-multiarch-manifest` action instead.

Molter73: Also, what's the difference between this step and the one before? isn't `collector-image` the same as `quay.io/rhacs-eng/collector:${{ collector-tag }}`?

.github/workflows/collector.yml
```diff
@@ -171,7 +171,13 @@ jobs:
           username: ${{ secrets.QUAY_STACKROX_IO_RW_USERNAME }}
           password: ${{ secrets.QUAY_STACKROX_IO_RW_PASSWORD }}
 
-      - name: Create and push multiarch manifest for stackrox-io -slim
+      - name: Create and push multiarch manifest for stackrox-io
+        uses: ./.github/actions/create-multiarch-manifest
+        with:
+          base-image: quay.io/stackrox-io/collector:${{ inputs.collector-tag }}
```
Molter73, line 177: Do we also need to add the `-latest` variants for both stackrox-io and rhacs-eng here?

ansible/ci-build-collector.yml
```diff
@@ -52,46 +52,60 @@
...
     - name: Print images pushed
       debug:
         msg:
         - "Pushed the following images:"
-        - "  {{ collector_image }}-{{ arch }}-slim"
+        - "  {{ collector_image }}-{{ arch }}"
         - "  {{ collector_image }}-{{ arch }}-base"
-        - "  {{ rhacs_eng_image }}-{{ arch }}-slim"
+        - "  {{ rhacs_eng_image }}-{{ arch }}"
         - "  {{ rhacs_eng_image }}-{{ arch }}-base"
```
Molter73, line 109: Instead of removing the -slim tags, we might want to add the -latest and this one to the list of images we push, that way CI can just create all the multiarch manifest in a single job.

```yml
        - "Pushed the following images:"
        - "  {{ collector_image }}-{{ arch }}"
        - "  {{ collector_image }}-{{ arch }}-latest"
        - "  {{ collector_image }}-{{ arch }}-slim"
        - "  {{ collector_image }}-{{ arch }}-base"
        - "  {{ rhacs_eng_image }}-{{ arch }}"
        - "  {{ rhacs_eng_image }}-{{ arch }}-latest"
        - "  {{ rhacs_eng_image }}-{{ arch }}-slim"
        - "  {{ rhacs_eng_image }}-{{ arch }}-base"
```

.github/workflows/collector.yml
```diff
@@ -209,44 +221,61 @@ jobs:
...
 
       - name: Retag and push rhacs-eng -base
         uses: stackrox/actions/images/retag-and-push@v1
         with:
-          src-image: ${{ inputs.collector-image }}-amd64-slim
+          src-image: ${{ inputs.collector-image }}-amd64
           dst-image: ${{ env.RHACS_ENG_IMAGE }}-base
           username: ${{ secrets.QUAY_RHACS_ENG_RW_USERNAME }}
           password: ${{ secrets.QUAY_RHACS_ENG_RW_PASSWORD }}
+
```
JoukoVirtanen, line 281: [Ultra nit] Remove blank line.

.github/workflows/cpaas.yml
```diff
@@ -105,7 +100,7 @@ jobs:
       rebuild-qa-containers: ${{ needs.init.outputs.rebuild-qa-containers == 'true' }}
     secrets: inherit
 
-  integration-tests:
+  amd64-integration-tests:
```
JoukoVirtanen, line 103: There is some duplication here with `amd64-integration-tests`, `s390x-integration-tests`, and `ppc64le-integration-tests`. Perhaps this can be reduced.

ansible/integration-tests.yml
```diff
@@ -11,7 +11,7 @@
       vars:
         vm_list: "{{ virtual_machines }}"
         collection_method: "{{ method }}"
-      loop: "{{ lookup('env', 'COLLECTION_METHODS', default='core_bpf,ebpf') | split(',') }}"
+      loop: "{{ lookup('env', 'COLLECTION_METHODS', default='core_bpf') | split(',') }}"
```
JoukoVirtanen, line 14: You don't need a loop anymore if you only have one collection method.

ansible/roles/run-test-target/tasks/main.yml
```diff
@@ -35,7 +35,7 @@
   vars:
     collection_method: "{{ item }}"
   with_items:
-    "{{ lookup('env', 'COLLECTION_METHODS', default='core_bpf,ebpf') | split(',') }}"
+    "{{ lookup('env', 'COLLECTION_METHODS', default='core_bpf') | split(',') }}"
```
JoukoVirtanen, line 38: Same here

.github/workflows/cpaas.yml
Stringy: I'm not sure I understand what you mean about using the create-multiarch-manifest action. We need to push -slim versions of the images before we're able to then create a -slim manifest.

As for the duplication with the previous step... yes that is odd. Not sure what I was thinking here, i'll double check.

.github/workflows/cpaas.yml
Stringy: This is sort of intentionally duplicated to make it clearer which architecture tests are running.

Stringy:
I have tested the manifest creation changes with `run-multiarch-builds` label added and removed, and in both scenarios the manifests are created correctly. I'll leave the multiarch builds on now.

Stringy:
Closing now in favour of #1667

PR #1546, opened 2024-02-11
Reduce integration test retries

robbycochran:
### Problem
Integration tests were taking an excessive amount of time due to retries on operations involving removed containers, as well as unnecessarily long health checks and redundant cleanups.

### Solution
* Implemented early exits in retry functions for remove, stop, and kill container operations if the container is no longer present based on the output message.
* Reduced collector health check wait time from 30s to 5s. This improves the common case where collector is healthier before 30 sec has elapsed.
* Eliminated redundant calls to collector cleanup functions.
* Corrected an error in the container name for jsonlabel, ending repeated cleanup retries.

### Impact
Integration test suite execution time (including VM provisioning) decreased significantly, from approximately 77 minutes to 29 minutes, representing a ~2.75x speedup.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)

### DIFF
--- a/integration-tests/suites/image_json.go
+++ b/integration-tests/suites/image_json.go
@@ -28,6 +28,5 @@ func (s *ImageLabelJSONTestSuite) TestRunImageWithJSONLabel() {
 }
 
 func (s *ImageLabelJSONTestSuite) TearDownSuite() {
-	s.StopCollector()
-	s.cleanupContainers("json-label")
+	s.cleanupContainers("jsonlabel")
 }
--- a/integration-tests/suites/base.go
+++ b/integration-tests/suites/base.go
@@ -30,7 +30,7 @@ const (
 
 	containerStatsName = "container-stats"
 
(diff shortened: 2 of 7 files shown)


Stringy (approved):
That's a pretty incredible speed up, and for relatively few changes. LGTM! Thanks Robby!

integration-tests/suites/image_json.go
```diff
@@ -28,6 +28,5 @@ func (s *ImageLabelJSONTestSuite) TestRunImageWithJSONLabel() {
 }
 
 func (s *ImageLabelJSONTestSuite) TearDownSuite() {
-	s.StopCollector()
```
Stringy, line 31: I think this is still required, because we're starting collector on line 13 (or we can remove line 13 if collector is not needed for this test)

PR #1553, opened 2024-02-13
Make thread cache size configurable

erthalion:
## Description

This value is important to manage memory consumption under process based loads, so expose it for configuration. The current value is in fact cut to 131072, and to address memory consumption concerns I've set the current size to the half of this value. 

## Checklist

## Testing Performed

Local testing.

### DIFF
--- a/collector/lib/CollectorConfig.h
+++ b/collector/lib/CollectorConfig.h
@@ -117,8 +117,8 @@ class CollectorConfig {
   // Max size of the thread cache. This parameter essentially translated into
   // the upper boundary for memory consumption. Note that Falco puts it's own
   // upper limit on top of this value, m_thread_table_absolute_max_size, which
-  // is 2^17 (131072).
-  unsigned int sinsp_thread_cache_size_ = 524288;
+  // is 2^17 (131072) and twice as large.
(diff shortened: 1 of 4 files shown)


collector/lib/CollectorConfig.h
```diff
@@ -113,6 +114,12 @@ class CollectorConfig {
   // Size of one ring buffer, in bytes
   unsigned int sinsp_buffer_size_;
 
+  // Max size of the thread cache. This parameter essentially translated into
+  // the upper boundary for memory consumption. Note that Falco puts it's own
+  // upper limit on top of this value, m_thread_table_absolute_max_size, which
+  // is 2^17 (131072).
+  unsigned int sinsp_thread_cache_size_ = 524288;
```
robbycochran, line 121: Should we just go ahead and make this the max falco value since the current setting will be altered by falco?

robbycochran (approved):

collector/lib/CollectorConfig.h
erthalion: Yeah, let me do this. As discussed, I'm also going to make the thread cache twice as smaller.

PR #1558, opened 2024-02-14
New metric to observe the thread cache size

### LINKED ISSUE #1456: Metrics thread table size
While troubleshooting vanilla Falco and previous memory related issues it
proved to be useful to have an understanding how the thread cache is growing.
To make it more visible, expose the current thread table size as a new
prometheus metric, e.g. `rox_collector_events{type="threadCacheSize"}`.

The numbers we're interested in could be obtained via libsinsp inspector
function `get_thread_count()`. Since this metric is not directly dependent on
event stream, we need to decide when exactly to take the counters current value
-- to not do unnecessary work if it's changing slowly, but to be fine-grained
enough to notice relevant spikes.

When troubleshooting vanilla Falco, the hacky solution I used was to log thread
count with throttled logging, and it was providing enough information. Thus, the
proposal is to update thread counter metric based on the number of processes
received, but with some throttling, e.g. when we receive every n'th process we
capture current thread table size.

Part of #1320

ovalenti:
## Description

This implements a simple gauge to report the current size of the thread cache. It is sampled
by the CollectorStatsExporter every 5s.

fixes #1456

## Checklist
- [x] Deployed and verified once that the exported value looks coherent

### DIFF
--- a/collector/lib/CollectorStatsExporter.cpp
+++ b/collector/lib/CollectorStatsExporter.cpp
@@ -115,6 +115,7 @@ void CollectorStatsExporter::run() {
   auto& preemptions = collectorEventCounters.Add({{"type", "preemptions"}});
   auto& userspaceEvents = collectorEventCounters.Add({{"type", "userspace"}});
   auto& grpcSendFailures = collectorEventCounters.Add({{"type", "grpcSendFailures"}});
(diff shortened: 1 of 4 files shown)


erthalion (approved):
Looks good, thanks.

PR #1565, opened 2024-02-19
Add dependabot checks for integration-tests

stefan-cornelius:
## Description

I've promised to create a PR to enable dependabot in collector. Unfortunately, while doing so, I noticed that a PR alone is not sufficient. It requires additional changes to repository settings related to dependabot.

This PR will add settings related to:
https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates
AFAICT, this is not limited to security updates only, so this may actually increase the noise (compared to enabling dependabot alerts). Also, the relevant option has to be enabled in the repository settings or this nothing will happen.

If you want only security updates and automatic PRs, the likely best way to achieve this is to enable the dependabot alert functionality. I've done so in my fork and it did the right thing, without any additional configuration. I don't think I can do this via PR (anymore?).

Given my permissions and githubs settings around dependabot, this is likely the most I can do for now. I'm sorry that I couldn't be of further assistance, but I would still like to see dependabot enabled. Please let me know how you want to proceed or what you need. Thanks!

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

No testing. Doesn't change any code, just sets some options for the dependabot version check feature. It quite possible that this feature is not enabled for this repository. Until it is, this change likely has no impact at all.


Stringy:
Hi @stefan-cornelius thanks for the PR! I think the changes look good.

I'll discuss with the team today what we want to do with regard to dependabot alerting, automated PRs, etc. For the tests in particular I think we're less concerned about the impact of updating go mod versions (and are probably happy with automated merging) -- it's usually the Dockerfiles that can cause some build trouble due to our sometimes quite sensitive package dependencies.

PR #1571, opened 2024-02-22
Adds a sentinel step to drivers build for release branches

Stringy:
## Description

Github Actions will skip a job if all of its dependencies are also skipped. This is ordinarily fine, but for release branches we don't want the drivers build to run, but we do want the full builds to run.

This adds a sentinel job in the drivers workflow to ensure that even if the driver build itself is skipped, the workflow runs at least one job, and therefore the full builds will also run.

This will not run on this PR, but I have tested the mechanism on one of my own repositories and it works: https://github.com/Stringy/wisdom/pull/10

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed
Tested on my own repo here https://github.com/Stringy/wisdom/pull/10

### DIFF
--- a/.github/workflows/drivers.yml
+++ b/.github/workflows/drivers.yml
@@ -25,6 +25,17 @@ env:
   DRIVERS_BUCKET: ${{ inputs.drivers-bucket }}
 
 jobs:
+  # This sentinel job exists such that this workflow call "runs" even
+  # if subsequent driver build steps are skipped. This means that other
+  # workflows can depend on this workflow and still run even if it is skipped
+  #
+  # This only affects release branches, so is limited to those.
+  sentinel:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref_name, 'release-')
+    steps:
+      - run: echo Drivers Build
+
   split-tasks:
     runs-on: ubuntu-latest
     if: |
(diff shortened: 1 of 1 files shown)


.github/workflows/drivers.yml
```diff
@@ -25,6 +25,17 @@ env:
...
 
 jobs:
+  # This sentinel job exists such that this workflow call "runs" even
+  # if subsequent driver build steps are skipped. This means that other
+  # workflows can depend on this workflow and still run even if it is skipped
+  #
+  # This only affects release branches, so is limited to those.
+  sentinel:
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref_name, 'release-')
```
Molter73, line 35: This will only trigger on the `release-*` branch, but that doesn't tag the versioned collector which is the one we need for releases. You might want to replace this check or complement it with `github.ref_type == 'tag'`
```suggestion
    if: startsWith(github.ref_name, 'release-') || github.ref_type == 'tag'
```

Molter73:
Side note, I thought (and the PR comment seems to confirm) that GHA skipped a step if ALL previous needed steps were skipped. Building the full image needs both `init` and `build-collector-slim`, which were not skipped, so I find the behavior a bit odd.

Stringy:
Seems that it was missed on the original PR, because the full builds didn't run there either: https://github.com/stackrox/collector/actions/runs/6704112012

Molter73 (approved):
LGTM!

PR #1573, opened 2024-02-22
Konflux(ROX-22195): Install RHEL RPMs with subscription manager workaround

tommartensen:
## Description

This PR changes the dependency installations in the Konflux images from using CentOS dependencies to using a UBI image with workaround entitlements to install RHEL RPMs. 

Also harmonized the Dockerfiles between `-slim` and `-full` based on midstream. 

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Passing Konflux pipeline is sufficient.


collector/container/konflux-slim.Dockerfile
```diff
@@ -1,55 +1,51 @@
...
+        git \
+        elfutils-libelf-devel \
+        tbb-devel \
+        jq-devel \
+        c-ares-devel && \
+    /tmp/.rhtap/scripts/subscription-manager-bro.sh cleanup && \
+    # We can do usual cleanup while we're here: remove packages that would trigger violations. \
+    dnf -y --installroot=/mnt clean all && \
+    rpm --root=/mnt --verbose -e --nodeps $(rpm --root=/mnt -qa 'curl' '*rpm*' '*dnf*' '*libsolv*' '*hawkey*' 'yum*') && \
+    rm -rf /mnt/var/cache/dnf /mnt/var/cache/yum
```
msugakov, line 42: Removing package managers isn't necessary from the builder stage, only the final stage. This removal is done to satisfy ACS own policy about having package manager in the image which was requested for some compliance policies.

```suggestion
    dnf -y --installroot=/mnt clean all
```

collector/container/konflux-slim.Dockerfile
```diff
@@ -96,15 +92,37 @@ RUN ./builder/install/install-dependencies.sh && \
...
+
+FROM scratch
 
 # TODO(ROX-20236): configure injection of dynamic version value when it becomes possible.
 ARG COLLECTOR_VERSION=0.0.1-todo
 
-ENV COLLECTOR_HOST_ROOT=/host
+COPY --from=rpm-implanter-app /mnt /
```
msugakov, line 123: Just a cosmetic reordering suggestion

```suggestion
FROM scratch

COPY --from=rpm-implanter-app /mnt /

# TODO(ROX-20236): configure injection of dynamic version value when it becomes possible.
ARG COLLECTOR_VERSION=0.0.1-todo
```

collector/container/konflux-slim.Dockerfile
```diff
@@ -96,15 +92,37 @@ RUN ./builder/install/install-dependencies.sh && \
     ctest -V --test-dir ${CMAKE_BUILD_DIR} && \
     strip -v --strip-unneeded "${CMAKE_BUILD_DIR}/collector/collector"
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+# Application
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS ubi-minimal
+# The installer must be ubi (not minimal) and must be 8.9 or later since the earlier versions complain:
+#  subscription-manager is disabled when running inside a container. Please refer to your host system for subscription management.
+FROM ubi-normal AS rpm-implanter-app
```
msugakov, line 99: You could use the same `rpm-implanter` for both builder and app, but I would not insist, and besides we have a blocker problem to solve now with EC checks.

.tekton/collector-slim-pull-request.yaml
```diff
@@ -213,12 +217,27 @@ spec:
       - name: basic-auth
         workspace: git-auth
 
+    - name: download-external-content
```
msugakov, line 220: How about calling this `fetch-external-content`? This way, the name is still ok-ish for `collector-slim` which does not need to _download_ things from network and only injects the activation key (which is in a way external).

msugakov (commented):
Few quick comments. Unfortunately, I can't spot the thing which would cause SBOM to be missing. I'm sure the setup is the same for `main` where this problem does not occur. Need to investigate further.

collector/container/konflux-slim.Dockerfile
tommartensen: https://github.com/stackrox/collector/pull/1573/commits/57288765f6b72593489447a18a1f6694ca9e94a8#diff-ae246d60ffd0ce9d7455810ae6cdb168e932b1285a90712fbae086b826b3f501R39

collector/container/konflux-slim.Dockerfile
tommartensen: https://github.com/stackrox/collector/pull/1573/commits/49c9860e6d637e3d750bebcd66057f922ece92e0

.tekton/collector-slim-pull-request.yaml
tommartensen: That's a copy&paste artifact. I renamed the step to `prepare-rhel-rpm-subscriptions` in https://github.com/stackrox/collector/pull/1573/commits/23d4abc2834f524c7d4eed81dd9e5e67d7be4d70

collector/container/konflux-slim.Dockerfile
tommartensen: If I understand your suggestion correctly and installed all dependencies in a single `rpm-implanter` stage, we would have all build time dependencies (like cmake, git, gcc, ...) in the final image. 
That would allow someone who has access to the container to compile new content, and cause more pain with CVEs. Additionally, the final image would be bigger.
Therefore, I would keep them as separate stages.

tommartensen: That said, I merged the `rpm-installer-builder` and `builder` stages into one (https://github.com/stackrox/collector/pull/1573/commits/82f38460032ababceb355fe64d60115dae206574).

collector/container/konflux-slim.Dockerfile
```diff
@@ -1,55 +1,46 @@
...
-        patchutils \
-        passwd \
-        pkgconfig \
-        rsync \
-        tar \
-        unzip \
-        valgrind \
         wget \
-        which \
+        unzip \
```
tommartensen, line 19: Updated the dependency list to https://gitlab.cee.redhat.com/stackrox/rhacs-midstream/-/blob/rhacs-1.0-rhel-8/distgit/containers/rhacs-collector-slim/Dockerfile.in?ref_type=heads. Previously it was the list of upstream dependencies because of the CentOS base image.

collector/container/konflux.Dockerfile
```diff
@@ -160,9 +192,6 @@ LABEL \
 
 WORKDIR /
 
-COPY collector/container/rhel/install.sh /
```
tommartensen, line 163: I replaced this with https://github.com/stackrox/collector/pull/1573/files#diff-dd2e1150313169ab18110018f7c1e085ab8c31e164916b63883f5962afb49e79R155-R165

tommartensen:
This is the diff between slim and full Dockerfiles:

```
88,125d87
< # 0.1.0 is a floating tag and it's used intentionally to pick up the most recent downstream drivers build without
< # having to routinely and frequently bump tags here.
< FROM brew.registry.redhat.io/rh-osbs/rhacs-drivers-build-rhel8:0.1.0 AS drivers-build
< 
< # TODO(ROX-20312): we can't pin image tag or digest because currently there's no mechanism to auto-update that.
< FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS unpacker
< 
< RUN microdnf install -y unzip findutils
< WORKDIR /staging
< 
< COPY staging/support-pkg.zip /staging/
< COPY kernel-modules/MODULE_VERSION MODULE_VERSION.txt
< RUN mkdir -p "/staging/kernel-modules/$(cat MODULE_VERSION.txt)"
< 
< # First, unpack upstream support package, only on x86_64
< RUN if [[ "$(uname -m)" == x86_64 ]]; then unzip support-pkg.zip ; fi
< # Fail build if there were no drivers in the support package matching the module version.
< RUN if [[ "$(uname -m)" == x86_64 && "$(ls -A /staging/kernel-modules/$(cat MODULE_VERSION.txt))" == "" ]] ; then \
<       >&2 echo "Did not find any kernel drivers for the module version $(cat MODULE_VERSION.txt) in the support package"; \
<       exit 1; \
<     fi
< 
< # Next, import modules from downstream build, which take priority over upstream, on non-x86 architectures
< COPY --from=drivers-build /kernel-modules /staging/downstream
< RUN if [[ "$(uname -m)" != x86_64 ]]; then \
<       cp -r /staging/downstream/. /staging/kernel-modules/ ; \
<     fi
< 
< # Create destination for drivers.
< RUN mkdir /kernel-modules
< # Move files for the current version to /kernel-modules
< RUN find "/staging/kernel-modules/$(cat MODULE_VERSION.txt)/" -type f -exec mv -t /kernel-modules {} +
< # Fail the build if at the end there were no drivers matching the module version.
< RUN if [[ "$(ls -A /kernel-modules)" == "" ]]; then \
<         >&2 echo "Did not find any kernel drivers for the module version $(cat MODULE_VERSION.txt)."; \
<         exit 1; \
<     fi
< 
159c121
<     com.redhat.component="rhacs-collector-container" \
---
>     com.redhat.component="rhacs-collector-slim-container" \
164c126
<     io.k8s.display-name="collector" \
---
>     io.k8s.display-name="collector-slim" \
167c129
<     name="rhacs-collector-rhel8" \
---
>     name="rhacs-collector-slim-rhel8" \
181d142
< COPY --from=unpacker /kernel-modules /kernel-modules
```

.konflux/scripts/subscription-manager-bro.sh
```diff
@@ -284,8 +289,8 @@ function assert_diff {
         echo '/var/cache(: |/)dnf'
     } >> "$var_exclusions"
 
-    if { diff --brief --recursive --no-dereference "$expected/var" "$actual/var" || true; } | \
-        grep -vEf "$var_exclusions" | { grep '.'; }; then
+    if { diff --brief --recursive --no-dereference "$expected/var" "$actual/var" || true; } \
+                                                                                            | grep -vEf "$var_exclusions" | { grep '.'; }; then
```
msugakov, line 293: [nit, ignorable] This indent looks weird but I don't mind.
image

.tekton/collector-pull-request.yaml
```diff
@@ -221,12 +225,27 @@ spec:
...
+      - name: source
+        workspace: workspace
+      - name: subscription-manager-activation-key
+        workspace: subscription-manager-activation-key
+      taskSpec:
+        steps:
+        # TODO(ROX-20651): use content sets instead of subscription manager for access to RHEL RPMs once available.
+        - name: smuggle-activation-key
+          image: registry.access.redhat.com/ubi8/ubi:latest
+          script: exec /workspace/source/source/.konflux/scripts/subscription-manager-bro.sh smuggle
```
msugakov, line 241: Don't know if this really matters, but you could also resolve workspace path dynamically

```suggestion
          script: exec "$(workspaces.source.path)/source/.konflux/scripts/subscription-manager-bro.sh" smuggle
```

Up to you.

.tekton/collector-pull-request.yaml
```diff
@@ -221,12 +225,27 @@ spec:
...
+
     - name: prefetch-dependencies
       params:
       - name: input
         value: $(params.prefetch-input)
       runAfter:
-      - clone-repository
+      - prepare-rhel-rpm-subscriptions
```
msugakov, line 248: Now, after seeing how collector-full pipeline graph is arranged, I'd nitpick on this.

You have:

```yaml
    - name: prepare-rhel-rpm-subscriptions
      runAfter:
      - clone-repository
```
I agree with this ^ one.

```yaml
    - name: prefetch-dependencies
      runAfter:
      - prepare-rhel-rpm-subscriptions
```
Prefetching dependencies does not need to depend on `prepare-rhel-rpm-subscriptions`. It can remain depending on `clone-repository`.

```yaml
    - name: fetch-support-package
      runAfter:
      - init
      - clone-repository
```
`clone-repository` already depends on `init`, therefore `fetch-support-package` does not need `init` in its list.

```yaml
    - name: build-container
      runAfter:
      - prefetch-dependencies
      - fetch-support-package
```
This would need `prepare-rhel-rpm-subscriptions` additionally in its list.

Basically, either all three `prepare-rhel-rpm-subscriptions`, `prefetch-dependencies` and `fetch-support-package` need to run in parallel with each other or sequentially for consistency. Currently, it's a mix of both and it's unclear whether that's on purpose.

collector/container/konflux-slim.Dockerfile
msugakov: I meant this
https://github.com/stackrox/stackrox/blob/25beb9ce8cf6da5bdd88fd5e20bc960c2ce7426e/scripts/konflux/subscription-manager/bro.self-test-multiple-targets.Dockerfile#L13-L34

collector/container/konflux-slim.Dockerfile
```diff
@@ -74,9 +65,7 @@ RUN mkdir kernel-modules \
     && cp -a /CMakeLists.txt CMakeLists.txt
```
msugakov, line 65: GitHub does not let me comment on lines above, and this was in your earlier PR, but I'd still wonder without blocking this PR.

The downstream Dockerfile has
```
ln -s builder/third_party third_party
```

whereas you have
```
&& cp -a /builder/third_party third_party \
```

I suppose both should make the same files visible. Was there a reason you could not use `ln` in Konflux?

collector/container/konflux-slim.Dockerfile
```diff
@@ -96,44 +85,70 @@ RUN ./builder/install/install-dependencies.sh && \
     ctest -V --test-dir ${CMAKE_BUILD_DIR} && \
     strip -v --strip-unneeded "${CMAKE_BUILD_DIR}/collector/collector"
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+# Application
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS ubi-minimal
+# The installer must be ubi (not minimal) and must be 8.9 or later since the earlier versions complain:
+#  subscription-manager is disabled when running inside a container. Please refer to your host system for subscription management.
```
msugakov, line 91: I find this info a bit redundant here but I'm also ok if you keep it.
```suggestion
```

collector/container/konflux-slim.Dockerfile
```diff
@@ -1,55 +1,46 @@
...
+        jq-devel \
+        c-ares-devel && \
+    /tmp/.konflux/scripts/subscription-manager-bro.sh cleanup && \
+    dnf -y --installroot=/mnt clean all
+
+FROM scratch as builder
+
+COPY --from=rpm-implanter-builder /mnt /
+
+COPY . .
```
msugakov, line 43: I realized this dumps everything to the root dir which may be problematic if there's a name collision.

How about if you introduce a staging location?

```suggestion
ARG SOURCES_DIR=/staging

COPY . $SOURCES_DIR
```

Then, later on, things like
```
&& cp -a /builder builder \
```
would become
```
&& cp -a $SOURCES_DIR/builder builder \
```

.tekton/collector-pull-request.yaml
tommartensen: https://github.com/stackrox/collector/pull/1573/commits/c41368a927d845cdf71235ecaef27abda93bef80

collector/container/konflux-slim.Dockerfile
tommartensen: Thanks for clarifying, I hadn't thought about that. I will keep it as is though.

collector/container/konflux-slim.Dockerfile
tommartensen: https://github.com/stackrox/collector/pull/1573/commits/87f70d22ca750cad1a7dd3f83ef92ae68cab833c

collector/container/konflux-slim.Dockerfile
tommartensen: https://github.com/stackrox/collector/pull/1573/commits/2eadedead357d172fda01842b7c766b6b9839e11

collector/container/konflux-slim.Dockerfile
tommartensen: > Was there a reason you could not use ln in Konflux?

No. 

https://github.com/stackrox/collector/pull/1573/commits/2eadedead357d172fda01842b7c766b6b9839e11

.tekton/collector-pull-request.yaml
tommartensen: Screenshot 2024-02-28 at 14 50 49

This looks better.

collector/container/konflux.Dockerfile
```diff
@@ -40,7 +40,9 @@ FROM scratch as builder
 
 COPY --from=rpm-implanter-builder /mnt /
 
-COPY . .
+ARG SOURCES_DIR=/staging
```
msugakov, line 43: Optional, cosmetics, and up to you.

```suggestion
ARG SOURCES_DIR=/src
```

collector/container/konflux-slim.Dockerfile
```diff
@@ -40,7 +40,9 @@ FROM scratch as builder
 
 COPY --from=rpm-implanter-builder /mnt /
 
-COPY . .
+ARG SOURCES_DIR=/staging
```
msugakov, line 43: Optional, cosmetics, and up to you.

```suggestion
ARG SOURCES_DIR=/src
```

msugakov (approved):
LGTM

I realized the naming suggestion I made was not the best one. Feel free to ship it with our without addressing the last two comments.

collector/container/konflux.Dockerfile
tommartensen: I think the chances of a directory in ubi existing called `/src` are slightly higher than `/staging`, so let's keep it like that.

PR #1575, opened 2024-02-26
Add retries

erthalion:
## Description

We frequently observe issues with spinning up new test VMs, that manifests itself as a non-reachable host. Try to reduce the problem via increasing number of retries (the default value seems to be 0 [[1]]).

[1]: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/ssh_connection.html#parameter-reconnection_retries

## Checklist

## Testing Performed

CI is sufficient.

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -1,11 +1,11 @@
 [defaults]
-stdout_callback=debug
-stderr_callback=debug
-display_skipped_hosts=false
-host_key_checking=False
+stdout_callback = debug
+stderr_callback = debug
+display_skipped_hosts = false
+host_key_checking = False
 remote_tmp = /tmp/ansible
 forks = 20
 
 [ssh_connection]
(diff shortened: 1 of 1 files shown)


ansible/ansible.cfg
```diff
@@ -8,3 +8,4 @@ forks = 20
 
 [ssh_connection]
 ssh_args = -o StrictHostKeyChecking=no -C -o ControlMaster=auto -o ControlPersist=60s -o ServerAliveInterval=30 -o ServerAliveCountMax=10
+retries=5
```
Molter73, line 11: Was gonna mention adding spaces around the `=`, then I noticed the file mixes these a bunch, so meh (we do need to get better at keeping coherent styles though).

Molter73 (approved):
Can't really think of any downsides to this, so, LGTM!

ansible/ansible.cfg
erthalion: Why not, lets unify the formatting.

PR #1576, opened 2024-02-26
Configure detailed metrics

erthalion:
## Description

Currently, Collector exposes two types of prometheus metrics: regular aggregates and per-syscall aggregates. The latter one is quite verbose and produces more than half of the whole output (181 out of 284 lines), thus it makes sense to make it configurable to be able to disable in the future.

From one side it would enable us to finally expose Collector prometheus metrics, at the same we still would be able to use per-syscall metrics for troubleshooting in development environment.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Local testing.


Molter73 (commented):
Wonder if it would make sense to have this as a run-time check instead of compile-time. Thinking from a user of collector, the compile time check would mean a new image needs to be pulled and deployed (maybe even built, since I don't expect us to maintain multiple versions of the image), whereas a run time check could be toggled via an env variable, have the container restart and presto! The metrics are now collected. We could still have it so these metrics are collected by default. WDYT?

erthalion:
> Wonder if it would make sense to have this as a run-time check instead of compile-time. Thinking from a user of collector, the compile time check would mean a new image needs to be pulled and deployed (maybe even built, since I don't expect us to maintain multiple versions of the image), whereas a run time check could be toggled via an env variable, have the container restart and presto! The metrics are now collected or ignored. We could still have it so these metrics are collected by default. WDYT?

We can, but to be honest I'm not sure there is a use case for switching between short/detailed metrics mode at runtime. The latter one most likely would be used for low level troubleshooting during development (i.e. when some tricky Falco modifications are not playing nice), but how could they be useful in the field?

erthalion:
Ok, don't see why not as well, so lets try it.

collector/lib/CollectorStatsExporter.cpp
```diff
@@ -168,6 +152,12 @@ void CollectorStatsExporter::run() {
   prometheus::Gauge* lineage_std_dev = &collectorProcessLineageInfo.Add({{"type", "std_dev"}});
   prometheus::Gauge* lineage_avg_string_len = &collectorProcessLineageInfo.Add({{"type", "lineage_avg_string_len"}});
 
+  // Following counters are used for detailed metrics
+  prometheus::Gauge* userspaceEvents;
+  prometheus::Family<prometheus::Gauge>* collectorTypedEventCounters;
+  prometheus::Family<prometheus::Gauge>* collectorTypedEventTimesTotal;
+  prometheus::Family<prometheus::Gauge>* collectorTypedEventTimesAvg;
```
Molter73, line 159: We might want to ensure these are initialized as `nullptr`, doesn't seem like it could reach a point where they are used uninitialized, but better be safe than sorry.
```suggestion
  prometheus::Gauge* userspaceEvents = nullptr;
  prometheus::Family<prometheus::Gauge>* collectorTypedEventCounters = nullptr;
  prometheus::Family<prometheus::Gauge>* collectorTypedEventTimesTotal = nullptr;
  prometheus::Family<prometheus::Gauge>* collectorTypedEventTimesAvg = nullptr;
```

docs/troubleshooting.md
```diff
@@ -323,7 +323,10 @@ Prometheus name: rox_collector_events_typed
 Units: microseconds
 ```
 
-For each syscall, and in each direction, the total time consummed by every step ("process", "parse") is available, as well as the computed average duration in micro-second.
+For each syscall, and in each direction, the total time consumed by every step
+("process", "parse") is available, as well as the computed average duration in
+micro-second. These metrics are enabled via `DETAILED_METRICS` environment
```
Molter73, line 328: Believe this should be `ROX_COLLECTOR_ENABLE_DETAILED_METRICS` now
```suggestion
micro-second. These metrics are enabled via `ROX_COLLECTOR_ENABLE_DETAILED_METRICS` environment
```

docs/references.md
```diff
@@ -69,6 +69,10 @@ workloads it's the main part of memory consumption, this value effectively
 translates into the upper limit for memory usage. Note, that Falco puts it's
 own upper limit on top of that, which is 2^17.
 
+* `DETAILED_METRICS`: Specified whether to expose per-syscall metrics. This
```
Molter73, line 72: Same as previous comment
```suggestion
* `ROX_COLLECTOR_ENABLE_DETAILED_METRICS`: Specified whether to expose per-syscall metrics. This
```

Molter73 (approved):
LGTM! Left a few comments, but nothing major left IMO.

PR #1580, opened 2024-02-28
 [DO NOT MERGE]: Drivers build branch 

Stringy:
## Description

This branch is the new drivers build branch, containing everything related to drivers. All other parts of the workflow have been removed, including all image builds, and all tests. This is simply responsible for building new drivers.

The idea and considerations are outlined [here](https://docs.google.com/document/d/1__SppYO4nrtxo6UtsJH7FkPXMCNB2pc2Gpc9hPHpqGU/edit#heading=h.pgt7ygw5xvwz)

test plain built
test with cpaas steps enabled


.github/workflows/drivers.yml
```diff
@@ -58,7 +49,7 @@ jobs:
           CHECKOUT_BEFORE_PATCHING: false
 
         run: |
-          git fetch
+          git fetch -a
```
Molter73, line 52: Are we overkilling here? Why the `-a` if we've already checked out with `fetch-depth: 0`?

.github/workflows/drivers.yml
```diff
@@ -25,21 +25,10 @@ env:
...
-    if: startsWith(github.ref_name, 'release-') || github.ref_type == 'tag'
-    steps:
-      - run: echo Drivers Build
-
   split-tasks:
     runs-on: ubuntu-latest
     if: |
-      (github.event_name == 'push' && github.ref_name == 'master') ||
+      (github.event_name == 'push' && github.ref_name == 'drivers') ||
       github.event_name == 'pull_request'
```
Molter73, line 32: We can probably get rid of this condition altogether, there's only one calling workflow and that only runs on push to drivers and PRs (I don't expect us to issue tags against this branch)

.github/workflows/main.yml
```diff
@@ -1,12 +1,9 @@
...
 
 on:
   push:
     branches:
-      - master
-      - release-*
-    tags:
-      - 3.*.*
+      - drivers
   pull_request:
```
Molter73, line 7: Do we want to limit the workflow to PRs against the drivers branch?
```suggestion
  pull_request:
    branches:
    - drivers
```

.github/workflows/drivers.yml
Stringy: ah yeah this was me being a bit overzealous when the `git show` command wasn't working... I'll fix it!

.github/workflows/integration-tests-vm-type.yml
```diff
@@ -42,7 +42,7 @@ jobs:
     env:
       COLLECTOR_IMAGE: quay.io/rhacs-eng/collector:${{ inputs.collector-tag }}
       COLLECTOR_QA_TAG: ${{ inputs.collector-qa-tag }}
-      TEST_IMAGE: quay.io/rhacs-eng/qa-multi-arch:collector-tests-${{ inputs.collector-tag }}
+      TEST_IMAGE: quay.io/rhacs-eng/qa-multi-arch:collector-tests-3.18.0
```
Molter73, line 45: Do we need to roll this change back? I see `main.yml` and `nightly-tests.yml` are passing `3.18.0-slim` as the version used to `integration-tests.yml` and that should forward it to this workflow.

Molter73 (commented):
Just one more comment, but other than that LGTM! 

Not approving so we don't accidentally merge though.

.github/workflows/integration-tests-vm-type.yml
Stringy: This is a bit trickier, because `inputs.collector-tag` contains the `-slim` suffix, which is not how we tag the test image. I think there's a couple of approaches we can use: 
- add the -slim suffix to the collector image here and revert to using `inputs.collector-tag` for the test image
- we can add a new input to provide the test image tag directly (which I've actually done [in the other PR](https://github.com/stackrox/collector/pull/1537/files#diff-f2221670add2a7da88201d6683873ebbc6571df15a32a427edf160b4c2c29cca) so it might be nice for consistency)

WDYT?

Stringy:
Closing now in favour #1667

PR #1583, opened 2024-03-01
[tests] Use a specific logs entry to detect Collector readiness

ovalenti:
## Description

Emitting a log statement after Collector has completed its startup procedure (including connecting to Sensor)
will prevent loosing some signals during tests.

## Checklist


robbycochran:
Could this be done with a message over GRPC? or identification of the first message that indicates the probe has been loaded successfully?

PR #1585, opened 2024-03-04
Split slim builds in local gha runner and remote vm

Molter73:
## Description

This separation should make it a little easier to follow the workflows, since each of them will only have the steps needed for the type of build taking place. Additional motivation for this is in having better handling of external PRs, since a PR from a fork will not have the required credentials to create a remote VM in our GCP projects.

For the time being, this is very much some personal investigating and team feedback is very much needed.

## Checklist

## Testing Performed
- [x] Run with the `run-multiarch-builds` label
- [x] Run without the `run-multiarch-builds` label.

### DIFF
--- a/.github/workflows/collector-slim.yml
+++ b/.github/workflows/collector-slim.yml
@@ -26,16 +26,41 @@ env:
   ANSIBLE_CONFIG: ${{ github.workspace }}/ansible/ansible.cfg
 
 jobs:
+  setup-environment:
+    name: Setup environment
+    runs-on: ubuntu-latest
+    outputs:
+      collector-append-cid: ${{ steps.vars.outputs.collector-append-cid || false }}
+      trace-sinsp-events: ${{ steps.vars.outputs.trace-sinsp-events || false }}
+      address-sanitizer: ${{ steps.vars.outputs.address-sanitizer || false }}
+    steps:
+      - name: Checks PR, main and release branches
+        id: vars
+        run: |
+          if [[ "${{ github.event_name }}" == 'pull_request' ]]; then
+            echo "collector-append-cid=true" >> "$GITHUB_OUTPUT"
+            echo "trace-sinsp-events=1" >> "$GITHUB_OUTPUT"
+
+            if [[ "${{ contains(github.event.pull_request.labels.*.name, 'address-sanitizer') }}" == "true" ]]; then
+              echo "address-sanitizer=true" >> "$GITHUB_OUTPUT"
+            fi
+          fi
+
   build-collector-image:
     name: Build the collector slim image
     runs-on: ubuntu-latest
+    needs:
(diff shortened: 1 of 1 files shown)


.github/workflows/collector-slim.yml
```diff
@@ -73,67 +152,27 @@ jobs:
...
-            --limit localhost \
-            -e arch='${{ matrix.arch }}' \
-            -e @'${{ github.workspace }}/ansible/secrets.yml' \
-            ansible/ci-build-collector.yml
-
-      - name: Build s390x image
-        if: |
-          (github.event_name == 'push' && matrix.arch == 's390x') ||
-          (contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds') && matrix.arch == 's390x')
+          cat << EOF > ${{ github.workspace }}/ansible/secrets.yml
```
JoukoVirtanen, line 160: Why is this duplicated?

Molter73: I considered using a single job for creating the file and then propagating it via an artifact, problem is that the file has secrets in it and I'd rather not have those be persisted in a file stored as an artifact. 

We might be able to come up with an alternative, the file itself provides 2 types of variables, some are secrets, some others are runtime variables that aren't secrets. We could use 2 separate files, add a secret to GHA that is the YAML contents of the secrets we need, then dump that directly into one file in the build job and use a separate job to create a file with the runtime variables we need and propagate that as an artifact. This all sounds like a whole bunch of effort when all we need is just a single duplicated step though. We could also abstract the step into an action, but not sure how much that is gonna save, since we still need to explicitly pass secrets into the action.

From my point of view, this small duplication is OK for the time being, WDYT?

Stringy (approved):
One minor suggestion about the environment step, but otherwise LGTM!

.github/workflows/collector-slim.yml
```diff
@@ -26,16 +26,41 @@ env:
...
+        run: |
+          if [[ "${{ github.event_name }}" == 'pull_request' ]]; then
+            echo "collector-append-cid=true" >> "$GITHUB_OUTPUT"
+            echo "trace-sinsp-events=1" >> "$GITHUB_OUTPUT"
+
+            if [[ "${{ contains(github.event.pull_request.labels.*.name, 'address-sanitizer') }}" == "true" ]]; then
+              echo "address-sanitizer=true" >> "$GITHUB_OUTPUT"
+            fi
+          fi
+
```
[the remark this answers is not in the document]
Molter73: Good call! I'll change it now.

.github/workflows/collector-slim.yml
JoukoVirtanen: I think this is okay for now. Maybe this can be fixed later.

PR #1588, opened 2024-03-05
Troubleshoot support_package build issues

ovalenti:
## Description

Support package build currently fails with a strange condition:
- the logs do not contain the end of the build
- the job fails, but the step is still running afterwards


Molter73 (approved):
LGTM!

Just for completions sake, 4.0 has been EOL since November 3rd and 4.1 since December 29th (https://access.redhat.com/support/policy/updates/rhacs)

PR #1590, opened 2024-03-07
Fixes duplicate driver artifacts when build uses multiple GHA runners

Stringy:
## Description

When the various GHA actions were updated, I missed a couple of breaking changes in the drivers build, specifically that multiple artifacts with the same name are no longer supported. This PR handles that case by making the names unique and ensuring they're downloaded appropriately.

## Checklist

## Testing Performed

CI with multiple driver runners should do it.


ovalenti (approved):
We had a successful build and test of for all vm types :partying_face: 

I think it is good to merge, thanks !

PR #1593, opened 2024-03-08
Native multiarch image builds

Molter73:
## Description

Rework de builder image workflow in a way similar to the slim image build on #1585. Move multiarch image builds so they are built as fast as possible. After some testing, most multiarch images are substantially faster on remote VMs, except for the slim image for ppc64le, which takes longer than building on GHA due to it taking a long time to provision.

There's a couple potential follow ups to this PR:
- Improve ppc64le VMs provisioning times.
- Unify builder and slim image builds so VMs can be reused when both images are needed to be built. This can be achieved by either moving more logic into ansible or brute forcing GHA.

## Checklist

## Testing Performed
- [x] Run with no labels. ([CI run](https://github.com/stackrox/collector/actions/runs/8296681816?pr=1593))
- [x] Run with the `build-builder-image` label. ([CI run](https://github.com/stackrox/collector/actions/runs/8297092431?pr=1593))
- [x] Run with the `run-multiarch-builds` label. ([CI run](https://github.com/stackrox/collector/actions/runs/8295721795?pr=1593))
- [x] Run with both the `build-builder-image` and `run-multiarch-builds` labels. ([CI run](https://github.com/stackrox/collector/actions/runs/8293080107?pr=1593))

### DIFF
--- a/.github/workflows/collector-builder.yml
+++ b/.github/workflows/collector-builder.yml
@@ -11,7 +11,7 @@ on:
     outputs:
       collector-builder-tag:
         description: The builder tag used by the build
-        value: ${{ jobs.builder-setup-environment.outputs.collector-builder-tag }}
+        value: ${{ jobs.builder-needs-rebuilding.outputs.collector-builder-tag }}
 
 env:
   COLLECTOR_TAG: ${{ inputs.collector-tag }}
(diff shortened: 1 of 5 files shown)


.github/workflows/collector-builder.yml
```diff
@@ -36,7 +36,23 @@ jobs:
               - builder/install/**
               - builder/third_party/**
               - builder/Dockerfile
-              - .github/workflows/collector-builder.yml
```
Molter73, line 39: - [ ] Add this line back before merging.

Molter73:
Closing since the PR has gotten very stale.

PR #1599, opened 2024-03-14
Block probes <=2.9.0 for linux >=6.7

ovalenti:
## Description

6.7 fix is only available from 2.9.1 and on, for now.

## Checklist


Molter73 (approved):
LGTM!

PR #1604, opened 2024-03-15
ROX-18383: remove upstream drivers from support packages

Stringy:
## Description

As part of the 4.4 release, upstream drivers are being removed from full images as well as support packages, relying instead on downstream built drivers.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)

### DIFF
--- a/.github/workflows/support-packages.yml
+++ b/.github/workflows/support-packages.yml
@@ -74,13 +74,10 @@ jobs:
           empty=()
           for mod_ver_dir in /tmp/support-packages/output/*; do
             mod_ver="$(basename "$mod_ver_dir")"
-            for pkg in "$mod_ver_dir"/*.zip; do
-              if ! unzip -l "$pkg" | grep -q "\.gz$"; then
-                empty+=("$mod_ver")
-                rm -rf "$mod_ver_dir"
-                break
-              fi
-            done
+            if [[ -e "$mod_ver_dir/empty" ]]; then
+              empty+=("$mod_ver")
+              rm -rf "$mod_ver_dir"
+            fi
           done
 
           if ((${#empty[@]})); then
--- a/.github/workflows/support-packages.yml
+++ b/.github/workflows/support-packages.yml
@@ -68,25 +68,6 @@ jobs:
             /tmp/support-packages/output \
             gs://${{ inputs.downstream-drivers-bucket }}
 
-      - name: Check empty support packages
-        id: empty-check
-        run: |
-          empty=()
-          for mod_ver_dir in /tmp/support-packages/output/*; do
-            mod_ver="$(basename "$mod_ver_dir")"
-            if [[ -e "$mod_ver_dir/empty" ]]; then
-              empty+=("$mod_ver")
-              rm -rf "$mod_ver_dir"
-            fi
-          done
-
-          if ((${#empty[@]})); then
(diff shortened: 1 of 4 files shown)


kernel-modules/support-packages/04-create-support-packages.sh
```diff
@@ -48,7 +48,12 @@ for mod_ver_dir in "${MD_DIR}/module-versions"/*; do
     # For now we create *full* kernel support packages, not only deltas, in order to
     # support the slim collector use-case.
     # Remains to be clarified; we might provide more fine granular download options in the future.
-    gsutil -m cp "${COLLECTOR_MODULES_BUCKET}/${mod_ver}/*.gz" "$probe_dir"
+
+    # From 2.9.0 support packages only include downstream built drivers
+    if ! use_downstream_only "$mod_ver"; then
+        gsutil -m cp "${COLLECTOR_MODULES_BUCKET}/${mod_ver}/*.gz" "$probe_dir"
+    fi
+
```
Molter73, line 56: We might want to add a check after the next if block to prevent us from building empty support packages and issuing an error instead, WDYT?

kernel-modules/support-packages/04-create-support-packages.sh
```diff
@@ -58,15 +63,22 @@ for mod_ver_dir in "${MD_DIR}/module-versions"/*; do
     filename="support-pkg-${mod_ver}-$(date '+%Y%m%d%H%M%S').zip"
     latest_filename="support-pkg-${mod_ver}-latest.zip"
 
-    cp "${LICENSE_FILE}" "${probe_dir}"/LICENSE
+    if [[ -z "$(ls -A "$probe_dir")" ]]; then
+        # non-fatal here to ensure all module versions are processed
+        echo >&2 "[WARNING] no probes downloaded for ${mod_ver}"
+        touch "${package_out_dir}/empty"
```
Molter73, line 69: I think we can simply `continue` here and drop the else part of the block

Molter73: We can also simplify logic on the GHA job a bit more if we do something like `echo "${mod_ver}" >> "${OUT_DIR}/empty"`. Then we can directly look for the file and we'll have the list of empty module versions by cat'ing the contents of the file.

.github/workflows/support-packages.yml
```diff
@@ -68,6 +68,25 @@ jobs:
...
+            mod_ver="$(basename "$mod_ver_dir")"
+            if [[ -e "$mod_ver_dir/empty" ]]; then
+              empty+=("$mod_ver")
+              rm -rf "$mod_ver_dir"
+            fi
+          done
+
+          if ((${#empty[@]})); then
+            echo 2>&1 "Found empty support packages for module versions: ${empty[*]}"
+            echo "has-empty-packages=true" >> "$GITHUB_OUTPUT"
```
Molter73, line 85: Instead of using an output var, we can probably do this check in the step that starts in line 146 and directly exit with an error. This only works if you apply my other comment on using a single `empty` file, rather than one per version.

Molter73 (approved):
Provided everything works on CI, LGTM!

PR #1606, opened 2024-03-20
K8s metadata from falco

Molter73:
## Description

As part of the [ROX-22981](https://issues.redhat.com/browse/ROX-22981) issue, requiring collector to filter what that is sent depending on k8s metadata like the namespace a given container belongs to, this PR re-enables Falco's full build, allowing for the container engines to be used again. By doing this, we can now query the CRI, CRI-O and containerd sockets for additional container information and get the namespace from that, as well as additional label information that might come in handy in the near future. Also added configuration options for falco's docker and podman container engine information gathering, since these might could be helpful for either testing or if we need to support native workloads.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [x] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Manually tested the changes alongside the newly implemented k8s based tests.


collector/lib/K8s.h
```diff
@@ -0,0 +1,56 @@
...
+
+    return ss.str();
+  }
+
+  inline std::string_view GetContainerLabel(const std::string& container_id, const std::string& label) {
+    const auto container = inspector_->m_container_manager.get_container(container_id);
+    if (container == nullptr || container->m_labels.count(label) == 0) {
+      return "";
+    }
+    return container->m_labels.at(label);
```
JoukoVirtanen, line 46: Using `count` seems inefficient.

collector/lib/K8s.h
```diff
@@ -0,0 +1,56 @@
...
+    for (const auto& [key, value] : container->m_labels) {
+      ss << key << ":" << value << ",";
+    }
+
+    return ss.str();
+  }
+
+  inline std::string_view GetContainerLabel(const std::string& container_id, const std::string& label) {
+    const auto container = inspector_->m_container_manager.get_container(container_id);
+    if (container == nullptr || container->m_labels.count(label) == 0) {
```
JoukoVirtanen, line 43: Using `count` seems inefficient.

JoukoVirtanen:
I don't think `K8s` is the best name either as this code can get labels other than those associated with K8s. I like `ContainerMetadata` more.

collector/lib/K8s.h
Molter73: It's what Falco does so I didn't pay it too much care, `count` returns either 1 or 0 since `map` doesn't allow duplicate keys (see https://en.cppreference.com/w/cpp/container/map/count)

I could change it to `find`, but I think the slowest operation will be creating the copy of the string either way

(And as I write this comment I realize there are some changes missing here that are in #1614, I'll add them back and rebase the other 2 PRs)

collector/lib/ContainerMetadata.h
```diff
@@ -0,0 +1,44 @@
...
+    const char* ns = event_extractor_.get_k8s_namespace(event);
+    return ns != nullptr ? ns : "";
+  }
+
+  inline std::string GetNamespace(const std::string& container_id) {
+    return GetContainerLabel(container_id, "io.kubernetes.pod.namespace");
+  }
+
+  inline std::string GetContainerLabel(const std::string& container_id, const std::string& label) {
+    const auto& containers = *inspector_->m_container_manager.get_containers();
```
ovalenti, line 24: The Inspector service thread might be updating the container list while other threads call this through GetNamespace() (typically the instrospection API, or the filters)

Molter73: That's why I chose to use `get_containers()`, it returns a locked mutex:
https://github.com/stackrox/falcosecurity-libs/blob/fecec54dc75acde7c7ad186175a055efb1dff8fa/userspace/libsinsp/container.cpp#L286-L289

collector/lib/ContainerMetadata.h
```diff
@@ -0,0 +1,44 @@
...
+    if (containers.count(container_id) == 0) {
+      return "";
+    }
+
+    const auto& container = containers.at(container_id);
```
ovalenti, line 29: I believe that this results in 2 map lookups. Using find() and dereferencing would save some cycles.

PR #1614, opened 2024-03-27
Add an introspection endpoint for container metadata

Molter73:
## Description

Add an introspection endpoint to collector which can be used to retrieve container metadata by providing it with a container ID. At the time, the only data that is returned by the endpoint is the same container ID that was provided and the k8s namespace the container belongs to. This endpoint will enable testing of the #1606 PR without need for changes to our communication with sensor.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [x] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Manually tested the new endpoint.


collector/lib/IntrospectionEndpoint.h
```diff
@@ -0,0 +1,27 @@
...
+#include <unordered_map>
+
+namespace collector {
+
+using QueryParams = std::unordered_map<std::string, std::string>;
+
+class IntrospectionEndpoint : public CivetHandler {
+ protected:
+  static QueryParams ParseParameters(const char* queryString);
+  static std::optional<std::string> GetParameter(const QueryParams& params, const std::string& paramName);
```
ovalenti, line 15: These methods are not used in the PR. However, I think they will be soon, as we add more endpoints.

ovalenti (approved):
Looks good, and thanks for extracting the IntrospectionEndpoint common functions !

My only concern is with the underlying concurrent access via GetNamespace(). I left a comment in the other PR.

docs/troubleshooting.md
```diff
@@ -450,3 +450,86 @@ $ curl collector:8080/profile/heap
...
+80bfaff4d03b
+```
+
+You can then port-forward the collector daemonset to be able to query the
+container metadata endpoint from your localhost. Alternatively, you can exec
+into a pod with access to an HTTP tool in order to query the endpoint from
+within the cluster itself.
+
+```sh
+$ krox port-forward ds/collector 8080:8080 &
```
ovalenti, line 528: It looks like the initial part of the command line is missing.

docs/troubleshooting.md
```diff
@@ -450,3 +450,86 @@ $ curl collector:8080/profile/heap
...
+```sh
+$ kubectl -n stackrox get pods -l "app=collector" -o json \
+    | jq -r '.items[].status.containerStatuses[].containerID' \
+    | sed -e 's#containerd://##' \
+    | cut -c -12
+01e8c0454972
+80bfaff4d03b
+```
+
+You can then port-forward the collector daemonset to be able to query the
```
ovalenti, line 522: If we are going to have more endpoints, this port-forward hint could be placed in the common section a bit above.

docs/troubleshooting.md
Molter73: Dang, `krox` is an alias I have for `kubectl -n stackrox`, I'll get it fixed.

docs/troubleshooting.md
Molter73: Sounds good, I'll move it to the common section.

ovalenti (approved):
Still looks good to me :+1:

PR #1627, opened 2024-04-05
ROX-25719: Runtime control channel

ovalenti:
## Description

Collector needs a new grpc API to manage its configuration dynamically, in particular for runtime-filtering.

We propose an approach similar to the iSensor service, where a bi-directional stream is opened on top of a single grpc method. Then multiple object types can flow in each direction by making them part of a oneof list containing all off them.

The main messages should be aimed at:
- exchange capabilities.
- set the initial filtering and update it at any time.
- retrieve statistics and troubleshooting information.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests


collector/lib/runtime-control/Service.cpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+void Service::Run() {
+  writer_ = DuplexClient::CreateWithReadCallback(
+      &sensor::CollectorService::Stub::AsyncCommunicate,
+      control_channel_,
+      &client_context_,
+      std::function([this](const sensor::MsgToCollector* message) {
+        Receive(message);
+      }));
```
JoukoVirtanen, line 48: Put some logging here.

collector/lib/runtime-control/Service.cpp
```diff
@@ -0,0 +1,82 @@
...
+    if (!should_run_)
+      break;
+  }
+
+  writer_->Shutdown();
+}
+
+void Service::Receive(const sensor::MsgToCollector* message) {
+  if (!message) {
+    return;
```
JoukoVirtanen, line 64: More logging here.

PR #1626, opened 2024-04-05
ROX-23499: Force EBPF when Power + OCP4.12

ovalenti:
## Description

The modern probe fails to load on RHEL 8.6 + Power.
Temporarily fallback on ebpf in this particular case while we troubleshoot this issue.

## Checklist
- [x] Investigated and inspected CI test results
- [x] Tested on Power + OCP 4.12

PR #1629, opened 2024-04-08
ROX-23499: Force EBPF when Power + OCP4.12 (#1626)

ovalenti:
## Description

Backport #1626 for 4.4.1 patch-release


Stringy (approved):

ovalenti:
The related changes have only effect on Power, while integration-tests don't cover it.
I will merge before the tests complete since everything is going to run again on the tip of release-3.18

PR #1630, opened 2024-04-08
Limit total ring buffer size

erthalion:
## Description

Allow to specify the upper limit for all ring buffers together, to avoid issues when a node has lots of CPUs, and for every single one buffer will be allocated, eventually running out of memory. The limit could be specified via ROX_COLLECTOR_SINSP_TOTAL_BUFFER_SIZE, but the default value of 512Mi should be good enough for the basic case, where the Collector DaemonSet memory limit is 1Gi.

If at the configuration phase Collector figures out that it can't match specified limit, sinsp_cpu_per_buffer_ will be adjusted to fit under the allowed threshold.

## Checklist

**Automated testing**
  - [x] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Manual testing.

### DIFF
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -276,8 +276,8 @@ void CollectorConfig::HandleSinspEnvVars() {
 
   sinsp_cpu_per_buffer_ = DEFAULT_CPU_FOR_EACH_BUFFER;
   sinsp_buffer_size_ = DEFAULT_DRIVER_BUFFER_BYTES_DIM;
-  // the default value for sinsp_thread_cache_size_ is not picked up from
-  // Falco, but set in the CollectorConfig class.
+  // the default values for sinsp_thread_cache_size_, sinsp_total_buffer_size_
+  // are not picked up from Falco, but set in the CollectorConfig class.
 
   if ((envvar = std::getenv("ROX_COLLECTOR_SINSP_CPU_PER_BUFFER")) != NULL) {
     try {
@@ -294,6 +294,15 @@ void CollectorConfig::HandleSinspEnvVars() {
       CLOG(INFO) << "Sinsp buffer size: " << sinsp_buffer_size_;
     } catch (...) {
       CLOG(ERROR) << "Invalid buffer size value: '" << envvar << "'";
+    }
+  }
+
+  if ((envvar = std::getenv("ROX_COLLECTOR_SINSP_TOTAL_BUFFER_SIZE")) != NULL) {
+    try {
+      sinsp_total_buffer_size_ = std::stoi(envvar);
+      CLOG(INFO) << "Sinsp total buffer size: " << sinsp_buffer_size_;
+    } catch (...) {
+      CLOG(ERROR) << "Invalid total buffer size value: '" << envvar << "'";
     }
   }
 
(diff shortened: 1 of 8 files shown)


collector/lib/CollectorConfig.cpp
```diff
@@ -360,4 +369,60 @@ std::ostream& operator<<(std::ostream& os, const CollectorConfig& c) {
...
+// In this scenario, n_buffers = 150 / 1 > 128, thus the adjusted value will be
+// returned: ceil(150 / 128) = 2. Allocating one buffer per 2 CPU cores will
+// allow us to fit into the total limit: (150 / 2) * 8Mi < 1Gi.
+//
+unsigned int CollectorConfig::GetSinspCpuPerBuffer() const {
+  unsigned int n_buffers, max_n_buffers;
+
+  if (sinsp_buffer_size_ == 0) {
+    CLOG(WARNING) << "Trying to calculate cpu-per-buffer without"
+                     "initialized buffer size. Return raw value.";
```
Molter73, line 389: We might want to let users know what that `raw value` is
```suggestion
    CLOG(WARNING) << "Trying to calculate cpu-per-buffer without "
                     "initialized buffer size. Defaulting to " << sinsp_cpu_per_buffer_ << " bytes.";
```

collector/lib/CollectorConfig.cpp
```diff
@@ -360,4 +369,60 @@ std::ostream& operator<<(std::ostream& os, const CollectorConfig& c) {
...
+    return sinsp_cpu_per_buffer_;
+  }
+
+  if (host_config_.GetNumPossibleCPUs() == 0) {
+    CLOG(WARNING) << "Trying to calculate cpu-per-buffer without"
+                     "number of possible CPUs. Return raw value.";
```
Molter73, line 395: Ditto
```suggestion
    CLOG(WARNING) << "Trying to calculate cpu-per-buffer without "
                     "number of possible CPUs. Defaulting to " << sinsp_cpu_per_buffer_ << " bytes.";
```

collector/lib/CollectorConfig.cpp
```diff
@@ -360,4 +369,60 @@ std::ostream& operator<<(std::ostream& os, const CollectorConfig& c) {
...
+    return sinsp_cpu_per_buffer_;
+  }
+
+  // Round to the larger value, since one buffer will be allocated even if the
+  // last group of CPUs is less than sinsp_cpu_per_buffer_
+  n_buffers = std::ceil((host_config_.GetNumPossibleCPUs() / sinsp_cpu_per_buffer_));
```
Molter73, line 401: Both sides of the division in `std::ceil` are `unsigned int`, so that would take precedence and `std::ceil` wouldn't do much, right? You might want to cast one of them to force the result to be a float or double, or get rid of `std::ceil` altogether.

collector/lib/CollectorConfig.cpp
```diff
@@ -360,4 +369,60 @@ std::ostream& operator<<(std::ostream& os, const CollectorConfig& c) {
...
+  max_n_buffers = std::ceil(sinsp_total_buffer_size_ / sinsp_buffer_size_);
+
+  // Baseline case, nothing to adjust
+  if (n_buffers <= max_n_buffers)
+    return sinsp_cpu_per_buffer_;
```
Molter73, line 406: [nitpick] Please use braces around single line if bodies too, it's a small thing, but it brings me peace of mind.
```suggestion
  if (n_buffers <= max_n_buffers) {
    return sinsp_cpu_per_buffer_;
  }
```

Stringy (approved):

PR #1636, opened 2024-04-18
Exclude unneeded progs

erthalion:
## Description

Reduce number of bpf programs we include into the final probe. This helps to reduce the probe size, and as a side effect helps fighting debug log truncation due to its large size (it contains relocations for all of those progs).

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Manual testing.


collector/CMakeLists.txt
```diff
@@ -103,6 +103,12 @@ set(SCAP_HOST_ROOT_ENV_VAR_NAME "COLLECTOR_HOST_ROOT" CACHE STRING "Host root en
 
 set(BUILD_LIBSCAP_MODERN_BPF ON CACHE BOOL "Enable modern bpf engine" FORCE)
 
-set(MODERN_BPF_EXCLUDE_PROGS "^(openat2|ppoll|setsockopt|clone3|io_uring_setup|nanosleep)$" CACHE STRING "Set of syscalls to exclude from modern bpf engine " FORCE)
+# Exclude unneeded progs, generated in a straightforward way as follow:
+# $ bpftool prog | grep tracing | awk '{print $4}' > all_progs.list
+# $ grep -v -f progs_in_use.list all_progs.list > collector/excluded_progs.list
+file(READ "excluded_progs.list" EXCLUDED_PROGS)
+STRING(REGEX REPLACE "\n" "" EXCLUDED_PROGS "${EXCLUDED_PROGS}")
```
Molter73, line 110: Oh Lord have mercy on our souls :repenting_intensifies:

collector/excluded_progs.list
```diff
@@ -0,0 +1,140 @@
+^(
```
Molter73, line 1: On a somewhat more serious note, could we change this to be a straight up list of progs, then use something like `JOIN` to put them together `|` and simply add the `^(` and `)$` afterwards?
https://cmake.org/cmake/help/latest/command/string.html#join

erthalion: Yeah, why not.

robbycochran:
How do we keep this synchronized when either new system calls are supported in Falco or new system calls are added to collector (`progs_in_use`)?

PR #1642, opened 2024-04-23
Introduce USDT trace points

erthalion:
## Description

Add systemtap into the builder to allow defining USDT, and add one to capture events when a new process was discovered.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Manual testing.

### DIFF
--- a/collector/lib/ProcessSignalHandler.cpp
+++ b/collector/lib/ProcessSignalHandler.cpp
@@ -35,15 +35,15 @@ bool ProcessSignalHandler::Stop() {
 
 SignalHandler::Result ProcessSignalHandler::HandleSignal(sinsp_evt* evt) {
   const auto* signal_msg = formatter_.ToProtoMessage(evt);
-  const char* name = signal_msg->signal().process_signal().name().c_str();
-  const int pid = signal_msg->signal().process_signal().pid();
-
-  DTRACE_PROBE2(collector, process_signal_handler, name, pid);
 
   if (!signal_msg) {
     ++(stats_->nProcessResolutionFailuresByEvt);
     return IGNORED;
   }
+
+  const char* name = signal_msg->signal().process_signal().name().c_str();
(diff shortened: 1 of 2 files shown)


Stringy (approved):
LGTM! I can only hope that this will mean we get USDT probes _everywhere_ 😄

collector/lib/ProcessSignalHandler.cpp
```diff
@@ -33,6 +35,11 @@ bool ProcessSignalHandler::Stop() {
 
 SignalHandler::Result ProcessSignalHandler::HandleSignal(sinsp_evt* evt) {
   const auto* signal_msg = formatter_.ToProtoMessage(evt);
+  const char* name = signal_msg->signal().process_signal().name().c_str();
```
Stringy, line 38: now that I look again, I think we want to put this after line 43 (after we've checked the `signal_msg` pointer)

collector/lib/ProcessSignalHandler.cpp
```diff
@@ -33,6 +35,11 @@ bool ProcessSignalHandler::Stop() {
...
+  const int pid = signal_msg->signal().process_signal().pid();
+
+  DTRACE_PROBE2(collector, process_signal_handler, name, pid);
+
   if (!signal_msg) {
```
ovalenti, line 43: This test suggests that signal_msg may sometimes be NULL.

ovalenti (approved):
Looks good :+1:

robbycochran:
Will this be part of release builds? Do you need to make any changes to midstream or konflux environments?

erthalion:
> Will this be part of release builds? Do you need to make any changes to midstream or konflux environments?

Indeed, Konflux is failing now. From what I see it doesn't use the builder image, this should fix it: https://github.com/stackrox/collector/pull/1646

PR #1645, opened 2024-04-24
Prevent waiting infinitely for processes

Molter73:
## Description

Calling time.After in a for-select loop causes a new channel to be spawned with a new timeout, so using it alongside a ticker causes the timeout to never occur. Assigning the channel to a separate variable and using that on the loop should take care of the issue. See [this SO answer](https://stackoverflow.com/a/66003389) for more details.

## Checklist

## Testing Performed

Fix is simple enough that shouldn't need testing outside CI.


Stringy (approved):
LGTM!

PR #1648, opened 2024-04-26
Use event watchers for namespace events

Molter73:
## Description

Listing namespace events is failing to get all events, limiting how useful they can be when debugging issues on the K8S based tests. This change makes it so goroutines are spawned with event watchers and events are continuously piped into output files, from the testing I've done this seems to be the most reliable method to always get useful events dumped.

There's also a minor refactor on the `k8s_namespace.go` file so that operations related to the target namespace are done closer together.

The trigger for this change was a CI run on the master branch that had no events on the output log, I managed to reproduce a similar error by attempting to run with a collector image that doesn't exist, causing it to go into pull backoff error. With this change the watchers still dump events stating the pod has gone into this state.

## Checklist

## Testing Performed

Tested manually with a non-existing collector image, as explained in the description section.

### DIFF
--- a/integration-tests/pkg/executor/executor_k8s.go
+++ b/integration-tests/pkg/executor/executor_k8s.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/stackrox/collector/integration-tests/pkg/common"
@@ -206,12 +207,12 @@ func (e *K8sExecutor) CreateNamespaceEventWatcher(testName, ns string) (watch.In
 		return nil, err
 	}
 
-	go func(watcher watch.Interface) {
-		logFile, err := common.PrepareLog(testName, ns+"-events.jsonl")
-		if err != nil {
-			fmt.Printf("Failed to open event file: %s\n", err)
-			return
-		}
+	logFile, err := common.PrepareLog(testName, ns+"-events.jsonl")
+	if err != nil {
+		return nil, err
+	}
+
+	go func(watcher watch.Interface, logFile *os.File) {
 		defer logFile.Close()
 
 		for event := range watcher.ResultChan() {
@@ -228,7 +229,7 @@ func (e *K8sExecutor) CreateNamespaceEventWatcher(testName, ns string) (watch.In
 				return
 			}
 		}
-	}(watcher)
+	}(watcher, logFile)
 
 	return watcher, nil
 }
--- a/integration-tests/pkg/executor/executor_k8s.go
+++ b/integration-tests/pkg/executor/executor_k8s.go
@@ -9,6 +9,7 @@ import (
 	"github.com/stackrox/collector/integration-tests/pkg/common"
 	coreV1 "k8s.io/api/core/v1"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


integration-tests/pkg/executor/executor_k8s.go
```diff
@@ -199,29 +200,35 @@ func (e *K8sExecutor) CapturePodConfiguration(testName, ns, podName string) erro
...
-	defer logFile.Close()
-
-	for _, event := range events.Items {
-		eventJson, err := json.Marshal(event)
+	go func(watcher watch.Interface) {
+		logFile, err := common.PrepareLog(testName, ns+"-events.jsonl")
 		if err != nil {
-			return err
+			fmt.Printf("Failed to open event file: %s\n", err)
+			return
```
ovalenti, line 213: The log file could be opened outside of the coroutine so that the test could fail if there is a problem with the file creation.

ovalenti (approved):
This is going to be super useful when troubleshooting.

Just a nitpick comment below

Molter73:
k8s tests have failed with a full image, the last event shows the collector image started being pulled but didn't finish before the timeout was reached. For the time being, I think I'll make it so CI explicitly pulls the image before running the tests, causing the playbook to push the image into the kind cluster and giving a more deterministic time to spawn pods. If in the future we decide to expand on these tests so they can be executed on environments different to KinD, we'll probably need to expand timeouts when waiting for images to be pulled.

<details>
<summary>Last k8s event captured</summary>

```json
{
  "Type": "ADDED",
  "Object": {
    "metadata": {
      "name": "collector.17c9d8b0a2002675",
      "namespace": "collector-tests",
      "uid": "75c3b3f6-f1af-4938-bb57-946cea3fa3ac",
      "resourceVersion": "467",
      "creationTimestamp": "2024-04-26T13:49:46Z",
      "managedFields": [
        {
          "manager": "kubelet",
          "operation": "Update",
          "apiVersion": "v1",
          "time": "2024-04-26T13:49:46Z",
          "fieldsType": "FieldsV1",
          "fieldsV1": {
            "f:count": {},
            "f:firstTimestamp": {},
            "f:involvedObject": {},
            "f:lastTimestamp": {},
            "f:message": {},
            "f:reason": {},
            "f:reportingComponent": {},
            "f:reportingInstance": {},
            "f:source": {
              "f:component": {},
              "f:host": {}
            },
            "f:type": {}
          }
        }
      ]
    },
    "involvedObject": {
      "kind": "Pod",
      "namespace": "collector-tests",
      "name": "collector",
      "uid": "aa35f13d-3bb0-4eed-84f7-83c09fc74d98",
      "apiVersion": "v1",
      "resourceVersion": "460",
      "fieldPath": "spec.containers{collector}"
    },
    "reason": "Pulling",
    "message": "Pulling image \"quay.io/stackrox-io/collector:3.18.x-180-g45cb84310f\"",
    "source": {
      "component": "kubelet",
      "host": "collector-tests-control-plane"
    },
    "firstTimestamp": "2024-04-26T13:49:46Z",
    "lastTimestamp": "2024-04-26T13:49:46Z",
    "count": 1,
    "type": "Normal",
    "eventTime": null,
    "reportingComponent": "kubelet",
    "reportingInstance": "collector-tests-control-plane"
  }
}

```
</detail>

ovalenti (approved):

PR #1649, opened 2024-04-26
Limit inclusion forwarding of the falco headers

ovalenti:
## Description

This is to avoid unexpected conflicts when updating falco, and because we plan to decouple Collector from the probe implementation (makes it explicit which parts make direct use of types).

# Isolation strategy

- Everywhere possible, the `sinsp.h` includes have been moved from the header side to the implementation (.cpp)
- Only EventExtractor was kept with this include in its own header, as its interface is rooted in falco types.
- Users of EventExtractor then forward declare it and include the EventExtractor header from their implementation.

## Checklist

### DIFF
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -2,6 +2,8 @@
 
 #include <sstream>
 
+#include <libsinsp/sinsp.h>
+
 #include "CollectorArgs.h"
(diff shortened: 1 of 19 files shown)


JoukoVirtanen (approved):
I have tested these changes here https://github.com/stackrox/collector/pull/1655/files

PR #1655, opened 2024-05-02
Ovalenti/isolate falco headers with jv features map

JoukoVirtanen:
## Description

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)

### DIFF
--- a/collector/lib/ResourceSelector.cpp
+++ b/collector/lib/ResourceSelector.cpp
@@ -1,7 +1,3 @@
-//
-// Created by Robby Cochran on 3/28/24.
-//
-
 #include "ResourceSelector.h"
 
 #include <regex>
--- a/collector/lib/ResourceSelector.cpp
+++ b/collector/lib/ResourceSelector.cpp
@@ -1,0 +1,143 @@
+//
+// Created by Robby Cochran on 3/28/24.
+//
+
+#include "ResourceSelector.h"
+
+#include <regex>
+
+#include "Hash.h"
+#include "Logging.h"
+
+namespace collector {
+
+UnorderedMap<std::string, std::regex> ResourceSelector::regexMap_;
+
+std::regex ResourceSelector::GetOrCompileRegex(const std::string& str) {
+  auto pair = regexMap_.find(str);
+  if (pair != regexMap_.end()) {
+    return pair->second;
+  } else {
+    std::regex pattern(str);
+    regexMap_[str] = pattern;
+    return pattern;
+  }
+}
+
+bool ResourceSelector::IsRuleValueFollowed(const storage::RuleValue& value, const std::string& resource_name) {
+  if (value.match_type() == storage::REGEX) {
+    std::regex pattern = GetOrCompileRegex(value.value());
+    return std::regex_match(resource_name, pattern);
+  } else {
+    return (resource_name == value.value());
+  }
+}
+
+bool ResourceSelector::IsRuleFollowed(const storage::SelectorRule& rule, const std::string& resource_name) {
+  const auto& values = rule.values();
+
+  if (values.size() == 0) {
+    return true;
+  }
+
+  auto op = [resource_name](const auto& value) -> bool {
+    return IsRuleValueFollowed(value, resource_name);
+  };
+
+  if (rule.operator_() == storage::BooleanOperator::OR) {
+    return std::any_of(values.begin(), values.end(), op);
+  } else if (rule.operator_() == storage::BooleanOperator::AND) {
+    return std::all_of(values.begin(), values.end(), op);
+  }
+
+  return false;
+}
+
+bool ResourceSelector::IsResourceInResourceSelector(const storage::ResourceSelector& rs, const std::string& resource_type, const std::string& resource_name) {
+  const auto& rules = rs.rules();
+
+  if (rules.size() == 0) {
+    return true;
+  }
+
+  return std::all_of(rules.begin(), rules.end(), [resource_type, resource_name](const auto& rule) -> bool {
+    return rule.field_name() != resource_type || IsRuleFollowed(rule, resource_name);
+  });
+}
+
+bool ResourceSelector::IsNamespaceInResourceSelector(const storage::ResourceSelector& rs, const std::string& ns) {
+  return IsResourceInResourceSelector(rs, "Namespace", ns);
+}
+
+bool ResourceSelector::IsClusterInResourceSelector(const storage::ResourceSelector& rs, const std::string& cluster) {
+  return IsResourceInResourceSelector(rs, "Cluster", cluster);
+}
+
+bool ResourceSelector::IsNamespaceSelected(const storage::ResourceCollection& rc, const std::string& ns) {
+  const auto& resource_selectors = rc.resource_selectors();
+
+  if (resource_selectors.size() == 0) {
+    return true;
+  }
+
+  return std::any_of(resource_selectors.begin(), resource_selectors.end(), [ns](const auto& rs) -> bool {
+    return IsNamespaceInResourceSelector(rs, ns);
+  });
+}
+
+bool ResourceSelector::AreClusterAndNamespaceSelected(const storage::ResourceCollection& rc, const std::string& cluster, const std::string& ns) {
+  const auto& resource_selectors = rc.resource_selectors();
+
+  if (resource_selectors.size() == 0) {
+    return true;
+  }
+
+  return std::any_of(resource_selectors.begin(), resource_selectors.end(), [cluster, ns](const auto& rs) -> bool {
+    return IsClusterInResourceSelector(rs, cluster) && IsNamespaceInResourceSelector(rs, ns);
+  });
+}
+
+bool ResourceSelector::AreClusterAndNamespaceSelected(const storage::ResourceCollection& rc, const UnorderedMap<std::string, storage::ResourceCollection> rcMap, const std::string& cluster, const std::string& ns) {
+  if (AreClusterAndNamespaceSelected(rc, cluster, ns)) {
+    return true;
+  }
+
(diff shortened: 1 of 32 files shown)


collector/lib/runtime-control/Config.h
```diff
@@ -0,0 +1,54 @@
...
+  bool WaitUntilInitialized(unsigned int timeout_ms);
+
+  void Update(const storage::RuntimeFilteringConfiguration& msg);
+  bool IsFeatureEnabled(uint64_t bitMask, storage::RuntimeFilterFeatures feature);
+  bool IsFeatureEnabled(std::string cluster, std::string ns, std::string container_id, storage::RuntimeFilterFeatures feature);
+  bool IsFeatureEnabled(std::string cluster, std::string ns, storage::RuntimeFilterFeatures feature);
+  void ConfigMessageToConfig(const storage::RuntimeFilteringConfiguration& msg);
+
+ private:
+  std::mutex mutex_;
```
Molter73, line 37: I'm very confused by this mutex, it only seems to be held on `WaitUntilInitialized` and `ConfigMessageToConfig`, the latter makes changes to `default_status_map_` and `config_by_feature_`, both of which are read by `IsFeatureEnabled` and there's no sign of the mutex being locked there, this smells of race conditions.

I think what you actually want is a `std::shared_mutex` that is `std::unique_lock`ed in methods that modify members and `std::shared_lock`ed when only reading.
https://en.cppreference.com/w/cpp/thread/shared_mutex

collector/lib/runtime-control/Config.h
```diff
@@ -0,0 +1,54 @@
...
+
+class Config {
+ public:
+  Config() {
+    for (int i = 0; i < storage::RuntimeFilterFeatures_ARRAYSIZE; i++) {
+      kFeatureFlags_[i] = 1UL << (i + 1);
+    }
+  }
+
+  static Config& GetOrCreate();
```
Molter73, line 25: [nitpick] If we want `Config` to be a singleton, the "traditional" naming for this is `GetInstance` I believe.
```suggestion
  static Config& GetInstance();
```

Molter73: You also want to make the default constructor private and `delete` all other default methods of construction so that we can't have multiple instances of the configuration floating around.

collector/lib/runtime-control/Config.cpp
```diff
@@ -0,0 +1,130 @@
...
+  for (storage::ResourceCollection rc : msg.resource_collections()) {
+    rcMap_[rc.id()] = rc;
+    CLOG(INFO) << "Adding rc.id()= " << rc.id();
+  }
+}
+
+bool Config::IsFeatureEnabled(std::string cluster, std::string ns, storage::RuntimeFilterFeatures feature) {
+  std::vector<storage::RuntimeFilter_RuntimeFilterRule> filteringRules;
+  bool defaultStatus = false;
+  auto filteringRulesPair = config_by_feature_.find(feature);
```
Molter73, line 60:
```suggestion
  const auto& filteringRulesPair = config_by_feature_.find(feature);
```

collector/lib/runtime-control/Config.cpp
```diff
@@ -0,0 +1,130 @@
...
+  if (filteringRulesPair != config_by_feature_.end()) {
+    filteringRules = filteringRulesPair->second;
```
Molter73, line 62: Is this making a copy of a vector every time we try to check if a feature is enabled? Please don't.

collector/lib/runtime-control/Config.h
```diff
@@ -0,0 +1,54 @@
...
+  std::mutex mutex_;
+  std::condition_variable condition_;
+  std::optional<storage::RuntimeFilteringConfiguration> config_message_;
+  uint64_t kFeatureFlags_[storage::RuntimeFilterFeatures_ARRAYSIZE];
+  UnorderedMap<std::string, uint64_t> namespaceFeatureBitMask_;
+  UnorderedMap<std::string, uint64_t*> containerFeatureBitMask_;
+  uint64_t SetFeatureBitMask(uint64_t bitMask, bool enabled, storage::RuntimeFilterFeatures feature);
+  void SetBitMask(std::string cluster, std::string ns);
+  void SetBitMask(std::string cluster, std::string container_id, std::string ns);
+  void SetBitMasksForNamespaces(std::string cluster);
```
Molter73, line 46: These also don't seem to need to make copies of every single argument, at best they should create a copy when they do the final storing on a member of this class.

collector/lib/runtime-control/Config.cpp
```diff
@@ -0,0 +1,130 @@
...
+  CLOG(INFO) << "ns= " << ns << " bitMask= " << bitMask;
+  namespaceFeatureBitMask_[ns] = bitMask;
+}
+
+void Config::SetBitMasksForNamespaces(std::string cluster) {
+  for (auto it = namespaceFeatureBitMask_.begin(); it != namespaceFeatureBitMask_.end();) {
+    auto& pair = *it;
+    std::string ns = pair.first;
+    SetBitMask(cluster, ns);
+    ++it;
```
Molter73, line 95:
```suggestion
  for (auto it = namespaceFeatureBitMask_.begin(); it != namespaceFeatureBitMask_.end(); ++it) {
    auto& pair = *it;
    std::string ns = pair.first;
    SetBitMask(cluster, ns);
```

collector/lib/ResourceSelector.cpp
```diff
@@ -0,0 +1,143 @@
+//
+// Created by Robby Cochran on 3/28/24.
+//
```
Molter73, line 3: Not to diminish Robby's contributions, but I think we want to remove this comment and the one on the header file. We can mention him in a commit message as a co-author.

collector/lib/runtime-control/Config.h
```diff
@@ -0,0 +1,54 @@
...
+
+  static Config& GetOrCreate();
+
+  // returns true when initialized, false for a timeout.
+  bool WaitUntilInitialized(unsigned int timeout_ms);
+
+  void Update(const storage::RuntimeFilteringConfiguration& msg);
+  bool IsFeatureEnabled(uint64_t bitMask, storage::RuntimeFilterFeatures feature);
+  bool IsFeatureEnabled(std::string cluster, std::string ns, std::string container_id, storage::RuntimeFilterFeatures feature);
+  bool IsFeatureEnabled(std::string cluster, std::string ns, storage::RuntimeFilterFeatures feature);
```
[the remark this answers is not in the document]
JoukoVirtanen: I wasn't focusing on quality. Now most things are passed by reference or inlined. Copying is avoided.

collector/lib/runtime-control/Config.cpp
```diff
@@ -0,0 +1,130 @@
...
+bool Config::WaitUntilInitialized(unsigned int timeout_ms) {
+  std::unique_lock<std::mutex> lock(mutex_);
+
+  return config_message_ || (condition_.wait_for(lock, std::chrono::milliseconds(timeout_ms)) == std::cv_status::no_timeout);
+}
+
+void Config::Update(const storage::RuntimeFilteringConfiguration& msg) {
+  std::unique_lock<std::mutex> lock(mutex_);
+
+  config_message_.emplace(msg);
```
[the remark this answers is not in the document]
JoukoVirtanen: I am not using optional for config_message_ anymore, as was the case in Olivier's PR, so this has simply become

```
config_message_ = std::move(msg);
```

Now that config_message_ is being converted to a different data structure and is not used anywhere directly, we don't need both. `config_message_` doesn't need to exist at all. Currently the following files exist `runtime-control/Config.cpp`, `runtime-control/Service/cpp`, and `ResourceSelector.cpp`. This needs to be refactored, but I am not sure how. I think I will wait until I decide that to get rid of config_message_.

collector/lib/runtime-control/Config.h
JoukoVirtanen: The mutex was added by Olivier. I haven't thought about how it should be used much yet.

collector/lib/runtime-control/Config.h
JoukoVirtanen: Done

collector/lib/runtime-control/Config.cpp
JoukoVirtanen: config_by_feature is now an array instead of a map so this no longer applies.

collector/lib/runtime-control/Config.cpp
JoukoVirtanen: See my first response. This is fixed now.

collector/lib/runtime-control/Config.h
JoukoVirtanen: See above comments. This is fixed.

collector/lib/runtime-control/Config.cpp
JoukoVirtanen: Done

collector/lib/ResourceSelector.cpp
JoukoVirtanen: Done

collector/lib/runtime-control/Config.cpp
```diff
@@ -0,0 +1,145 @@
...
+        CLOG(INFO) << "Setting default to true";
+      } else {
+        default_status_[feature] = false;
+        CLOG(INFO) << "Setting default to false";
+      }
+
+      config_by_feature_[feature].clear();
+      for (auto runtimeConfigRule : runtimeConfig.rules()) {
+        config_by_feature_[feature].push_back(runtimeConfigRule);
+      }
```
JoukoVirtanen, line 42: Setting config_by_feature this way is inefficient. Should be fixed.

PR #1656, opened 2024-05-03
Re-enable benchmark CI step

erthalion:
## Description

Currently benchmark CI step is disabled. Turn it on with following changes:

* Use berserker instread of hackbench
* Use regular ttest instead of IQR
* Compare CPU/Memory usage instead of hackbench time

#run-benchmark

## Checklist

## Testing Performed

Manual testing.


Stringy:
Just a heads up on the socat integration test failures - the socat image is using `alpine:latest` which pulls in a new version of socat that seems to consistently fail (I've not looked into precisely what's wrong with it) but we can pin the image to `alpine:3.18.4` to stick with the old and working version.

I had to do this in the syscall latency branch as well: https://github.com/stackrox/collector/pull/1640/commits/65cf13545a4bcdb1225448672b9c9ae195286c6f

.github/workflows/benchmarks.yml
```diff
@@ -34,9 +34,10 @@ jobs:
         with:
           python-version: "3.10"
 
-      - uses: 'google-github-actions/auth@v2'
+      - name: Authenticate with GCP
+        uses: 'google-github-actions/auth@v2'
         with:
-          credentials_json: '${{ secrets.GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT }}'
+          credentials_json: '${{ secrets.GOOGLE_CREDENTIALS_COLLECTOR_CI_VM_SVC_ACCT }}'
```
Stringy, line 40: Not something this this PR, but in a follow up it might make sense to rename this secret now that it's being used more widely than just VM creation. Also, is this the last use of the original `GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT` secret?

.github/workflows/benchmarks.yml
```diff
@@ -48,29 +49,39 @@ jobs:
       - name: Install python deps
         run: python3 -m pip install -r ./integration-tests/scripts/baseline/requirements.txt
 
-      - name: Calculate Baseline
+      - name: Compare with the Baseline
         run: |
-          jq -s 'flatten' perf.json > integration-tests/perf-all.json
+          jq -s 'flatten' ./container-logs/*/*/perf.json > integration-tests/perf-all.json
+
+          cat integration-tests/perf-all.json
```
Stringy, line 56: [nit] perhaps only cat the file in debug mode?
```suggestion
          if [[ "${RUNNER_DEBUG}" == "1" ]]; then
              cat integration-tests/perf-all.json
          fi
```

docs/troubleshooting.md
```diff
@@ -451,6 +451,30 @@ $ curl collector:8080/profile/heap
 The resulting profile could be processed with `pprof` to get a human-readable
 output with debugging symbols.
 
+## Benchmark CI step
+
+Whenever in doubt about performance implications of your changes, there is an
+option to run a benchmark against the PR and compare the results with the known
+baseline numbers. To do that add a "run-benchmark" label to the PR, the
```
Stringy, line 458: [nit]
```suggestion
baseline numbers. To do that, add a "run-benchmark" label to the PR. The
```

docs/troubleshooting.md
```diff
@@ -451,6 +451,30 @@ $ curl collector:8080/profile/heap
...
+performance comparison will be reported via commentary to the PR with metrics
+for CPU utilization and memory consumption. Along with the median values for
+each resource a p-value sign will be reported, which could be interpreted as
+how high are the chances that the observed difference is purely due to the
+noise. The underlying mechanism is t-test, more than 80% probability it's just
+a noise will result in green, less than that -- in red.
+
+The benchmark will be conducted with two short simultaneous workloads, you can
```
Stringy, line 466: [nit]
```suggestion
The benchmark will be conducted with two short simultaneous workloads. You can
```

docs/troubleshooting.md
```diff
@@ -451,6 +451,30 @@ $ curl collector:8080/profile/heap
...
+find the configuration for them inside the berserker integration test image.
+The baseline numbers are stored on GCS, and contains last 10 runs from the main
+branch. The reporting filters out distinctly different results with small median
+difference, to not bother without significant reasons, e.g. differences in CPU
+utilization less than 1% and in memory less than 10 MiB are ignored. Keep in
+mind, that false-positives are definitely possible due to the noisyness of the
```
Stringy, line 472: [nit]
```suggestion
mind that false-positives are definitely possible due to the noisiness of the
```

integration-tests/scripts/baseline/main.py
```diff
@@ -46,22 +46,78 @@
...
+        if k in ('Mem') and isinstance(v, str) and len(v) > 3:
+            suffix = v[-3:]
+            multiplier = multipliers.get(suffix, 0)
+
+            if multiplier == 0:
+                result[k] = 0
+            else:
+                result[k] = float(v[:-3]) * multiplier
+        else:
+            result[k] = v
```
Stringy, line 80: Is there any chance of nested data that also contains memory fields?

integration-tests/scripts/baseline/main.py
```diff
@@ -46,22 +46,78 @@
...
+def memory_parser(data):
+    result = {}
+    multipliers = {
+        "GiB": 1024,
+        "MiB": 1,
+        "KiB": 1 / 1024,
+    }
+
+    for k, v in data.items():
+        if k in ('Mem') and isinstance(v, str) and len(v) > 3:
```
Stringy, line 71: I think this either needs to be `k in ('Mem',)` to make it a tuple, or perhaps simply `k == 'Mem'` since there's only one value

integration-tests/scripts/baseline/main.py
```diff
@@ -46,22 +46,78 @@
...
+            result[k] = v
+
+    return result
+
+
+def datetime_parser(data):
+    result = {}
+
+    for k, v in data.items():
+        if k in ('Timestamp', 'LoadStartTs', 'LoadStopTs'):
```
Stringy, line 89: Do we want to validate the type of `v` here, in the same way as we do for memory?

integration-tests/scripts/baseline/main.py
```diff
@@ -209,19 +250,29 @@ def process(content):
...
-        if record.get("Metrics", {}).get("hackbench_avg_time")
+        for stats in record.get("ContainerStats")
+
+        # Filter statistics from Collector only, bounded to the timeframe when
+        # load was actually running. LoadStopTs is adjusted by 1 sec, since
+        # this timestamp is taken after the workload container was stopped, so
+        # that the actual load might have stopped slightly earlier.
+        if (stats.get("Name") == "collector" and
+            stats.get("Timestamp") > record.get("LoadStartTs") and
+            stats.get("Timestamp") < record.get("LoadStopTs") - timedelta(seconds=1))
```
Stringy, line 267: This list comprehension is pretty big now. It's fairly readable currently but I think it might be clearer if it was a set of conventional loops

integration-tests/scripts/baseline/main.py
erthalion: My understanding is that the object_hook will be invoked for every nested object, but I can verify.

integration-tests/scripts/baseline/main.py
erthalion: Right, there was something about the comma for tuples.

integration-tests/scripts/baseline/main.py
erthalion: I think the act of parsing the datetime serves as validation here. Or do you think something more rigorous is needed?

.github/workflows/benchmarks.yml
erthalion: At least on this branch there are few more examples of this secret being used. Not sure how would it be after turning off eBPF though.

integration-tests/scripts/baseline/main.py
Stringy: No, that's a fair point, it can be left as-is

integration-tests/scripts/baseline/main.py
erthalion: Yep, verified, it's applied to the nested objects as well.

Stringy (approved):
LGTM! It'll be nice to have these back

PR #1659, opened 2024-05-06
fix: update external-contributor triaging workflow

tommartensen:
Update the external-contributors workflows to add new issues and external PRs to the OSS Triaging board with project workflows instead.


Maddosaurus (approved):

erthalion:
While you're at this, would it be possible to add issues to the OSS Triaging board only in case if they're not yet present in any other? It's super annoying, when we use github issues for tracking something for our self, and it's getting automatically added to another dashboard.

tommartensen:
> While you're at this, would it be possible to add issues to the OSS Triaging
board only in case if they're not yet present in any other?

@erthalion this is possible by only adding new issues without a certain label:

- https://github.com/stackrox/collector/issues/1662 -> added to the OSS Triaging board
- https://github.com/stackrox/collector/issues/1661 -> not added to the OSS Triaging board (has the `stackrox-owned` label - I take recommendations for a better label name)

PR #1667, opened 2024-05-10
eBPF turn-off-ening

Stringy:
## Description

- Make slim and full collector images the same.
    - “full” images will not contain any drivers

- Keep legacy driver builds on master, but separate them from the main CI workflow
    - This will maintain observability while still removing complexity
    - no kernel packer changes required in this case, though we could reduce the crawl frequency

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


.github/workflows/collector.yml
```diff
@@ -66,8 +76,19 @@ jobs:
           collector_image: ${{ inputs.collector-image }}
           collector_tag: ${{ inputs.collector-tag }}
           debug_mode: ${{ github.event_name == 'pull_request' }}
+          driver_version: ${DRIVER_VERSION}
+          context_path: ${{ github.workspace }}/kernel-modules/container
```
Molter73, line 80: Bit puzzling that this is not `${{ env.CONTEXT_DRIVERS_DIR }}`, can we unify them somehow?

.github/workflows/collector.yml
```diff
@@ -39,6 +44,7 @@ jobs:
 
     env:
       PLATFORM: linux/${{ matrix.arch }}
+      CONTEXT_DRIVERS_DIR: ${{ github.workspace }}/kernel-modules/container/kernel-modules
```
Molter73, line 47: Think if we set this to download the drivers to somewhere like `${{ github.workspace }}/collector/kernel-modules` and we put the `LICENSE-kernel-modules.txt ` file in that directory as well renamed to `LICENSE`, we can change [this line in the Dockerfile](https://github.com/stackrox/collector/blob/6be77738ebe0c5f5994597fd5f630215d08505b3/collector/container/Dockerfile#L29) to directly copy the license and drivers, WDYT? (We might also need to change the Konflux dockerfile)

.github/workflows/main.yml
```diff
@@ -163,26 +120,6 @@ jobs:
...
     needs:
     - init
-    - build-collector-slim
-    - build-collector-full
+    - build-collector
     - build-test-containers
     secrets: inherit
-
-  build-support-packages:
-    uses: ./.github/workflows/support-packages.yml
```
Molter73, line 172: Can we also remove this file? Or do we want to change it to be run by the master branch as part of the `main-drivers.yml` workflow?

ansible/ci-build-collector.yml
```diff
@@ -40,10 +40,33 @@
...
+              required_layers: 0
+            path: "{{ context_path }}"
+
+      - name: Retag image
+        community.docker.docker_image:
+          name: "{{ collector_image }}-full"
+          repository: "{{ collector_image }}"
+          source: local
+
+      when: arch == "ppc64le"
```
Molter73, line 64: This goes with one of my previous comments, I don't think we need to be this fancy for the power image to get its drivers copied in. The main reason for the splitting in layers was, IIRC, that dockerhub had a sort of hard limit when pulling big layers and was causing time outs on the massive collector images. When we moved to quay.io, we had collector images that were 1GB+ on a single layer and we had no complains from users. On top of that, the power image is A LOT smaller, so I'm pretty sure we can just drop the entire layering mechanism and go for a direct copy of the kernel-modules directory.

Molter73 (commented):
The turn-off-ening?! So big a kerfuffle was never seen before!!

.github/workflows/collector.yml
Stringy: Missed this when I was fiddling around :) I've removed it now

ansible/ci-build-collector.yml
Stringy: Yeah agreed. I've consolidated all this like you suggested and it's waaaay cleaner now

.github/workflows/main.yml
Stringy: Definitely the latter. I'll fix this.

.github/workflows/main-drivers.yml
```diff
@@ -0,0 +1,59 @@
+name: Collector Drivers CI
+
+on:
```
Molter73, line 3: Think we want to add `pull_request:` here, otherwise we won't be able to test the workflow if/when it breaks. AFAICT, the safeguards for running this on PRs are still there, so it should be OK for us to add it in.

ansible/ci-build-collector.yml
```diff
@@ -59,7 +59,14 @@
 
     - name: Push slim to quay.io/stackrox-io
       community.docker.docker_image:
-        name: "{{ collector_image }}-{{ arch }}-slim"
+        name: "{{ collector_image }}-{{ arch }}"
+        push: true
+        source: local
+
+    - name: Push base to quay.io/stackrox-io
```
Molter73, line 66:
```suggestion
    - name: Push slim to quay.io/stackrox-io
```

ansible/ci-build-collector.yml
```diff
@@ -59,7 +59,14 @@
 
     - name: Push slim to quay.io/stackrox-io
```
Molter73, line 60:
```suggestion
    - name: Push to quay.io/stackrox-io
```

ansible/ci-build-collector.yml
```diff
@@ -70,6 +77,13 @@
         push: true
         source: local
 
+    - name: Push latest to quay.io/stackrox-io
+      community.docker.docker_image:
+        name: "{{ collector_image }}-{{ arch }}-slim"
```
Molter73, line 82: [nit] This doesn't make a difference since all tags will be copies of each other, but if the image we build is the one with no suffix, I feel we should use that one as the base for all these pushes for consistency if nothing else. WDYT?
```suggestion
        name: "{{ collector_image }}-{{ arch }}"
```

ansible/ci-build-collector.yml
```diff
@@ -78,26 +92,44 @@
 
     - name: Push slim to quay.io/rhacs-eng
```
Molter73, line 93:
```suggestion
    - name: Push to quay.io/rhacs-eng
```

ansible/ci-build-collector.yml
```diff
@@ -78,26 +92,44 @@
...
       community.docker.docker_image:
-        name: "{{ collector_image }}-{{ arch }}-slim"
+        name: "{{ collector_image }}-{{ arch }}"
+        repository: "{{ rhacs_eng_image }}-{{ arch }}"
+        push: true
+        source: local
+
+    - name: Push base to quay.io/rhacs-eng
```
Molter73, line 100:
```suggestion
    - name: Push slim to quay.io/rhacs-eng
```

ansible/ci-build-collector.yml
```diff
@@ -78,26 +92,44 @@
...
-    - name: Push base to quay.io/stackrox-io
+    - name: Push base to quay.io/rhacs-eng
       community.docker.docker_image:
-        name: "{{ collector_image }}-{{ arch }}-slim"
+        name: "{{ rhacs_eng_image }}-{{ arch }}"
         repository: "{{ rhacs_eng_image }}-{{ arch }}-base"
         push: true
         source: local
 
+    - name: Push base to quay.io/rhacs-eng
```
Molter73, line 114:
```suggestion
    - name: Push latest to quay.io/rhacs-eng
```

ansible/group_vars/all.yml
```diff
@@ -18,18 +18,8 @@ redhat_password: "{{ lookup('env', 'REDHAT_PASSWORD', default='') }}"
...
-  # No ebpf support for arm... Yet...?
-  - [ 'rhel-9-arm64', 'ebpf' ]
-  - [ 'ubuntu-2004-lts-arm64', 'ebpf' ]
-  - [ 'ubuntu-2204-lts-arm64', 'ebpf' ]
-  - [ 'sles-15-arm64', 'ebpf' ]
-  - [ 'fedora-coreos-stable-arm64', 'ebpf' ]
-  # eBPF on Garden Linux is not supported.
-  - [ "{{ lookup('file', 'group_vars/gardenlinux-image.txt') }}", 'ebpf' ]
-
-  - [ 'p', 'ebpf' ]
```
Molter73, line 32: Noice!

ansible/roles/create-vm/tasks/main.yml
```diff
@@ -2,12 +2,12 @@
...
 - name: Check family excluded
   set_fact:
-    excluded: "{{ [vm_family, vm_collection_method] in excluded_pairs }}"
+    excluded: "{{ [vm_family, vm_collection_method] in excluded_pairs or (test_platform != 'rhel-ppc64le' and vm_collection_method == 'ebpf') }}"
   when: vm_image is not defined
 
 - name: Check image excluded
   set_fact:
-    excluded: "{{ [vm_image, vm_collection_method] in excluded_pairs }}"
+    excluded: "{{ [vm_image, vm_collection_method] in excluded_pairs or (test_platform != 'rhel-ppc64le' and vm_collection_method == 'ebpf') }}"
```
Molter73, line 10: Maybe we can use a separate task to try and simplify these conditions a bit, WDYT?
```yml
  - name: Check eBPF excluded
    set_fact:
      excluded: "{{ test_platform != 'rhel-ppc64le' and vm_collection_method == 'ebpf' }}"

  - name: Check family excluded
    set_fact:
      excluded: "{{ [vm_family, vm_collection_method] in excluded_pairs }}"
    when: vm_image is not defined and excluded == false

  - name: Check family excluded
    set_fact:
      excluded: "{{ [vm_image, vm_collection_method] in excluded_pairs }}"
    when: vm_image is defined and excluded == false
```

ansible/roles/create-vm/tasks/main.yml
```diff
@@ -2,12 +2,12 @@
...
   when: vm_image is defined
 
 - name: Create IBM Z VM
```
Molter73, line 13: We don't need to do this in this PR, but it looks like all tasks starting with this one could be added to a block and have a global `when: excluded == false` instead of checking on every single one, (except for the very last task that run unconditionally).

collector/lib/HostHeuristics.cpp
```diff
@@ -109,7 +109,7 @@ class PowerHeuristic : public Heuristic {
       return;
     }
 
-    if (host.IsRHEL86()) {
+    if (k.kernel == 4 && k.major == 18 && k.build_id < 477) {
       CLOG(INFO) << "RHEL 8.6 on ppc64le does not support CORE_BPF, switching to eBPF collection method.";
```
Molter73, line 113: This message is good enough, but maybe we want to be a bit more specific?

kernel-modules/BLOCKLIST
```diff
@@ -92,3 +92,6 @@
 # Since kmods are not supported anyway, block these combination
 ~.*\.el8_9\..* * mod
 ~.*\.el9_4\..* ~2\.[3-9]\..* *
+
+# Block all kernels for module 2.10 and newer
+* ~2\.\d{2}\..* *
```
Molter73, line 97: Seems blocking them in the scripts didn't work? I don't particularly like us depending on the blocklist for this.

This will not stop the scripts from configuring Falco to build the eBPF probe, which means at best we'll be wasting time configuring and copying files around that we don't need and at worst upstream might break the step and we'll need to fix it for newer versions... That again, we'll not build.

Can we take another crack at blocking this in the scripts? I'm willing to give it a shot myself if needed.

.github/workflows/collector.yml
```diff
@@ -84,7 +104,7 @@ jobs:
             ansible/ci-build-collector.yml
 
   build-collector-image-remote-vm:
-    name: Build the collector slim image on a remote VM
+    name: Build Collector (Remote)
```
JoukoVirtanen, line 107:
```suggestion
    name: Build Collector on a remote VM
```

.github/workflows/main-drivers.yml
```diff
@@ -0,0 +1,59 @@
...
+    uses: ./.github/workflows/support-packages.yml
+    with:
+      branch-name: ${{ needs.init.outputs.branch-name }}
+      upstream-drivers-bucket: ${{ needs.init.outputs.drivers-bucket }}
+      downstream-drivers-bucket: ${{ needs.init.outputs.cpaas-drivers-bucket }}/x86_64
+      support-packages-bucket: ${{ needs.init.outputs.support-packages-bucket }}
+      support-packages-index-bucket: ${{ needs.init.outputs.support-packages-index-bucket }}
+      public-support-packages-bucket: ${{ needs.init.outputs.public-support-packages-bucket }}
+    if: |
+      always() &&
```
JoukoVirtanen, line 53: Is `always()` needed here?

Molter73: It's needed, there are some cases were steps are skipped but the support packages still need to be built and this ensures that (can't recall the exact case where not having it cause issues, but I did get it not so long ago).

.github/workflows/main-drivers.yml
Stringy: Agreed, but I think we only want to run it on PRs that request driver builds, rather than running it in general

Molter73: Yeah, that should be handled by the fact that we only build drivers that are missing from the cache bucket, right? Same way we handle it currently, or did you have something else in mind?

ansible/ci-build-collector.yml
Stringy: Yeah fair point - I'll fix this

ansible/roles/create-vm/tasks/main.yml
erthalion: What about doing one more step, and try to generate `excluded_pairs`? Something like:

```yaml
vm_families: ["rhel-ppc64le"]
vm_collection_methods: [ "core_bpf", "ebpf", "magic"]
custom_excluded_pairs: [...]
excluded_pairs: "{{ [vm_families | product(methods), custom_excluded_pairs] | flatten(1) | unique }}"
```

In this way we could still tread `excluded_pairs` as a single source of truth, but assemble it from blocks. Another advantage to me is that it moves us closer to the idea of declarative test matrix. What do you think?

kernel-modules/BLOCKLIST
Stringy: I wasn't that happy with how it looked in the scripts, but I'll have another go at it

ansible/roles/create-vm/tasks/main.yml
Stringy: I like this idea in principle, but it might be overkill since we're on the cusp of having a single collection method. Once we're just using core-bpf we can simply have a list of families / images that are excluded

ansible/roles/create-vm/tasks/main.yml
```diff
@@ -1,30 +1,46 @@
...
+    include_tasks: create-ppc64le-vm.yml
+    vars:
+      pi_name: "{{ vm_name }}"
+    when: vm_arch == "ppc64le"
+
+  - name: Create GCP VM
+    include_tasks: create-gcp-vm.yml
+    when: (vm_arch == "amd64" or vm_arch == "arm64")
+
+  when: excluded == false
```
erthalion, line 43: I'm a bit concerned about the fact that with this approach to answer a question "what is excluded from the tests" one have to run the pipeline (physically or virtually in their head) to answer. Is there a way to structure this in a way that will make answering this question trivial?

erthalion (commented):
After a quick look everything sounds good, thanks. The CI pipeline seems to be unhappy though, is it expected?

ansible/roles/create-vm/tasks/main.yml
Stringy: yeah that's a fair point. We could extend the current `excluded_pairs` variable to define entire platforms (e.g. `- [ 'rhel', 'ebpf' ]`) so then everything is centralised in the group_vars/all.yml file, and the first check here matches the others - `excluded: {{ [test_platform, vm_collection_method] in excluded_pairs }}` 

Ideally we'd want to sort of reverse the mechanism slightly for eBPF so power can opt-in, rather than having everything else opt-out, but that might require much more extensive changes to the way `excluded_pairs` works.

Hopefully we'll get an answer in the next couple of days about whether we can just drop eBPF altogether, at which point we can simplify back to how excluded_pairs worked to begin with, and in fact we could even drop the collection method and just exclude families or images

Stringy: for reference, this is what I mean https://github.com/stackrox/collector/pull/1677/commits/be8138db6fc6b9cbb368739fe7bd052e3d1517c0

.github/workflows/main-drivers.yml
erthalion: What is exactly understood here under safeguards? If it's only conditions inside the pipeline to not do anything, unless `no-cache` labels is set, in long term this sounds a bit fragile to me. How about keeping it on master push only, and start the development workflow with "turn it on for PRs"?

collector/lib/HostHeuristics.cpp
```diff
@@ -109,9 +109,8 @@ class PowerHeuristic : public Heuristic {
       return;
     }
 
-    if (host.IsRHEL86()) {
-      CLOG(INFO) << "RHEL 8.6 on ppc64le does not support CORE_BPF, switching to eBPF collection method.";
-      hconfig->SetCollectionMethod(CollectionMethod::EBPF);
+    if (k.kernel == 4 && k.major == 18 && k.build_id < 477) {
```
erthalion, line 112: It wont hurt to reference something here to explain why is that.

erthalion (approved):
Looks good to me. A few commentaries, but they could be addressed in follow-up PRs.

PR #1668, opened 2024-05-10
Fix ansible full build to ensure drivers are copied correctly

Stringy:
## Description

Fixes ansible to copy drivers into multiarch images, instead of just on x86

## Checklist


Molter73 (approved):
LGTM!

PR #1676, opened 2024-05-16
ROX-20232: Reduce duplication in Collector Konflux pipelines

msugakov:
## Description

Follows the pattern set in https://github.com/stackrox/stackrox/pull/10866.
YAML formatting applied from https://github.com/stackrox/scanner/pull/1515.

It's best to review this change by commits where I tried supplying meaningful messages.

## Checklist

**Automated testing**
  - ~~[ ] Added unit tests~~

~~If any of these don't apply, please comment below.~~

## Testing Performed

Only looking at CI.


tommartensen (approved):

tommartensen (approved):

msugakov:
Pending https://redhat-internal.slack.com/archives/CFMQ5C2TT/p1716389096302639

Molter73 (approved):
Full disclosure, I understand very little of what's going on in this PR, so this is a "blind trust LGTM"

.tekton/collector-component-pipeline.yaml
```diff
@@ -0,0 +1,401 @@
...
+    - name: PREFETCH_INPUT
+      value: $(params.prefetch-input)
+    - name: IMAGE_EXPIRES_AFTER
+      value: $(params.image-expires-after)
+    - name: COMMIT_SHA
+      value: $(tasks.clone-repository.results.commit)
+    - name: TARGET_STAGE
+      value: $(params.build-target-stage)
+    runAfter:
+    - prefetch-dependencies
```
erthalion, line 246: This seems to be slightly different from the original version, where `prefetch-dependencies` was run after `prepare-rhel-rpm-subscriptions`. Just out of curiosity, is the order here important?

erthalion (approved):

.tekton/collector-component-pipeline.yaml
msugakov: The order matters in general but not here.

Both `prefetch-dependencies` and `prepare-rhel-rpm-subscriptions` (as well as earlier `clone-repository` and `determine-image-tag`) are prerequisites for `build-container`. They were put sequentially because, well, we could. They can also be put parallel because they don't have a dependency on each other.

In fact, in `.tekton/collector-(pull-request|push).yaml` they were already in parallel and this caused no issue.
`.tekton/collector-slim-(pull-request|push).yaml` were simply not made consistent.

PR #1684, opened 2024-05-22
Remove drivers from Konflux images

Stringy:
## Description

Do not merge... yet.

### DIFF
--- a/collector/container/konflux.Dockerfile
+++ b/collector/container/konflux.Dockerfile
@@ -89,32 +89,8 @@ RUN ./builder/install/install-dependencies.sh && \
     ctest -V --test-dir ${CMAKE_BUILD_DIR} && \
     strip -v --strip-unneeded "${CMAKE_BUILD_DIR}/collector/collector"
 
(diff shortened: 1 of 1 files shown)


collector/container/konflux.Dockerfile
```diff
@@ -197,5 +173,3 @@ LABEL \
     com.redhat.component="rhacs-collector-container" \
     io.k8s.display-name="collector" \
     name="rhacs-collector-rhel8"
```
msugakov, line 175: Technically, these labels don't control anything in Konflux. They are informative (and are checked by EC) but nothing more. If we're to get rid of `collector-slim` at some point, I would suggest you can drop `FROM collector-common AS collector-slim` stage altogether and set `build-target-stage` to empty in `.tekton/collector-slim*.yaml` files.

This could be done later as well.

msugakov (approved):

msugakov:
In my opinion, there's nothing blocking a merge of this PR and I'd vote it to be merged since it unblocks my https://github.com/stackrox/collector/pull/1676

erthalion (approved):

msugakov:
I badly miss a jira ticket reference here.

PR #1687, opened 2024-05-23
ROX-20232: Remove all but single collector pipeline

msugakov:
## Description

This follows up on https://github.com/stackrox/collector/pull/1676 applying the approach from <https://github.com/stackrox/scanner/pull/1526>.

Also follows up on <https://github.com/stackrox/collector/pull/1684> cleaning slim stage in Dockerfile and removing slim pipelines.

I use [apply-tags task](https://redhat-appstudio.github.io/docs.appstudio.io/Documentation/main/how-to-guides/configuring-builds/proc_custom-tags/#using-additional-tags-parameter) to ensure both `-latest` and `-slim` suffixes are provided. I use it to resolve ROX-24116 too.

The only thing I worry here is that we'll have a dangling `collector-slim` Konflux component and I wonder if it will cause any trouble. At some point, we'll just kill it.

I suggest to review this PR by commits because I tried splitting changes and provided brief messages.

## Checklist

These aren't relevant and won't be done:
- [ ] Added unit tests
- [ ] Added integration tests
- [ ] Added regression tests

## Testing Performed

- [x] Checked output images by tags.

All these exist and point to the same image:
1. `quay.io/rhacs-eng/collector:3.18.x-264-ga99ac1e8c3-fast`
2. `quay.io/rhacs-eng/collector:3.18.x-264-ga99ac1e8c3-fast-slim`
3. `quay.io/rhacs-eng/collector:3.18.x-264-ga99ac1e8c3-fast-latest`
4. `quay.io/rhacs-eng/collector:konflux-a99ac1e8c3cc2e19d2714aac0ee4dd5523be2dc4`


.tekton/collector-component-pipeline.yaml
```diff
@@ -262,6 +258,27 @@ spec:
     - name: source
       workspace: workspace
 
+  - name: apply-tags
+    params:
+    - name: IMAGE
+      value: $(tasks.build-container.results.IMAGE_URL)
+    - name: ADDITIONAL_TAGS
+      value:
+      - $(tasks.determine-image-tag.results.image-tag)-latest
```
tommartensen, line 267: When (if at all) will the `-latest` and `-slim` tags be removed? Can we link a Jira issue here?

tommartensen (approved):

tommartensen:
> The only thing I worry here is that we'll have a dangling collector-slim Konflux component and I wonder if it will cause any trouble. At some point, we'll just kill it.

It won't build anymore, but will cause trouble in the enterprise contract runs (outdated task definitions), because it is part of every snapshot. I suggest to remove it now.

.tekton/collector-component-pipeline.yaml
msugakov: So far, it's a [distant plan](https://redhat-internal.slack.com/archives/CFMQ5C2TT/p1716822459076269) for which I created the epic ROX-24402.

msugakov:
> I suggest to remove it now.

Ok, will remove after I merge this and after I back up components.

.tekton/collector-component-pipeline.yaml
tommartensen: Thanks for updating the comment.

.tekton/collector-build.yaml
```diff
@@ -9,13 +9,13 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "500"
     # TODO(ROX-21073): re-enable for all PR branches
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && (source_branch.contains("rhtap") || source_branch.contains("konflux"))
+    pipelinesascode.tekton.dev/on-cel-expression: (event == "push" && target_branch == "master") || (event == "pull_request" && (source_branch.contains("rhtap") || source_branch.contains("konflux")))
```
Molter73, line 12: Asking out of curiosity, since we are limiting the build to master and PRs with rhtap/konflux in the branch, is there some way we could wait for the build to finish from GHA? Has anyone come up with a solution for it? If so, I assume we could start integrating these builds into our regular pipeline and launching tests with these images.

tommartensen: Konflux is not intended to replace GHA as the daily driver in the current phase. It seems too unreliable (pipeline timeouts, not starting builds because of too many concurrent builds, ...)

That said, there is [this action](https://github.com/stackrox/actions/blob/main/release/wait-for-image/README.md) that waits for an image to appear in Quay.io 😉

msugakov: > If so, I assume we could start integrating these builds into our regular pipeline and launching tests with these images.

Very much yes please! That'll help make sure Konflux not only builds but its results are functional.

> is there some way we could wait for the build to finish from GHA?

No, I don't know about existing solutions and would need to research to answer that. There's a `#konflux-users` channel where you could ask.
Also, a simple thing would be to poll repos until images become available, we know image tags upfront. I think, CI in stackrox/stackrox does that.

Molter73: Sounds good, I might give it a try when I get a minute. Thanks to you both for the answers!

msugakov: JFYI, we had https://issues.redhat.com/browse/ROX-22242 but I created https://issues.redhat.com/browse/ROX-24427 for explicitness, clarity and for us not to forget. Feel free to reuse this ROX-24427 or not.

PR #1703, opened 2024-06-06
Fix driver notifications and remove RHEL 8 builder

Stringy:
## Description

Makes sure that failures are notified to the oncall channel, and removes the RHEL 8 builder that was failing to build (previously only used for kernel modules, which we no longer build)

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed
CI is enough


Molter73 (approved):
LGTM!

Molter73 (approved):

PR #1706, opened 2024-06-11
Adjust buffer size instead of cpu-per-core

erthalion:
## Description

It turns out making many CPU cores working with a single ringbuf has a potential deadlock hazard, especially visible on RT kernels. To avoid even a slightest danger, reduce ringbuf size instead. In the new scenario if we do not fit into the total ringbuf size, the size of each individual ringbuf will be adjusted and aligned for readability.

## Checklist

**Automated testing**
  - [x] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Manual testing.


Molter73 (approved):
LGTM!

PR #1711, opened 2024-06-19
Have the RHEL VMs in our integration tests use podman instead of docker

Molter73:
## Description

This is also a workaround for an issue we've been having with RHEL 8 VMs when installing docker, moving over to podman seems to fix the issue.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

VM creation and provisioning was tested manually, still running integration tests on all RHEL systems by adding the `run-multiarch-builds` label.


Molter73:
Due to an issue when mixing images built with docker and podman, we'll need to keep using docker on Z and Power VMs. In the future we might want to have a stronger separation between builder and tester VMs so we can provision them with different packages.

Stringy (approved):
LGTM! Thanks for taking a look at this. Should make things a lot easier for podman on other platforms if we need it

PR #1736, opened 2024-07-04
Builds integration tests image on demand rather than for every commit

Stringy:
## Description

Builds the integration test image on demand, rather than for every commit. 

- [x] test image not rebuilt without changes
- [x] test image built with changes
- [x] test tag is correct

### DIFF
--- a/.github/workflows/integration-test-containers.yml
+++ b/.github/workflows/integration-test-containers.yml
@@ -16,14 +16,46 @@ on:
         type: boolean
         required: true
         description: Whether the QA containers should be rebuilt
+    outputs:
+      collector-tests-tag:
+        description: The tag used for the integration test image
+        value: ${{ jobs.build-test-image.outputs.collector-tests-tag || 'master' }}
 
 jobs:
+  should-build-test-image:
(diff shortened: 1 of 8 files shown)


.github/workflows/integration-test-containers.yml
```diff
@@ -16,15 +16,47 @@ on:
...
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: dorny/paths-filter@v3
+        id: changed
+        with:
+          filters: |
+            tests-changed:
+              - integration-tests/pkg/**
```
Molter73, line 39: Think we are missing the suites directory here
```suggestion
              - integration-tests/pkg/**
              - integration-tests/suites/**
```

Molter73 (commented):
Looking good, left a small comment on a missing directory.

.github/workflows/integration-test-containers.yml
Stringy: Damn it, I knew I thought I'd missed something

Molter73 (approved):

PR #1738, opened 2024-07-10
Downgrade libbpf DEBUG logs to TRACE

Stringy:
## Description

The logs coming from libbpf are marked by Falco as DEBUG logs, but they are extremely verbose (>35000 lines) and contain little information useful to general debugging of Collector. This PR downgrades those logs to TRACE, so they are still available for debugging of CORE_BPF loading, but not always written to our usual test logs or under typical testing load.

### DIFF
--- a/collector/lib/Logging.cpp
+++ b/collector/lib/Logging.cpp
@@ -103,6 +103,10 @@ bool ParseLogLevelName(std::string name, LogLevel* level) {
 void InspectorLogCallback(std::string&& msg, sinsp_logger::severity severity) {
   auto collector_severity = (LogLevel)severity;
 
+  if (!collector::logging::CheckLogLevel(collector_severity)) {
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


collector/lib/Logging.cpp
```diff
@@ -102,7 +102,20 @@ bool ParseLogLevelName(std::string name, LogLevel* level) {
...
+    // downgrade libbpf debug logs to TRACE to avoid thousands of lines
+    // of verbose relocation logging
+    collector_severity = LogLevel::TRACE;
+  }
+
+  // remove any newlines to avoid additional empty log lines
+  // because our logging already appends a newline
+  msg.erase(std::remove(msg.begin(), msg.end(), '\n'), msg.cend());
+
+  if (collector::logging::CheckLogLevel(collector_severity)) {
```
Molter73, line 116: We might want to move this check to the top of the function, otherwise we'll be doing an `rfind` and `erase` on messages that might not be logged at all.

Molter73 (approved):
LGTM!

PR #1741, opened 2024-07-12
Improve tracking of connectionless UDP syscalls

Molter73:
## Description

The recvmmsg and sendmmsg are used by passt in order to handle UDP messages. Because it does so in a stateless manner, we need to track these in order to support UDP on OCP Virt.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [x] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Added relevant tests to CI.


Molter73:
RHEL-SAP 8.4 is failing due to a verifier error on an unimplemented ebpf instruction. After some investigation, the instruction seems to be trying to load the address to the `handle_exit` function that needs to be called, (see [this comment](https://github.com/stackrox/falcosecurity-libs/pull/79#discussion_r1765151290) for more details on why it needs to be called). I'll look to see if it's possible to make falco skip a program that fails the verifier and keeps loading the rest of them, but for now I would like to disable the test unless we have a good reason to not do so.

Molter73:
I have a potential fix for the RHEL-SAP 8.4 verifier issue on these 2 PRs:
- https://github.com/stackrox/collector/pull/1855
- https://github.com/stackrox/falcosecurity-libs/pull/84

Basically we can circumvent the verifier error by not loading the offending probe if collector is not configured to track send/recv syscalls.

I need to investigate why the `TestUdpNetworkFlow` suite is not being run, that one should still hit the verifier issue and will require it to be excluded from the tests run on RHEL-SAP 8.4.

integration-tests/suites/udp_networkflow.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+func newServerCmd(recv string, port uint16) []string {
+	return []string{
+		"--syscall", recv,
+		"--port", strconv.FormatUint(uint64(port), 10),
+	}
+}
+
+func (s *UdpNetworkFlow) runServer(cfg config.ContainerStartConfig, port uint16) containerData {
```
Stringy, line 247: Can we turn this into a generic thing to reduce duplication between runServer/runClient?
e.g.

```go
func (s *UdpNetworkFlow) runServer(cfg config.ContainerStartConfig, port uint16) containerData {
    return s.runContainer(cfg, port)
}

func (s *UdpNetworkFlow) runClient(cfg config.ContainerStartConfig) containerData {
    return s.runContainer(cfg, 0)
}

func (s *UdpNetworkFlow) runContainer(cfg config.ContainerStartConfig, port uint16) containerData {
    // ...
}
```

integration-tests/suites/udp_networkflow.go
```diff
@@ -0,0 +1,293 @@
...
+
+	for _, server := range servers {
+		serverConnection := types.NetworkInfo{
+			LocalAddress:   fmt.Sprintf(":%d", server.port),
+			RemoteAddress:  client.ip,
+			Role:           "ROLE_SERVER",
+			SocketFamily:   "SOCKET_FAMILY_UNKNOWN",
+			CloseTimestamp: types.NilTimestamp,
+		}
+		s.Sensor().ExpectConnections(s.T(), server.id, 5*time.Second, serverConnection)
```
Stringy, line 188: [nit] it might be more performant to construct the list of all server connections and then Expect all of them, and avoids a worst-case sleep for `5 * len(servers)` seconds

Stringy (approved):
Couple of questions and comments, but overall LGTM!

integration-tests/suites/udp_networkflow.go
Molter73: The problem is that the test uses 3 separate servers, we need to check them separately, unless we put together a `MultiExpectConnections` method or something that can take a `map[string]types.NetworkInfo` or similar so that we can state what NetworkInfos are expected on which containers.

Stringy: Ahh I see that I am blind, my bad. That totally makes sense!

Molter73: Oh, don't worry, I made this comment to myself like 3 times during development 🤣

PR #1742, opened 2024-07-12
Add compile time configuration for bpf_printk support

Molter73:
## Description

A small flag is added in order to enable `bpf_printk` in the CORE BPF probe.

By default there are no instances of `bpf_printk` in the probe, but when debugging these can be added to get additional information by doing `cat /sys/kernel/tracing/trace_pipe`. This has proven useful to me when testing some code in the Falco probe, so i think having an easy way to configure is is worth it, even if it won't be used by default.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Added some `bpf_printk` to the execve tracepoints, compiling with `BPF_DEBUG_MODE=ON make image` produced an image that prints to the trace pipe. Compiling with `make image` produces an image that doesn't print to the trace pipe.


Stringy (approved):
LGTM! This will be extremely useful and I'm surprised that there's no use of bpf_printk in the modern probe already - the eBPF probe has all sorts of logging built in (not all of it useful, though)

Molter73:
> LGTM! This will be extremely useful and I'm surprised that there's no use of bpf_printk in the modern probe already - the eBPF probe has all sorts of logging built in (not all of it useful, though)

Actually, I just noticed there are a few instances, but they are just a few on error paths, so they stay mostly out of the way
image

PR #1743, opened 2024-07-16
Nightly CI for Collector

Stringy:
## Description

Includes konflux workflow.

Primarily converting lots of checks from
event_name == push to event_name != pull_request
to account for schedule event.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Not entirely sure how to test this just yet.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -85,7 +85,7 @@ jobs:
 
       - name: Update Baseline
         if: |
-          ((github.event_name == 'push' && github.ref == 'master') ||
+          ((github.event_name != 'pull_request' && github.ref == 'master') ||
           contains(github.event.pull_request.labels.*.name, 'update-baseline'))
         run: |
           ./integration-tests/scripts/baseline/main.py --update integration-tests/perf-all.json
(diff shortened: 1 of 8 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -85,7 +85,7 @@ jobs:
 
       - name: Update Baseline
         if: |
-          ((github.event_name == 'push' && github.ref == 'master') ||
+          ((github.event_name != 'pull_request' && github.ref == 'master') ||
```
Molter73, line 88: Will this just update the baseline every day with the nightly? Is that the behavior we want or do we still only want to update the baseline on new commits?

Molter73 (commented):
Looks good! Just a question on the benchmark baseline job.
Also, do we want this PR with the multiarch and build builder labels? Just to make sure those still work.

.github/workflows/benchmarks.yml
Stringy: yeah it would update the baseline every night. I think this is better since we'd get a better baseline from the additional data, compared to what we have now where the benchmarks are run infrequently - @erthalion do I understand this correctly?

Molter73 (approved):
LGTM!

I still would like from @erthalion w.r.t. the baseline being updated on every nightly run though.

.github/workflows/benchmarks.yml
Stringy: I spoke to @erthalion on Slack and he has agreed that it is a good opportunity to run update the baseline

PR #1745, opened 2024-07-17
Small fix to libbpf log downgrading

Molter73:
## Description

The check for log level needs to happen after we've downgraded them to trace. In order to prevent looking for the `libbpf:` prefix on every log line issued by Falco, we first check if we are running in debug mode, limiting the impact to systems that are actively being debugged.

This bug was introduced by a change I suggested in https://github.com/stackrox/collector/pull/1738#discussion_r1679500633

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Check the integration test logs don't have the libbpf lines in them.

### DIFF
--- a/collector/lib/Logging.cpp
+++ b/collector/lib/Logging.cpp
@@ -103,14 +103,14 @@ bool ParseLogLevelName(std::string name, LogLevel* level) {
 void InspectorLogCallback(std::string&& msg, sinsp_logger::severity severity) {
   auto collector_severity = (LogLevel)severity;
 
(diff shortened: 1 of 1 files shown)


Stringy (approved):
LGTM! thanks for noticing and fixing my mistake :sweat_smile:

Molter73:
Checked the logs for the ubuntu integration tests and they don't have any mentions to libbpf, merging.

PR #1765, opened 2024-07-31
Simplify GCP VM name generation

Stringy:
## Description

Remove collection method (unused)
Remove collector- prefix (it's implied)
Truncate family name to max 16 characters
Hash and truncate the job id suffix

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)

### DIFF
--- a/ansible/roles/create-all-vms/tasks/by-family.yml
+++ b/ansible/roles/create-all-vms/tasks/by-family.yml
@@ -20,7 +20,7 @@
     apply:
       environment: "{{ ibm_env | default({}) }}"
   vars:
-    vm_name: "{{ gcp_instance_prefix }}-{{ normalized_collection_method }}-{{ item.1 }}-{{ job_id }}"
+    vm_name: "{{ gcp_instance_prefix }}-{{ item.1 | truncate(16, True, '') }}-{{ job_id }}"
     vm_project: "{{ item.0.value.project }}"
     vm_family: "{{ item.1 }}"
     vm_platform: "{{ item.0.key }}"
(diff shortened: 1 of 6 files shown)


ansible/roles/create-all-vms/tasks/by-family.yml
```diff
@@ -20,7 +21,7 @@
     apply:
       environment: "{{ ibm_env | default({}) }}"
   vars:
-    vm_name: "{{ gcp_instance_prefix }}-{{ normalized_collection_method }}-{{ item.1 }}-{{ job_id }}"
+    vm_name: "{{ gcp_instance_prefix }}-{{ item.1 | truncate(16, True, '') }}-{{ hashed_job_id }}"
```
Molter73, line 24: Should we just get rid of this `set_fact` block? 
https://github.com/stackrox/collector/blob/39f5c66645cd61417baa97067e06b9be24477499/ansible/roles/create-all-vms/tasks/main.yml#L3-L4

Molter73: Also, I'm not 100% sure we need to truncate the family, if I'm not mistaken, the longest name is used by fedora arm at 26 characters: `fedora-coreos-stable-arm64`.
Changing the prefix to `dev|ci` means we need 3 for that + the `-` + 26 for the family + another `-` + another 16 characters for the hash = 47 chars out of the 64 that is the limit. Just saying, we are probably good to leave the family as is.

Molter73 (commented):
Looks good, just a couple comments and waiting on CI to check everything is in order.

PR #1767, opened 2024-08-01
Removes eBPF logging references and eBPF switching host heuristics

Stringy:
## Description

Now we live in a post-eBPF world, the log messages in the heuristics can be confusing. They have been reworded to suggest that NO_COLLECTION can be used, since we do not have another collection method to switch to if any of these checks fail.

Also includes some minor cleanup in the unit tests and bootstrap shell script (which contained references to kernel modules... remember them?)

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)

### DIFF
--- a/collector/test/HostHeuristicsTest.cpp
+++ b/collector/test/HostHeuristicsTest.cpp
@@ -51,7 +51,7 @@ class MockHostInfoHeuristics : public HostInfo {
 class MockCollectorConfig : public CollectorConfig {
  public:
   MockCollectorConfig()
-      : CollectorConfig() {};
+      : CollectorConfig(){};
 
   void SetCollectionMethod(CollectionMethod cm) {
     if (host_config_.HasCollectionMethod()) {
--- a/collector/test/HostHeuristicsTest.cpp
+++ b/collector/test/HostHeuristicsTest.cpp
@@ -31,14 +31,9 @@ using namespace testing;
 
 namespace collector {
 
-class MockS390xHeuristics : public S390XHeuristic {
+class MockCPUHeuristic : public CPUHeuristic {
  public:
-  MockS390xHeuristics() = default;
-};
-
-class MockARM64Heuristics : public ARM64Heuristic {
- public:
-  MockARM64Heuristics() = default;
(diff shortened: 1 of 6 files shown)


collector/test/HostHeuristicsTest.cpp
```diff
@@ -65,36 +61,17 @@ class MockCollectorConfig : public CollectorConfig {
...
-  EXPECT_CALL(host, GetKernelVersion()).WillOnce(Return(version));
-
-  s390xHeuristics.Process(host, config, &hconfig);
-
-  EXPECT_EQ(hconfig.GetCollectionMethod(), CollectionMethod::CORE_BPF);
-}
-
-TEST(HostHeuristicsTest, TestARM64Heuristic) {
-  MockARM64Heuristics heuristic;
+TEST(HostHeuristicsTest, TestCPUHeuristic) {
```
Molter73, line 64: I'm a bit confused as to why is this test being added as part of this PR, it doesn't feel like it's related to the logging changes. Just for the sake of cleanliness, maybe we could consider moving it to a separate PR? Even if it ends up being just 20 lines of changes.

Stringy: This is a fun one. If I remove all the tests from this file then the compilation of the unit tests fails at link time because there's certain functions omitted from the suite, so the "fix" is to just delete the file. 

But I thought it's a bit silly to delete if we have some stuff here for testing HostHeuristics anyway, so I thought I could just add one to keep the test suite alive.

Molter73: I mean, I think I would just delete the file, but to each their own.

Stringy: but to delete it and then re-add it in a subsequent PR feels like more effort, and this heuristic doesn't have a test so I thought it'd be worth fixing now. The others still need tests I think, but since they all call CLOG(FATAL), they're a bit trickier and should probably be done in a follow up

Molter73 (approved):
LGTM!

PR #1770, opened 2024-08-02
Removes driver downloads and driver candidates

Stringy:
## Description

Final (?) removal of driver downloads and collection method switching. This removes the file downloader, the kernel object downloader, driver candidates, and thus simplifies various places where they were used. 

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


collector/lib/CollectionMethod.cpp
```diff
@@ -13,4 +15,10 @@ std::ostream& operator<<(std::ostream& os, CollectionMethod method) {
   }
 }
 
+std::string CollectionMethodName(CollectionMethod method) {
```
Molter73, line 18: I feel like we should have this the other way around, `CollectionMethodName` should check the `CollectionMethod` value and return a `const char*`, then the `operator<<` can just call into this function and any other code requiring a string can just build it, WDYT? It would save us some stream shenanigans.

For the unknown case, realistically, we should never get that, so just using `"unknown"` would be enough IMO (this wouldn't be a problem in Rust BTW).

Stringy: yeah seems very reasonable!

collector/lib/CollectionMethod.cpp
```diff
@@ -1,15 +1,26 @@
...
     case CollectionMethod::EBPF:
-      return os << "ebpf";
+      return "ebpf";
     case CollectionMethod::CORE_BPF:
-      return os << "core_bpf";
+      return "core_bpf";
     default:
-      return os << "unknown(" << static_cast<uint8_t>(method) << ")";
+      CLOG(WARNING) << "Unexpected CollectionMethod: " << static_cast<uint8_t>(method);
+      return "unknown";
```
Molter73, line 23: I would argue `ERROR` is a more appropriate severity, since we are not going to be able to continue working with an invalid collection method. I would go as far as to go into `FATAL`, but I'm an extremist.

Molter73 (approved):
This is the kind of PR all devs dream of, just remove a bunch of unneeded stuff!

Just a small comment left on log severity, but I wouldn't block merging on it. LGTM!

PR #1776, opened 2024-08-05
chore: Add `^release-` branches for on-push Konflux builds

msugakov:
## Description

This should enable Konflux builds from release branches for the case when [COLLECTOR_VERSION](https://github.com/stackrox/stackrox/blob/master/COLLECTOR_VERSION) points to a tag/commit that was on a release branch, not on `master`.

Prerequisite for ROX-24468

Available CEL functions are here <https://github.com/google/cel-spec/blob/master/doc/langdef.md#list-of-standard-definitions>.

## Checklist

**Automated testing**
No contributions.

## Testing Performed

None, only CI.

### DIFF
--- a/.tekton/collector-build.yaml
+++ b/.tekton/collector-build.yaml
@@ -9,7 +9,7 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "500"
     # TODO(ROX-21073): re-enable for all PR branches
-    pipelinesascode.tekton.dev/on-cel-expression: (event == "push" && target_branch == "master") || (event == "pull_request" && (source_branch.contains("rhtap") || source_branch.contains("appstudio") || source_branch.contains("konflux")))
+    pipelinesascode.tekton.dev/on-cel-expression: (event == "push" && (target_branch == "master" || target_branch.startsWith("release-"))) || (event == "pull_request" && (source_branch.contains("rhtap") || source_branch.contains("appstudio") || source_branch.contains("konflux")))
   creationTimestamp: null
   labels:
     appstudio.openshift.io/application: acs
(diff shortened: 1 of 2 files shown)


Molter73:
Could you re-add the `- release-*` branch [here](https://github.com/stackrox/collector/blob/fc2d6b2131057ef0145e3a1308e54fd132babc09/.github/workflows/konflux.yml#L8)? We had to remove it in #1764 because we had failures on release branches because konflux was not being triggered.

.tekton/collector-build.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "500"
     # TODO(ROX-21073): re-enable for all PR branches
-    pipelinesascode.tekton.dev/on-cel-expression: (event == "push" && target_branch == "master") || (event == "pull_request" && (source_branch.contains("rhtap") || source_branch.contains("appstudio") || source_branch.contains("konflux")))
+    pipelinesascode.tekton.dev/on-cel-expression: (event == "push" && target_branch.matches("^(master|release-.*)$")) || (event == "pull_request" && source_branch.matches("(konflux|appstudio|rhtap)"))
```
tommartensen, line 12: IIUC `matches` expects a regex to compare to the source branch here.

`(konflux|appstudio|rhtap)` would only match branches literally `konflux`, `appstudio`, `rhtap` - not `tm/konflux-something`.

msugakov: "matches" has different semantics in different languages. Here I assumed it checks that the string has a match, not that the entire string must match. What I assumed seems to be the case because the Konflux CI **runs** in this PR.

Just to be on the safe side and check that I did not mess up the condition so that it matches just any string, I tried with a "wrong" branch name https://github.com/stackrox/collector/pull/1777 and the Konflux CI does not run there.

msugakov:
@Molter73 Thanks for the pointer, added.

tommartensen (approved):

Molter73 (approved):

PR #1783, opened 2024-08-08
Add a test summary for the integration tests

Molter73:
## Description

This publishes a friendlier report on GHA, which should make it a little easier to inspect failures on our integration tests.

It also removes the unused `COLLECTOR_OFFLINE_MODE` env variable, only because I saw it while working on this.

For an example on what the report looks like when a test fails, see: https://github.com/stackrox/collector/actions/runs/10317337453

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI is enough.

### DIFF
--- a/.github/workflows/integration-tests-vm-type.yml
+++ b/.github/workflows/integration-tests-vm-type.yml
@@ -23,11 +23,6 @@ on:
           Tag used for the test container
         type: string
         required: true
-      offline-mode:
-        description: |
-          Set to true to enable collector running in offline mode
-        type: boolean
-        default: false
       run-benchmarks:
         description: |
           Whether to run the benchmarks instead of the integration tests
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-tests-vm-type.yml
```diff
@@ -23,11 +23,6 @@ on:
           Tag used for the test container
         type: string
         required: true
-      offline-mode:
-        description: |
-          Set to true to enable collector running in offline mode
-        type: boolean
-        default: false
```
ovalenti, line 30: Is there a relation between the removal of `offline-mode` and the test summary ?

Molter73: > It also removes the unused COLLECTOR_OFFLINE_MODE env variable, only because I saw it while working on this.

I basically forgot to remove it as part of #1773 and stumbled across it while making these changes. I can move it to a separate PR if you'd like.

ovalenti (approved):
This is going to make it really easier to visualize the test messages :+1:

.github/workflows/integration-tests-vm-type.yml
ovalenti: It is fine (and in the same file :smile: )

PR #1786, opened 2024-08-12
Removes SSH and GCP command builders

Stringy:
## Description

The SSH and GCP command builders are no longer used. This PR simply removes them. CI testing should be sufficient.

### DIFF
--- a/integration-tests/integration_test.go
+++ b/integration-tests/integration_test.go
@@ -21,7 +21,7 @@ func TestImageLabelJSON(t *testing.T) {
 
 // TestMissingProcScrape only works with local fake proc directory
 func TestMissingProcScrape(t *testing.T) {
-	if config.HostInfo().IsLocal() {
+	if !config.HostInfo().IsK8s() {
 		suite.Run(t, new(suites.MissingProcScrapeTestSuite))
 	}
 }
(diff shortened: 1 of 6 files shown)


ovalenti (commented):
Do we still have a use of Host.Kind ?

https://github.com/stackrox/collector/blob/giles/remove-unused-command-builders/integration-tests/pkg/config/config.go#L119

Stringy:
> Do we still have a use of Host.Kind ?

We do, because it's used to differentiate between local tests and k8s tests, but the rest of the HostInfo config values I think are unused, so I've removed them.

ovalenti (approved):
This looks all good to me !

Molter73:
> > Do we still have a use of Host.Kind ?
> 
> We do, because it's used to differentiate between local tests and k8s tests, but the rest of the HostInfo config values I think are unused, so I've removed them.

Not something to do as part of this PR, but given that tests are meant to either run on docker of k8s (at least the way they are put together now), we can probably delegate the decision of what kind of runner to use to the tests themselves, instead of making it configurable and risk someone attempting to run tests on a system they are not meant to work on.

If we ever put together a test that is able to run on docker or k8s independently we might need to put some mechanism for this back in place, but I don't think we have plans to do this any time soon.

Stringy:
> we can probably delegate the decision of what kind of runner to use to the tests themselves

Yeah good point. I'll do that in a follow up.

PR #1790, opened 2024-08-13
ROX-25788: fix integration tests handling of connections and endpoints

Stringy:
## Description

The ExpectConnections and ExpectEndpoints are incorrectly succeeding due to a logic bug in in the event filtering:

- the caller specifies a set of expected connections or endpoints
- the function attempts filter that list to get a list of entities that have not been seen
- the filter condition is backwards, instead giving a list of entities that have been seen
- At this early stage in the function, no events have been seen, but the function incorrectly believes that all events have been seen and returns success.

### DIFF
--- a/ansible/roles/run-test-target/tasks/test-collection-method.yml
+++ b/ansible/roles/run-test-target/tasks/test-collection-method.yml
@@ -33,6 +33,14 @@
     run_args: -test.timeout 120m -test.short
   when: collector_test == "ci-integration-tests"
 
+- name: Remove old logs directory
+  file:
+    path: "{{ remote_log_mount }}"
+    state: absent
+  # don't error if the directory doesn't exist yet
+  ignore_errors: true
+  become: true
+
 - name: Make logs directory
   file:
     path: "{{ remote_log_mount }}"
@@ -72,6 +80,7 @@
     become: "{{ runtime_as_root }}"
     shell: |
       {{ runtime_command }} run -it --rm \
+        --pull always \
         -v {{ remote_log_mount }}:{{ container_logs_root }} \
         -v ~/.docker:/root/.docker \
         -v /tmp/:/tmp \
--- a/integration-tests/pkg/common/utils.go
+++ b/integration-tests/pkg/common/utils.go
@@ -2,10 +2,13 @@ package common
 
 import (
 	"bytes"
+	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/stackrox/collector/integration-tests/pkg/config"
 	"golang.org/x/sys/unix"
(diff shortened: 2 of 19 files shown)


ansible/roles/run-test-target/tasks/test-collection-method.yml
Stringy, line 1: Changes here are just to make it a little easier to test custom test images on ansible-created VMs. They are unrelated to the test bug, but I'm including them for simplicity

integration-tests/pkg/common/utils.go
```diff
@@ -64,3 +67,9 @@ func PrepareLog(testName string, logName string) (*os.File, error) {
 	logPath := filepath.Join(logDirectory, strings.ReplaceAll(testName, "/", "_")+"-"+logName)
 	return os.Create(logPath)
 }
+
+func Sleep(duration time.Duration) {
```
Stringy, line 71: Another change that's not strictly related to the bug, but is helpful to understand the sleep behaviour of the tests (may help us to narrow down timing flakes in the future)

integration-tests/pkg/types/endpoint.go
```diff
@@ -62,5 +62,9 @@ func endpointComparison(endpoint1 EndpointInfo, endpoint2 EndpointInfo) bool {
 		return process1.ProcessArgs < process2.ProcessArgs
 	}
 
+	if endpoint1.IsActive() == endpoint2.IsActive() {
```
erthalion, line 65: This stands a bit out of the pattern used in this function -- all of them are compared for not equality. Maybe to reduce the surprise level, compare for inequality here as well?

integration-tests/pkg/types/endpoint.go
```diff
@@ -15,10 +15,10 @@ func (n *EndpointInfo) IsActive() bool {
 }
 
 func SortEndpoints(endpoints []EndpointInfo) {
-	sort.Slice(endpoints, func(i, j int) bool { return endpointComparison(endpoints[i], endpoints[j]) })
+	sort.Slice(endpoints, func(i, j int) bool { return CompareEndpoints(endpoints[i], endpoints[j]) })
 }
 
-func endpointComparison(endpoint1 EndpointInfo, endpoint2 EndpointInfo) bool {
+func CompareEndpoints(endpoint1 EndpointInfo, endpoint2 EndpointInfo) bool {
```
erthalion, line 21: Unrelated to the change set, the function handling of nil looks strange: a comparison will always succeed if a second address or an originator process is nil, which might be quite a surprise to the caller.

integration-tests/pkg/mock_sensor/server.go
```diff
@@ -208,8 +208,12 @@ func (m *MockSensor) HasEndpoint(containerID string, endpoint types.EndpointInfo
 	defer m.networkMutex.Unlock()
 
 	if endpoints, ok := m.endpoints[containerID]; ok {
-		_, exists := endpoints[endpoint]
-		return exists
+		// _, exists := endpoints[endpoint]
```
erthalion, line 211: Leftover comment?

integration-tests/pkg/common/utils.go
erthalion: Like this one, it should help reducing need in any kind of waiting in tests.

erthalion (approved):
Looks good, thanks. Few minor comments below.

integration-tests/pkg/types/endpoint.go
Stringy: Agreed, that's a bit odd. I'll rework this function a bit

Stringy: In fact I see that my use of this function is incorrect too. I'll fix it.

erthalion (approved):

PR #1795, opened 2024-08-14
build: Adjust timeout waiting for Konflux images

msugakov:
## Description

I noticed that GHA builds were failing on me https://github.com/stackrox/collector/actions/runs/10375095581/job/28724093219

This should fix it.

## Checklist

**Automated testing**

No change to automated tests.

## Testing Performed

GHA should pass.

### DIFF
--- a/.github/workflows/konflux-tests.yml
+++ b/.github/workflows/konflux-tests.yml
@@ -47,7 +47,7 @@ jobs:
         with:
           token: ${{ secrets.QUAY_RHACS_ENG_BEARER_TOKEN }}
           image: rhacs-eng/collector:${{ needs.init.outputs.collector-tag }}
-          limit: 5400 # 1h30m
+          limit: 9000 # 2h30m
 
   integration-tests-containers:
     uses: ./.github/workflows/integration-test-containers.yml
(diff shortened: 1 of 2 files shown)


Molter73 (approved):

msugakov:
When Konflux build fails, `wait-for-images` does not know that and keeps waiting for images. That's something to be fixed, but too much for this PR. At least here I saw it waiting. Just want to see GHA konflux workflows succeeding.

Molter73:
> When Konflux build fails, `wait-for-images` does not know that and keeps waiting for images. That's something to be fixed, but too much for this PR. At least here I saw it waiting. Just want to see GHA konflux workflows succeeding.

Yeah, the `wait-for-images` step is not smart enough, it only checks if the image is available in the registry, not if the konflux workflow is running. I believe the main repo has the same issue since it uses this action, but unless we get some clear way to get the konflux cluster that is running associated to a given GHA workflow so that we can query some sort of API/hook in there, I don't know if we can do any better.

In any case, as you say, this is not to be addressed in this PR.

kylape (approved):

PR #1803, opened 2024-08-19
Debug

erthalion:
## Description

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


Molter73 (commented):
Looks good, if we were to merge it I would argue against store copies of the collector configuration in other objects, but for the purpose of troubleshooting it is good enough as is.

erthalion:
> I would argue against store copies of the collector configuration in other objects

AFAICT it's passed by reference in this PR, why would there be any copying involved?

Molter73:
> > I would argue against store copies of the collector configuration in other objects
> 
> AFAICT it's passed by reference in this PR, why would there be any copying involved?

It's passed by reference to the constructors, but it is stored as a concrete value in these 2 places:
https://github.com/stackrox/collector/blob/61089fe71fb4edd676f7406e65b77fcd02a934f5/collector/lib/ProcessSignalFormatter.h#L57
https://github.com/stackrox/collector/blob/61089fe71fb4edd676f7406e65b77fcd02a934f5/collector/lib/ProcessSignalHandler.h#L40

So you'll get copies in the end, if you delete the implicit copy constructor for `CollectorConfig` you'll see compilation fails on those two places and in CollectorService.h

PR #1805, opened 2024-08-20
Add username to dev VMs to make it easier to differentiate them

Stringy:
## Description

To make GCP VMs easier to differentiate, this PR prefixes the (local) username to the VM name upon creation. 

Tested locally.

Regarding VM name length limits, this won't affect CI VMs at all, and will break locally a fair while before it breaks for VMs (though with the recent changes to the naming convention, we have a significant buffer zone before limits become a problem again)


Molter73 (approved):
LGTM!

Thanks for adding this, it'll make it easier to distinguish which VM is the one I created 😅

PR #1807, opened 2024-08-21
Update Falco bumping instructions

Molter73:
## Description

With the removal of the MODULE_VERSION file in #1773 our handling of the falco libs fork is somewhat outdated, since now there is no module version to use as a guideline. Instead, I'm proposing that we start using the upstream version we are based on as part of our branching strategy.

With this new approach, branches with our changes will be of the form M.m.p-stackrox, with M.m.p being the version numbers for the upstream libs tag the branch is based on.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

This is a documentation only PR, no testing is necessary.

### DIFF
--- a/docs/falco-update.md
+++ b/docs/falco-update.md
@@ -14,24 +14,39 @@ order to retain versions of released code and avoid maintenance hell.
 The process, at a high level is:
 
 - Update the `upstream-main` branch to the tip of master
-- Create a new module version branch from the previous version
-- Rebase the new module version branch to a newer tag on `upstream-main`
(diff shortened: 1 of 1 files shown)


ovalenti (approved):
I think that this new scheme makes sense. Although @Stringy should have the final word here.

Stringy (approved):
LGTM! everything will be a lot clearer now :ok_hand:

PR #1813, opened 2024-08-28
Fix grpc-server image tag in docs

SimonBaeumer:
## Description

Fix grpc-server image tag in docs

### DIFF
--- a/docs/how-to-start.md
+++ b/docs/how-to-start.md
@@ -68,7 +68,7 @@ services:
     depends_on:
       - grpc-server-debug
   grpc-server-debug:
-    image: quay.io/rhacs-eng/grpc-server:latest
+    image: quay.io/rhacs-eng/grpc-server:<image-tag> # e.g. 4.3.8
     container_name: grpc-server
     network_mode: host
     user: 1000:1000


docs/how-to-start.md
```diff
@@ -68,7 +68,7 @@ services:
     depends_on:
       - grpc-server-debug
   grpc-server-debug:
-    image: quay.io/rhacs-eng/grpc-server:latest
+    image: quay.io/rhacs-eng/grpc-server:<image-tag> # e.g. 4.3.8
```
Molter73, line 71: These docs may be more outdated than I expected, we haven't built a mock grpc server in the last 3 months and I'm not fully sure the last built version is still compatible with the latest version for collector. Further, I believe we've removed all the bits for building the image from the main repo.

@Stringy, is there any way we could repurpose some of the code for our integration tests so that we can spin up a mock server and start building the image in our CI? Otherwise I propose we remove this section of the docs and replace it with a pointer to deploying stackrox K8S/Openshift/minikube/kind/whatever in the main repo, (or get our act together and finally make it so we can just `make -j$(nproc) collector && ./collector` in a regular terminal).

Molter73 (commented):
Thanks for the PR @SimonBaeumer! Left a comment because these docs are painfully outdated, we'll look into it as a team.

docs/how-to-start.md
SimonBaeumer: @Molter73 So far I can run Collector with this approach, I am not sure that it is 100% working though. I know collector not good enough.

Is there documentation for the recommended path to work with collector?

Molter73: It's probably mostly working because we don't update our protobufs that often, so if this approach works for you it's fine, I just want to make sure we provide a valuable way to run collector without needing to spin up a full k8s cluster once we change the doc 🙂 .

In all honestly, the easiest way if you are used to having a cluster running is to just deploy collector there and set its log level to debug by changing the `COLLECTOR_CONFIG` env variable to look like this:
```
{"tlsConfig":{"caCertPath":"/var/run/secrets/stackrox.io/certs/ca.pem","clientCertPath":"/var/run/secrets/stackrox.io/certs/cert.pem","clientKeyPath":"/var/run/secrets/stackrox.io/certs/key.pem"},"logLevel":"Debug"}
```
Notice the `logLevel` field at the end and the lack of any spaces (I know, JSON in an env var is insane, I'm actually trying to provide alternatives to it right now).

If you need to run collector under GDB there are instructions to do so on a cluster in this same doc file.

PR #1829, opened 2024-09-05
ROX-25723: add ecosystem preflight check

tommartensen:
## Description

Adds ecosystem cert preflight check to Konflux pipeline.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Passing Konflux CI is sufficient.

### DIFF
--- a/.tekton/collector-component-pipeline.yaml
+++ b/.tekton/collector-component-pipeline.yaml
@@ -505,6 +505,24 @@ spec:
       operator: in
       values: [ "false" ]
 
+  - name: ecosystem-cert-preflight-checks
+    params:
+    - name: image-url
+      value: $(tasks.build-image-manifest.results.IMAGE_URL)
+    taskRef:
+      params:
+      - name: name
+        value: ecosystem-cert-preflight-checks
+      - name: bundle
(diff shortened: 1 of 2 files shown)


.tekton/collector-component-pipeline.yaml
```diff
@@ -527,6 +527,24 @@ spec:
       operator: in
       values: [ "false" ]
 
+  - name: ecosystem-cert-preflight-checks
```
msugakov, line 530: I think, it was placed before `sast-snyk-check` in other pipelines. Let's have it here also that way for consistency (e.g. I sometimes diff pipelines in different repos and it could be inconvenient to see lines moved).

msugakov (approved):
Green light after you move the task higher in the file.

.tekton/collector-component-pipeline.yaml
tommartensen: fixed in https://github.com/stackrox/collector/pull/1829/commits/e5fdfadcb232881f0f4830f0b2da13f5dd45aeda

Molter73 (approved):

PR #1841, opened 2024-09-11
Run konflux builds and tests on all PRs

Molter73:
## Description

This should help us prevent breaking this build on master.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Check all konflux jobs run in the PR.


.github/workflows/konflux.yml
```diff
@@ -20,10 +20,85 @@ concurrency:
   cancel-in-progress: true
```
msugakov, line 21: [two things, not about this line]

One:
```yaml
name: Run konflux workflows
```
may suggest that GitHub drives/triggers Konflux which is not the case. I'd propose to change the name along the lines of "Run post-Konflux workflows" or "Test Konflux builds" or something.

Two:
This has cron schedule, I wonder what's the use?

msugakov (approved):
The comment/s is/are not blocking.

.github/workflows/konflux.yml
Molter73: > One:
> 
> ```yaml
> name: Run konflux workflows
> ```

I agree, I'll rename the workflow.

> Two:
> This has cron schedule, I wonder what's the use?

I believe this was added by @Stringy in order to run the konflux tests on nightlies, but I don't think it has much point if we don't trigger the konflux build as well. I'll ask him about it.

msugakov (approved):

Stringy (approved):
LGTM!

PR #1848, opened 2024-09-17
refactor if-statements for better readability 

SimonBaeumer:
## Description

Added brackets to ifs.


erthalion (approved):

Molter73 (commented):
Not something for this PR, but we could enable the `InsertBraces` style check in our `.clang-format` file and apply it everywhere that matters

https://clang.llvm.org/docs/ClangFormatStyleOptions.html#insertbraces

PR #1849, opened 2024-09-17
Initial expansion of if-brackets in all `collector/lib` files

SimonBaeumer:
## Description

This PR runs the expand brackets option of the `clang-format` tool.

I've extracted some other parts of the PR, mainly because the clang-format requires an update. To avoid merge conflicts this PR only provides the initial formatting.

Subsequent PRs:

 - Enable flag in clang-format config: https://github.com/stackrox/collector/pull/1851
 - Add init-githooks target: https://github.com/stackrox/collector/pull/1850

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,9 +121,6 @@ shellcheck-all-dockerized:
 	docker build -t shellcheck-all $(CURDIR)/utilities/shellcheck-all
 	docker run --rm -v "$(CURDIR):/scripts" shellcheck-all:latest
 
-.PHONY: init-githook
-init-githook:
-	git config core.hooksPath ./githooks/
 
 # This defines a macro that can be used to add pre-commit targets
 # to check staged files (check-<linter name>) or all files (check-<linter name>-all)
--- a/Makefile
+++ b/Makefile
@@ -121,6 +121,9 @@ shellcheck-all-dockerized:
 	docker build -t shellcheck-all $(CURDIR)/utilities/shellcheck-all
 	docker run --rm -v "$(CURDIR):/scripts" shellcheck-all:latest
 
+.PHONY: init-githook
+init-githook:
+	git config core.hooksPath ./githooks/
 
 # This defines a macro that can be used to add pre-commit targets
 # to check staged files (check-<linter name>) or all files (check-<linter name>-all)
--- a/collector/lib/CollectorStatsExporter.cpp
+++ b/collector/lib/CollectorStatsExporter.cpp
@@ -248,6 +248,7 @@ void CollectorStatsExporter::run() {
         if (counters.parse_micros_total) {
           counters.parse_micros_total->Set(parse_micros_total);
         }
+
         if (counters.process_micros_total) {
           counters.process_micros_total->Set(process_micros_total);
         }
--- a/collector/lib/CollectorStatsExporter.cpp
+++ b/collector/lib/CollectorStatsExporter.cpp
@@ -241,12 +241,20 @@ void CollectorStatsExporter::run() {
 
         nUserspace += userspace;
 
-        if (counters.userspace) counters.userspace->Set(userspace);
+        if (counters.userspace) {
+          counters.userspace->Set(userspace);
+        }
 
-        if (counters.parse_micros_total) counters.parse_micros_total->Set(parse_micros_total);
-        if (counters.process_micros_total) counters.process_micros_total->Set(process_micros_total);
+        if (counters.parse_micros_total) {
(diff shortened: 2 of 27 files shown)


Makefile
```diff
@@ -121,6 +121,9 @@ shellcheck-all-dockerized:
 	docker build -t shellcheck-all $(CURDIR)/utilities/shellcheck-all
 	docker run --rm -v "$(CURDIR):/scripts" shellcheck-all:latest
 
+.PHONY: init-githook
+init-githook:
+	git config core.hooksPath ./githooks/
```
Molter73, line 126: Scope just keeps getting bigger 🤣

collector/lib/CollectorStatsExporter.cpp
```diff
@@ -241,12 +241,20 @@ void CollectorStatsExporter::run() {
...
-        if (counters.process_micros_total) counters.process_micros_total->Set(process_micros_total);
-
-        if (counters.parse_micros_avg) counters.parse_micros_avg->Set(userspace ? parse_micros_total / userspace : 0);
+        if (counters.userspace) {
+          counters.userspace->Set(userspace);
+        }
+
+        if (counters.parse_micros_total) {
+          counters.parse_micros_total->Set(parse_micros_total);
+        }
```
Molter73, line 250: Can we add one empty space here? It looks a bit off with it being 1 block with 1 if, then another block with 2 ifs, then a third one with 1 if.

Alternatively, we could get all ifs together with no additional space between them.

collector/lib/optionparser.h
Molter73, line 1: This file should not be formatted, did you run clang-format directly? If so, you should instead run `make -C collector/ format` which ignores this file.

The reason for it not being formatted is that it is a header only third party thing and we just keep it as is.

Molter73 (commented):
Looks good! Thanks for putting this together!

Main concern is that the optionparser.h file shouldn't be formatted, other than that LGTM!

Molter73:
Looks like CI is using `clang-format` v10, I'm looking into why it's doing that and how we can update it to at least 15.

Molter73:
> Looks like CI is using an old version of `clang-format`, I'm looking into why it's doing that and how we can update it to at least 15.

Ok, seems there are 2 options here:
- The `ubuntu-latest` runner which is used in CI has `clang-format` v13, v14 and v15 already installed in it, so you could go down the rabbit hole that is telling ubuntu to use the right version of clang-format (15 in this case).
  - https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#language-and-runtime

-  Easier alternative, though a bit overkill, you can just update the lint workflow to use `ubuntu-24.04` which has `clang-format` v16, v17 and v18, any of which will work with the new option.
  - https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#language-and-runtime

collector/lib/CollectorStatsExporter.cpp
SimonBaeumer: done

collector/lib/optionparser.h
SimonBaeumer: Good catch. Would be possible to move `optionsparser.h` out of `collector/lib`?

SimonBaeumer:
> > Looks like CI is using an old version of `clang-format`, I'm looking into why it's doing that and how we can update it to at least 15.
> 
> Ok, seems there are 2 options here:
> 
> * The `ubuntu-latest` runner which is used in CI has `clang-format` v13, v14 and v15 already installed in it, so you could go down the rabbit hole that is telling ubuntu to use the right version of clang-format (15 in this case).
>   
>   * https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#language-and-runtime
> * Easier alternative, though a bit overkill, you can just update the lint workflow to use `ubuntu-24.04` which has `clang-format` v16, v17 and v18, any of which will work with the new option.
> * https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#language-and-runtime

Tanks for looking it up! I've created a new PR to update the clang-format: https://github.com/stackrox/collector/pull/1851

This PR now focuses only on inserting the braces to avoid tedious merge conflicts.

Molter73 (approved):
LGTM!

collector/lib/optionparser.h
Molter73: Yeah, we probably want to set it up as a module or something, for now I wouldn't mess about with it too much though.

PR #1850, opened 2024-09-18
Add `init-githooks` Makefile target

SimonBaeumer:
## Description

Add `init-githooks` Makefile target similar to stackrox.


Molter73 (approved):
LGTM!

PR #1851, opened 2024-09-18
Enable InsertBraces in clang-format

SimonBaeumer:
## Description

Enable expand brackets clang-format for better readability.

https://github.com/stackrox/collector/pull/1851

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -7,6 +7,8 @@ PointerAlignment: Left
 
 IndentPPDirectives: AfterHash
 
+InsertBraces: true
+
 IncludeBlocks: Regroup
 IncludeCategories:
 - Regex: '^((\"|<)(api|internalapi|storage)/)'
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,7 +4,7 @@ on: [pull_request]
 
 jobs:
   lint:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
         with:
@@ -18,4 +18,9 @@ jobs:
           # Install actionlint to ~/bin/, otherwise pre-commit won't be able to find the executable.
           GOBIN="${HOME}/bin/" go install github.com/rhysd/actionlint/cmd/actionlint@latest
 
-      - uses: pre-commit/action@v3.0.0
+      - name: Run pre-commit
+        run: |
+          python3 -m venv .venv
+          source .venv/bin/activate
+          python3 -m pip install pre-commit
+          pre-commit run --show-diff-on-failure --color=always --all-files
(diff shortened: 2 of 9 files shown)


Molter73:
So the lint workflow is now broken because this line:
https://github.com/stackrox/collector/blob/3f10453b255cba73f85a5b678eb9ec663c87d062/.github/workflows/lint.yml#L21

Is trying to install packages using pip on a system that has python managed by the OS (this is actually a good thing, a lot of systems have been broken by pip installing packages over the actual system ones).

Additionally, the `pre-commit/action` workflow is now in maintanance mode and they suggest using `pre-commit-ci` instead, but that looks like a tool that fixes commits and is not really what we want, we want to check PRs, we don't want to have additional commits pushed on us.

I suggest we change the pre-commit/action line for the following block instead:
```yaml
      - name: Run pre-commit
        run: |
          python3 -m venv .venv
          source .venv/bin/activate
          python3 -m pip install pre-commit
          pre-commit run --all-files
```

Not as convenient as a prebuilt action, but it's only 4 lines of bash, so 🤷🏻‍♂️.

SimonBaeumer:
> So the lint workflow is now broken because this line:
> 
> https://github.com/stackrox/collector/blob/3f10453b255cba73f85a5b678eb9ec663c87d062/.github/workflows/lint.yml#L21
> 
> Is trying to install packages using pip on a system that has python managed by the OS (this is actually a good thing, a lot of systems have been broken by pip installing packages over the actual system ones).
> 
> Additionally, the `pre-commit/action` workflow is now in maintanance mode and they suggest using `pre-commit-ci` instead, but that looks like a tool that fixes commits and is not really what we want, we want to check PRs, we don't want to have additional commits pushed on us.
> 
> I suggest we change the pre-commit/action line for the following block instead:
> 
> ```yaml
>       - name: Run pre-commit
>         run: |
>           python3 -m venv .venv
>           source .venv/bin/activate
>           python3 -m pip install pre-commit
>           pre-commit run --all-files
> ```
> 
> Not as convenient as a prebuilt action, but it's only 4 lines of bash, so 🤷🏻‍♂️.

Let's try it.

Molter73:
> Let's try it.

Well it sorta worked, looks like the clang-format step failed, maybe we need to rebase this PR so the changes you did to the braces are picked up? Also, could you add `--show-diff-on-failure --color=always` to the pre-commit command in the lint workflow? That should help see what the differences actually are.

There's also a lint error I was not expecting here: 
https://github.com/stackrox/collector/blob/3f10453b255cba73f85a5b678eb9ec663c87d062/integration-tests/container/perf/scripts/init.sh#L143-L146

That block seems unreachable because the previous if block does `exit  0` on both of its branches. @Stringy, can we remove the block or should we move it above the previous one?

Molter73:
Discussed offline and decided I'll be taking over the PR, @SimonBaeumer sorry your small change got so out of hand 😅

Stringy (approved):
LGTM!

Molter73:
Looks like konflux is gonna time out, but we don't really care about it, so I'm merging as is.

PR #1855, opened 2024-09-19
Only load interesting syscalls

Molter73:
## Description

This is a small-ish change that makes it so tracepoints associated to syscalls that are not interesting will not be loaded. With this change, we can decide to exclude a tracepoint at runtime, for instance to work around a verifier issue.

In particular, the new implementations for sendmmsg and recvmmsg tracepoints have verifier issues, because these syscalls are opt-in, people that don't enable these syscalls should also not hit the verifier issue.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Check the RHEL-SAP 8.4 VM no longer hits the verifier issue (TestUdpNetworkflow will still fail).


.github/workflows/integration-test-containers.yml
```diff
@@ -45,6 +45,7 @@ jobs:
               - integration-tests/Dockerfile
               - .github/workflows/integration-test-containers.yml
 
+  # Trigger rebuild
```
Stringy, line 48: Shouldn't it be rebuilding anyway, or is the list of filters above not working properly?

Stringy (approved):
LGTM!

.github/workflows/integration-test-containers.yml
Molter73: It is not triggering because this PR doesn't have any changes to the files in the filter. I was gonna suggest we should add the `base` argument, but the docs say this is ignored when the trigger is `pull_request`, so not entirely sure what we could do about it 🤷🏻‍♂️

PR #1857, opened 2024-09-23
ROX-26025: Sanitize UTF-8 strings in process signals

ovalenti:
## Description

Containers can produce non-UTF8 process names or paths. We want to prevent this from causing issues in protobuf, which doesn't like it at all.

As a quick workaround, we reuse `google::protobuf::internal::UTF8CoerceToStructurallyValid` to replace any invalid sequence with '?'.

In order to minimize string copies in the "no modifications required" case, we introduce a Holder<> class to either point to the original unmodified string, or to an altered copy of it.

## Checklist

### DIFF
--- a/collector/lib/Utility.h
+++ b/collector/lib/Utility.h
@@ -108,6 +108,12 @@ ScopedLock<Mutex> Lock(Mutex& mutex) {
 #define ssizeof(x) static_cast<ssize_t>(sizeof(x))
 
 std::optional<std::string_view> ExtractContainerIDFromCgroup(std::string_view cgroup);
-}  // namespace collector
 
+// Replace any occurrence of an invalid UTF-8 sequence with the '?' character
+// Returns :
+//  - a new string with invalid characters replaced.
+//  - nullopt if there is no invalid character (the input string is valid).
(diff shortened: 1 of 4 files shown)


collector/lib/Utility.h
```diff
@@ -108,6 +108,36 @@ ScopedLock<Mutex> Lock(Mutex& mutex) {
 #define ssizeof(x) static_cast<ssize_t>(sizeof(x))
 
 std::optional<std::string_view> ExtractContainerIDFromCgroup(std::string_view cgroup);
+
+// A Holder refers to an object that it can own, or not.
```
Molter73, line 112: This sounds like a lifetime nightmare. Can't we do something similar to what protobuf does and simply use pointers? Or maybe make `SanitizeUTF8` return a `std::optional<std::string>`? That way if the return value is empty it would mean the original string is valid and, if it is not empty, it will hold the sanitized string.

ovalenti: I see your point, I wasn't convinced myself.

The problem is with the immutability of std::string, which makes you copy and copy again. I will figure out a more maintainable way.

Molter73 (approved):
LGTM!

PR #1860, opened 2024-09-23
Replace processes-listening-on-ports.c with plop.py

Molter73:
## Description

Recently stumbled onto a flake of the TestProcessListeningOnPort test and decided to look into it a bit further. I noticed the way works is by using a file that is read by a C binary, once the file is read, the binary deletes it. I believe the flake is due to the file being created via an exec into the container, then checking if the file exists on a mounted directory, which could very well lead to a race condition with the exec being done, but the file not showing up on the local file system, leading to believe it has already been processed and it being overwritten.

In order to have a more reliable way of opening and closing ports, the C binary has been replaced with a flask app that can open ports on its /open endpoint and closing them with its /close endpoint. These operations should be a little more reliable, since the http.Get operations will only return once the socket has been properly opened or closed and we will get confirmation via the HTTP response code.

Some additional changes are needed for tests to pass with the new version of the container, but most are minor.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Run multiarch tests.


integration-tests/container/processes-listening-on-ports/plop.py
```diff
@@ -0,0 +1,28 @@
...
+app = Flask(__name__)
+
+
+opened_sockets = {}
+
+
+@app.route('/open/<int:port>')
+def open_port(port: int):
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.bind((socket.gethostname(), port))
```
Stringy, line 16: What happens when this port is already taken? does the error propagate through the HTTP response?

integration-tests/suites/listening_ports.go
```diff
@@ -130,3 +142,15 @@ func (s *ProcessListeningOnPortTestSuite) TestProcessListeningOnPort() {
 func getProcessListeningOnPortsImage() string {
 	return config.Images().QaImageByKey("qa-plop")
 }
+
+func (s *ProcessListeningOnPortTestSuite) openPort(port uint16) {
+	res, err := http.Get(fmt.Sprintf("http://%s:%s/open/%d", s.serverIp, s.serverPort, port))
```
Stringy, line 147: [nit] can we store the url in the suite?

e.g.

```
res, err := http.Get(fmt.Sprintf("%s/open/%d", s.serverURL, port))
```

now that I've written it out, I see it maybe doesn't buy us much. Feel free to take it or leave it :laughing:

integration-tests/suites/symlink_process.go
```diff
@@ -75,5 +68,5 @@ func (s *SymbolicLinkProcessTestSuite) TestSymbolicLinkProcess() {
 	assert.Equal(s.T(), endpoints[0].Originator.ProcessName, lnProcess.Name)
 	assert.Equal(s.T(), endpoints[0].Originator.ProcessExecFilePath, lnProcess.ExePath)
 	assert.Equal(s.T(), endpoints[0].Originator.ProcessArgs, lnProcess.Args)
-	assert.Equal(s.T(), 9092, endpoints[0].Address.Port)
+	assert.Equal(s.T(), 5000, endpoints[0].Address.Port)
```
Stringy, line 71: [nit] could do with a small comment describing why it's this port number, because it's not clear in the test that it's the port the image exposes

integration-tests/container/processes-listening-on-ports/plop.py
Molter73: It does:
image

integration-tests/suites/listening_ports.go
Molter73: I think it makes sense, I'll make the change

integration-tests/suites/symlink_process.go
Molter73: Makes sense, I'll add it.

integration-tests/container/processes-listening-on-ports/plop.py
Stringy: Hmm but the error we'll see will just be the assertion on the response code being 500. It might be worth adding our own error response so we can propagate the precise exception

Molter73: Alright, let me try to catch errors and give meaningful messages back.

Molter73: This should be a bit better
image

PR #1864, opened 2024-09-26
Remove unneeded packages from final collector image

Molter73:
## Description

Reasoning for removing each package:
- kmod has not been needed since we stopped supporting kernel modules
- libyaml and libarchive have vulnerabilities, are unused by the the collector binary and show no other package depend on them when querying with rpm.
- Replaced `xargs rpm -e` with the method used in konflux.Dockerfile, so findutils is no longer needed either.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Run tests in the main repo, ensuring collector still works properly in real operation environments.

### DIFF
--- a/collector/container/rhel/install.sh
+++ b/collector/container/rhel/install.sh
@@ -3,8 +3,11 @@ set -eo pipefail
 
 # UBI 9 requires confirmation with -y flag.
 microdnf upgrade -y --nobest
-microdnf install -y kmod findutils elfutils-libelf
+microdnf install -y elfutils-libelf
 
 microdnf clean all
-rpm --query --all 'curl' '*rpm*' '*dnf*' '*libsolv*' '*hawkey*' 'yum*' 'findutils' | xargs -t rpm -e --nodeps
+# shellcheck disable=SC2046
+rpm --verbose -e --nodeps $(
+    rpm -qa 'curl' '*rpm*' '*dnf*' '*libsolv*' '*hawkey*' 'yum*' 'libyaml*' 'libarchive*'
+)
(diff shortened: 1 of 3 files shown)


ovalenti (approved):
LGTM !

collector/container/rhel/install.sh
```diff
@@ -3,8 +3,11 @@ set -eo pipefail
 
 # UBI 9 requires confirmation with -y flag.
 microdnf upgrade -y --nobest
-microdnf install -y kmod findutils elfutils-libelf
+microdnf install -y elfutils-libelf
 
 microdnf clean all
-rpm --query --all 'curl' '*rpm*' '*dnf*' '*libsolv*' '*hawkey*' 'yum*' 'findutils' | xargs -t rpm -e --nodeps
+# shellcheck disable=SC2046
+rpm --verbose -e --nodeps $(
```
janisz, line 10: How about removing `--nodeps`

Molter73: I'm not entirely sure why `--nodeps` is used here, it was like this when I found it. Looking into rpms help
```
--nodeps                       don't verify package dependencies
```

So I'm guessing that flag is there to leave dependencies of the affected packages alone. Could you elaborate on why you think it would be worth it to remove it? Is it so dependencies that are no longer used get removed?

Molter73: On second thought, I've managed to break collector by removing a kerberos package that was listed as unused but was still needed, so that might be the entire reason the flag is there.

PR #1865, opened 2024-09-26
 Use python:3.12-slim for plop.py

Molter73:
## Description

 The non-slim variant is about 1GB in size, slim is closer to 100MB.
 This is not a production grade image, but still, no reason to be
 bloating images for the sake of it.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI should be enough.


Stringy (approved):
LGTM!

PR #1866, opened 2024-09-26
ROX-26344: adds runtime configuration to collector

Stringy:
## Description

Includes support for enabling/disabling external IPs based on runtime config, deferring to existing feature flag if none is provided

In subsequent PRs, the runtime configuration will be populated. 

## Checklist
~- [x] Updated documentation accordingly~

**Automated testing**
  - [x] Added unit tests

## Testing Performed

Unit tests added to exercise the intended behaviour. It is not possible to integration-test this until Sensor->Collector communication has been built.

### DIFF
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -47,7 +47,7 @@ BoolEnvVar set_import_users("ROX_COLLECTOR_SET_IMPORT_USERS", false);
 
 BoolEnvVar collect_connection_status("ROX_COLLECT_CONNECTION_STATUS", true);
 
-BoolEnvVar enable_external_ips("ROX_ENABLE_EXTERNAL_IPS", false);
+BoolEnvVar enable_external_ips("ROX_EXTERNAL_IPS", false);
 
 BoolEnvVar enable_connection_stats("ROX_COLLECTOR_ENABLE_CONNECTION_STATS", true);
 
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -47,7 +47,7 @@ BoolEnvVar set_import_users("ROX_COLLECTOR_SET_IMPORT_USERS", false);
 
 BoolEnvVar collect_connection_status("ROX_COLLECT_CONNECTION_STATUS", true);
 
-BoolEnvVar enable_external_ips("ROX_EXTERNAL_IPS", false);
+BoolEnvVar enable_external_ips("ROX_ENABLE_EXTERNAL_IPS", false);
 
 BoolEnvVar enable_connection_stats("ROX_COLLECTOR_ENABLE_CONNECTION_STATS", true);
 
--- a/collector/lib/CollectorConfig.h
+++ b/collector/lib/CollectorConfig.h
@@ -84,8 +84,8 @@ class CollectorConfig {
   // otherwise, we rely on the feature flag (env var)
   bool EnableExternalIPs() const {
     if (runtime_config_.has_value()) {
-      auto cfg = runtime_config_.value();
-      auto network_cfg = cfg.network_connection_config();
+      const auto& cfg = runtime_config_.value();
+      const auto& network_cfg = cfg.network_connection_config();
       return network_cfg.enable_external_ips();
     }
     return enable_external_ips_;
@@ -112,7 +112,7 @@ class CollectorConfig {
     runtime_config_ = std::move(runtime_config);
   }
 
-  std::optional<sensor::CollectorConfig> GetRuntimeConfig() const {
+  const std::optional<sensor::CollectorConfig>& GetRuntimeConfig() const {
     return runtime_config_;
   }
 
--- a/collector/lib/CollectorConfig.h
+++ b/collector/lib/CollectorConfig.h
@@ -1,12 +1,15 @@
 #ifndef _COLLECTOR_CONFIG_H_
 #define _COLLECTOR_CONFIG_H_
 
+#include <optional>
 #include <ostream>
 #include <vector>
 
 #include <json/json.h>
 
 #include <grpcpp/channel.h>
+
+#include <internalapi/sensor/collector.pb.h>
 
 #include "CollectionMethod.h"
 #include "HostConfig.h"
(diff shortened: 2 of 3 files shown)


ovalenti (commented):
Using the CollectorConfig object makes sense, good idea !

I think that the next steps should be:
- on the Sensor side:
  - Update the [proto defs](https://github.com/stackrox/stackrox/blob/7b6bc042b9352869ccfbc5e928889cdc61ece7b5/proto/internalapi/sensor/network_connection_iservice.proto) to introduce an enum type wrapping and replacing `NetworkFlowsControlMessage`
  - introduce the runtime-config object in the new enum
  - implement a switch-case in NetworkStatusNotifier in [onRecvControlMessage](https://github.com/stackrox/collector/blob/giles/ROX-26344-integrate-runtime-config/collector/lib/NetworkStatusNotifier.cpp#L58C29-L69) or its successor
  - when receiving a runtime-config in the new switch-case, update the value in CollectorConfig
- on the ConnTracker side:
  - everytime we start a scrape in either `NetworkStatusNotifier::RunSingle` or `NetworkStatusNotifier::RunSingleAfterglow`, set the ExtIP attribute using [`ConnectionTracker::EnableExternalIPs`](https://github.com/stackrox/collector/blob/giles/ROX-26344-integrate-runtime-config/collector/lib/ConnTracker.h#L130C8-L130C25)

ovalenti (dismissed):
As discussed previously and in [#11746 ](https://github.com/stackrox/stackrox/pull/11746), we should remember to unify the names of the environment variables that enable the ExtIP feature. 

https://github.com/stackrox/collector/blob/9828784f88b7b7f07b0af6b5af0653ce85af823a/collector/lib/CollectorConfig.cpp#L50

I think that it would make sense to update CollectorConfig's env var name to match the one in Central ([`ROX_EXTERNAL_IPS`](https://github.com/stackrox/stackrox/blob/ovalenti/learned_external_IPs/pkg/features/list.go#L133)) as part of this PR.

collector/lib/CollectorConfig.h
```diff
@@ -75,7 +78,19 @@ class CollectorConfig {
   bool IsProcessesListeningOnPortsEnabled() const { return enable_processes_listening_on_ports_; }
   bool ImportUsers() const { return import_users_; }
   bool CollectConnectionStatus() const { return collect_connection_status_; }
-  bool EnableExternalIPs() const { return enable_external_ips_; }
+
+  // EnableExternalIPs will check for the existence
+  // of a runtime configuration, and defer to that value
+  // otherwise, we rely on the feature flag (env var)
+  bool EnableExternalIPs() const {
```
JoukoVirtanen, line 85: This function could be called `IsExternalIPsEnabled` or `GetExternalIPsEnabled` to make it clear that this checks the if external IPs should be enabled and it doesn't actually set any values. However, your naming is consistent with the rest of the code, so you can keep it.

collector/lib/CollectorConfig.h
```diff
@@ -75,7 +78,19 @@ class CollectorConfig {
...
+    if (runtime_config_.has_value()) {
+      auto cfg = runtime_config_.value();
+      auto network_cfg = cfg.network_connection_config();
+      return network_cfg.enable_external_ips();
```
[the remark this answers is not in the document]
Stringy: In this case, `network_connection_config()` returns a reference, so it can't be null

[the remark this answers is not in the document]
Molter73: In that case, could we change line 88 slightly to make it clearer?
```cpp
const auto& network_cfg = cfg.network_connection_config();
```

Also, `auto` on its own will create a copy of the object, which we might want to avoid here. So we probably want to change line 87 as well:
```cpp
const auto& cfg = runtime_config_.value();
```

https://stackoverflow.com/questions/29859796/c-auto-vs-auto
https://en.cppreference.com/w/cpp/language/auto

collector/lib/CollectorConfig.h
```diff
@@ -93,6 +108,14 @@ class CollectorConfig {
 
   static std::pair<option::ArgStatus, std::string> CheckConfiguration(const char* config, Json::Value* root);
 
+  void SetRuntimeConfig(sensor::CollectorConfig runtime_config) {
+    runtime_config_ = std::move(runtime_config);
+  }
+
+  std::optional<sensor::CollectorConfig> GetRuntimeConfig() const {
```
Molter73, line 115: Lets not copy the object every time it's accessed.
```suggestion
  const std::optional<sensor::CollectorConfig>& GetRuntimeConfig() const {
```

collector/lib/CollectorConfig.h
```diff
@@ -93,6 +108,14 @@ class CollectorConfig {
 
   static std::pair<option::ArgStatus, std::string> CheckConfiguration(const char* config, Json::Value* root);
 
+  void SetRuntimeConfig(sensor::CollectorConfig runtime_config) {
+    runtime_config_ = std::move(runtime_config);
+  }
```
Molter73, line 113: I know we discussed this offline yesterday, but if we don't expect the caller to keep using the `runtime_config` object, we should probably just use an rvalue here:
```suggestion
  void SetRuntimeConfig(sensor::CollectorConfig&& runtime_config) {
    runtime_config_ = runtime_config;
  }
```
Alternatively, method overloading is a thing, we can have both and let the caller decide if they are `std::move`ing the object or not.

collector/lib/CollectorConfig.cpp
```diff
@@ -47,7 +47,7 @@ BoolEnvVar set_import_users("ROX_COLLECTOR_SET_IMPORT_USERS", false);
 
 BoolEnvVar collect_connection_status("ROX_COLLECT_CONNECTION_STATUS", true);
 
-BoolEnvVar enable_external_ips("ROX_ENABLE_EXTERNAL_IPS", false);
+BoolEnvVar enable_external_ips("ROX_EXTERNAL_IPS", false);
```
Molter73, line 50: I actually think the `ENABLE` makes it clearer what the env variable is for. But I would argue a variable in the form `<PROGRAM>_<FEATURE>_<PARAM>` is even more clear, so in this case we might want to name the variable `ROX_EXTERNAL_IPS_ENABLE` or `ROX_COLLECTOR_EXTERNAL_IPS_ENABLE`

collector/lib/CollectorConfig.h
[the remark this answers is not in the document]
Stringy: I didn't know that was a thing, thanks for the heads up!

collector/lib/CollectorConfig.cpp
Stringy: In general I think I agree. @ovalenti what do you think? should we update the central feature flag?

ovalenti: In Central, it looks like the convention is to use short names and no verb... https://github.com/stackrox/stackrox/blob/0b83c7607cb58d28e0165e20936f30d0c11d12c0/pkg/features/list.go

Although unifying the flag names seemed like a good idea in the first place, the collector one is likely not going to be used directly by any customer. So I am starting to think that we could keep the names as they were, or as Mauro suggests. If it was only my, the thing would not be named `EXTERNAL_IP` at all...

JoukoVirtanen: I don't think `ENABLE` is needed in the name. In stackrox/stackrox the function to check the value of feature flags is `Enable()`, so it would seem to be redundant.

collector/lib/CollectorConfig.h
JoukoVirtanen: Resolving as the naming here is consistent with the rest of the code.

JoukoVirtanen (approved):
Approving, but before merging update the stackrox/stackrox submodule as I merged the internal protobuf definitions. Also get approvals from others before merging.

collector/lib/CollectorConfig.cpp
Molter73: > In stackrox/stackrox the function to check the value of feature flags is `Enable()`, so it would seem to be redundant.

It wouldn't be redundant to a user that is trying to configure collector without any knowledge of what the code in the main repo looks like 🙂

Stringy: I will revert this for now and we can decide on naming conventions later on

PR #1867, opened 2024-09-26
Refactor RunSingle and RunSingleAfterglow into a single method

Molter73:
## Description

Original motivation for this PR was to narrow down a flake we've been having with some of the afterglow tests. However, in order to make it easy to add the log messages, the RunSingle and RunSingleAfterglow methods have been unified, since their difference was minimal.

Unrelated to this, the HandleAfterglowEnvVars method has been slightly rewritten to make its workflow more streamlined and easier to follow.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI should be enough.

### DIFF
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -36,7 +36,8 @@ StringListEnvVar ignored_networks("ROX_IGNORE_NETWORKS", std::vector<std::string
 // Connection endpoints matching a network prefix listed here will never be aggregated.
 StringListEnvVar non_aggregated_networks("ROX_NON_AGGREGATED_NETWORKS", std::vector<std::string>());
 
-BoolEnvVar set_enable_afterglow("ROX_ENABLE_AFTERGLOW", true);
+BoolEnvVar enable_afterglow("ROX_ENABLE_AFTERGLOW", true);
+FloatEnvVar afterglow_period("ROX_AFTERGLOW_PERIOD", 300.0);
 
 BoolEnvVar set_enable_core_dump("ENABLE_CORE_DUMP", false);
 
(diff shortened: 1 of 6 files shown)


Molter73:
After retriggering the tests a bunch of times, I managed to get the flake to show up and determine the event we are missing is the first of the 3 thanks to the added logs in the network status notifier. I believe a connection error at the start of the test is causing it to not complete the first curl correctly and so we miss the event, the script goes on and we do capture the other 2, [setting the script to fail on a failed curl](https://github.com/stackrox/collector/blob/5b53bdc154dd08840fa0e29283caecd6c7798542/integration-tests/container/schedule-curls/schedule-curls.sh#L3) and adding all container logs to the uploaded logs (see #1869) should help us confirm the theory or narrow down the flake even further.

collector/lib/CollectorConfig.cpp
```diff
@@ -278,41 +279,27 @@ void CollectorConfig::InitCollectorConfig(CollectorArgs* args) {
...
   } else {
-    CLOG(ERROR) << "Afterglow period set to 0";
+    if (afterglow_period_micros_ > max_afterglow_period_micros) {
+      CLOG(WARNING) << "User set afterglow period of " << afterglow_period_micros_ / SECOND
+                    << "s is greater than the maximum allowed afterglow period of " << max_afterglow_period_micros / SECOND << "s";
+      CLOG(WARNING) << "Setting the afterglow period to " << max_afterglow_period_micros / SECOND << "s";
+      afterglow_period_micros_ = max_afterglow_period_micros;
+    }
+
+    enable_afterglow_ = enable_afterglow.value();
```
ovalenti, line 299: The patch introduces a change in the behavior here. _Afterglow_ used to be disabled when the period is negative or 0.

ovalenti (commented):
Good job at making the config parsing more uniform. Also nice factorization for the loops !

I have one little comment about the case where the _afterflow period_ is invalid.

collector/lib/CollectorConfig.cpp
Molter73: It shouldn't, I changed the value that is stored in `enable_afterglow_` to be false by default here:
https://github.com/stackrox/collector/blob/5b53bdc154dd08840fa0e29283caecd6c7798542/collector/lib/CollectorConfig.h#L113
So if the period is negative or 0, it will go unchanged and keep afterglow disabled. If the period is properly configured or left with no configuration, `enable_afterglow_` will get changed based on the values in the EnvVars here: https://github.com/stackrox/collector/blob/5b53bdc154dd08840fa0e29283caecd6c7798542/collector/lib/CollectorConfig.cpp#L39-L40

ovalenti: Bad reading of the number of spaces from my part. I thought that the following part was always executed:

```c
enable_afterglow_ = enable_afterglow.value();
```

ovalenti (approved):

PR #1869, opened 2024-09-27
Capture logs for all containers used in tests

Molter73:
## Description

Some flakes we have in CI may be due to errors on the auxiliar containers we run alongside collector. In order to make it easier to debug this type of flakes, container logs could be used to get more information on what is going on.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI should be enough.

- [x] Check log artifacts don't become huge

### DIFF
--- a/integration-tests/pkg/executor/executor_docker_api.go
+++ b/integration-tests/pkg/executor/executor_docker_api.go
@@ -199,49 +199,45 @@ func (c *dockerAPIExecutor) CaptureLogs(testName, containerName string) (string,
 		return "", log.Error("logs error (%v) for container %s\n", err, containerName)
 	}
 
-	if logs == (ContainerLogs{}) {
-		// Nothing to log
-		return "", nil
-	}
-
-	if logs.Stderr == "" || logs.Stdout == "" {
-		// We only have stdout OR stderr to log
-		logName := fmt.Sprintf("%s.log", containerName)
-		logFile, err := common.PrepareLog(testName, logName)
-		if err != nil {
-			return "", err
-		}
-		defer logFile.Close()
-
-		l := logs.GetSingleLog()
-		_, err = logFile.WriteString(l)
-		return l, nil
-	}
-
-	// We need to log both stdout and stderr, do so on separate files
-	for _, log := range []struct {
+	type logFile = struct {
 		name    string
 		content string
-	}{
-		{
-			name:    fmt.Sprintf("%s-stdout.log", containerName),
-			content: logs.Stdout,
-		},
-		{
+	}
+
+	var logFiles []logFile
+	if logs.Empty() {
+		// Nothing to log, still we create an empty file for awareness
+		logFiles = []logFile{{
+			name: fmt.Sprintf("%s.log", containerName),
+		}}
+	} else if logs.Stderr == "" || logs.Stdout == "" {
+		// We only have stdout OR stderr to log
+		logFiles = []logFile{{
+			name:    fmt.Sprintf("%s.log", containerName),
+			content: logs.GetSingleLog(),
+		}}
+	} else {
+		// We need to log both stdout and stderr, do so on separate files
+		logFiles = []logFile{{
 			name:    fmt.Sprintf("%s-stderr.log", containerName),
 			content: logs.Stderr,
-		},
-	} {
-		logFile, err := common.PrepareLog(testName, log.name)
+		}, {
+			name:    fmt.Sprintf("%s-stdout.log", containerName),
+			content: logs.Stdout,
+		}}
+	}
+
+	for _, lf := range logFiles {
+		file, err := common.PrepareLog(testName, lf.name)
 		if err != nil {
 			return "", err
 		}
-		defer logFile.Close()
+		defer file.Close()
 
-		_, err = logFile.WriteString(log.content)
+		file.WriteString(lf.content)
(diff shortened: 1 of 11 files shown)


Molter73:
I checked manually and the artifacts created in the integration tests now hold the logs for the containers that run alongside collector.

Stringy:
For a slight quality of life improvement, can we change the log files to be in a directory per test rather than prefixed with the test name? There's more and more files now, and it'd be nice to separate them to make it easier to go through them

Molter73:
> For a slight quality of life improvement, can we change the log files to be in a directory per test (or per suite?) rather than prefixed with the test name? There's more and more files now, and it'd be nice to separate them to make it easier to go through them

Makes sense, I'll get it done.

Molter73:
Because of how the ansible fetch module works, either flattening files directly onto the destination, overwriting files with the same name, or keeping a long path that includes the hostname and the full path to the files, I decided to use archives and fetch those to ensure no collisions happen. Now when you download the zip archive from GHA and unzip it you get the following structure:
```
.
├── container-logs
│   ├── cos_cos-beta-core_bpf.tar.gz
│   ├── cos_cos-dev-core_bpf.tar.gz
│   ├── cos_cos-stable-core_bpf.tar.gz
│   ├── integration-test.log
│   └── integration-test-report.xml
└── cos-logs.zip
```
If you `tar xzf` one of those archives you get:
```
.
├── container-logs
│   ├── cos_cos-beta
│   │   └── core_bpf
│   │       ├── perf.json
│   │       ├── TestAsyncConnectionBlocked
│   │       │   ├── client.log
│   │       │   ├── collector.log
│   │       │   ├── container-stats.log
│   │       │   └── events.log
│   │       ├── TestAsyncConnectionBlockedWithDisableTracking
│   │       │   ├── client.log
│   │       │   ├── collector.log
│   │       │   ├── container-stats.log
│   │       │   └── events.log
│   │       ├── TestAsyncConnectionSuccess
│   │       │   ├── client-stderr.log
│   │       │   ├── client-stdout.log
│   │       │   ├── collector.log
│   │       │   ├── container-stats.log
│   │       │   ├── events.log
│   │       │   └── server.log
...
```

It's not ideal, but I think it's good enough for now.

integration-tests/pkg/executor/executor_docker_api.go
```diff
@@ -170,30 +170,78 @@ func (d *dockerAPIExecutor) ExecContainer(containerName string, command []string
...
+}
+
+func (c *dockerAPIExecutor) CaptureLogs(testName, containerName string) (string, error) {
+	log.Info("%s: Gathering logs for %q", testName, containerName)
+	logs, err := c.GetContainerLogs(containerName)
+	if err != nil {
+		return "", log.Error("logs error (%v) for container %s\n", err, containerName)
+	}
+
+	if logs == (ContainerLogs{}) {
```
Stringy, line 202: maybe an `Empty` method for the `ContainerLogs` struct might make this a little clearer

integration-tests/pkg/executor/executor_docker_api.go
```diff
@@ -170,30 +170,78 @@ func (d *dockerAPIExecutor) ExecContainer(containerName string, command []string
...
+	}{
+		{
+			name:    fmt.Sprintf("%s-stdout.log", containerName),
+			content: logs.Stdout,
+		},
+		{
+			name:    fmt.Sprintf("%s-stderr.log", containerName),
+			content: logs.Stderr,
+		},
+	} {
```
Stringy, line 234: Consider constructing the list of files dynamically (based on whether we have one or two log files to write) and doing the writes in one go. The current split between having one file, or having two, is a little confusing

```suggestion
    type output struct {
        name string
        content string
    }
    
    var files []output
	if logs.Stderr == "" || logs.Stdout == "" {
	    files = []output {
	        {
	            name: fmt.Sprintf("%s.log", containerName),
	            content: logs.GetSingleLog(),
	        },
	    }
	} else {
	    files = []output {
		    {
			    name:    fmt.Sprintf("%s-stdout.log", containerName),
			    content: logs.Stdout,
		    },
		    {
			    name:    fmt.Sprintf("%s-stderr.log", containerName),
			    content: logs.Stderr,
		    },
	    }
	}

	// We need to log both stdout and stderr, do so on separate files
	for _, log := range files {
```

integration-tests/pkg/executor/executor_docker_api.go
```diff
@@ -170,30 +170,78 @@ func (d *dockerAPIExecutor) ExecContainer(containerName string, command []string
...
+		logFile, err := common.PrepareLog(testName, log.name)
+		if err != nil {
+			return "", err
 		}
-		return sbStdErr.String(), nil
+		defer logFile.Close()
+
+		_, err = logFile.WriteString(log.content)
```
Stringy, line 241: this error is currently ignored - if intentional, we should just use `_`

Molter73: Is this a rustc reference?

Stringy: it isn't :sweat_smile:, not intentionally anyway

Stringy (approved):
LGTM!

PR #1870, opened 2024-09-27
ROX-25719: Receive runtime-configuration in network channel

ovalenti:
## Description

Sensor is now sending RuntimeConfiguration objects controlling External-IP. This occurs over the existing network-connection channel; however the method used was renamed to introduce a new function prototype.

This PR adapts Collector to this new interface to handle the previously existing messages, and expands it by consuming the new RuntimeConfiguration object and forwarding its content in CollectorConfig.

## Checklist


collector/lib/DuplexGRPC.h
```diff
@@ -751,41 +751,41 @@ class StdoutDuplexClientWriter : public IDuplexClientWriter<W> {
     return Result(Status::OK);
   }
 
-  bool Sleep(const gpr_timespec& deadline) {
+  bool Sleep(const gpr_timespec& deadline) override {
     return true;
   }
 
-  Result WritesDoneAsync() {
+  Result WritesDoneAsync() override {
```
SimonBaeumer, line 758: For my info where can I find the definition of `override`?

collector/lib/NetworkConnectionInfoServiceComm.h
```diff
@@ -28,7 +28,7 @@ class INetworkConnectionInfoServiceComm {
 
   virtual sensor::NetworkConnectionInfoService::StubInterface* GetStub() = 0;
 
-  virtual std::unique_ptr<IDuplexClientWriter<sensor::NetworkConnectionInfoMessage>> PushNetworkConnectionInfoOpenStream(std::function<void(const sensor::NetworkFlowsControlMessage*)> receive_func) = 0;
+  virtual std::unique_ptr<IDuplexClientWriter<sensor::NetworkConnectionInfoMessage>> PushNetworkConnectionInfoOpenStream(std::function<void(const sensor::MsgToCollector*)> receive_func) = 0;
```
SimonBaeumer, line 31: Hm, I think it is dangerous to change the existing message, this will break installations which are using Collector and Sensor in different versions.
Is this something we have done before like this? What was the expierence? (e.g. support cases)

Preferably we find a solution which is backward compatible.

collector/lib/DuplexGRPC.h
Molter73: `override` is a keyword, it's specifying `WritesDoneAsync` is a virtual method that is also present in one of the classes this one inherits from, in this case it is defined as a pure virtual method in `IDuplexClient`.
https://github.com/stackrox/collector/blob/55187025b877e44f8b36e5b58f45b7469006d02d/collector/lib/DuplexGRPC.h#L239

https://en.cppreference.com/w/cpp/language/override

SimonBaeumer: Thanks!

collector/lib/NetworkConnectionInfoServiceComm.h
ovalenti: As Sensor and Collector are deployed using the same Daemonset, we usually consider that a version mismatch can only be temporary. Introducing a new method here will make Sensor reply with a "method not found" if there is any mismatch.

erthalion: > As Sensor and Collector are deployed using the same Daemonset

Are they? Sensor has its own Deployment.

SimonBaeumer: @erthalion I think @ovalenti  refers to the same Secured Cluster deployment, in it they share the same version. I think its better to continue the discussion in the slack thread.

PR #1876, opened 2024-10-04
Output the number of times a log has been throttled

Molter73:
## Description

Adding the number of times a log message has been throttled can add value in the sense that having the message show up once or twice may not be as worrisome as it showing up thousands of times.

In order to achieve this, the log message can be split between its header and the actual message. The header will now be handled by classes separate to LogMessage that will hold the information necessary to print them and use a common ILogHeader interface. This is done in order to reduce the amount of code necessary in macros, there shouldn't be a huge performance difference, but the current approach might take a bit more memory in order to store this data, some testing is needed. Calling virtual methods for printing the header could also have a performance impact, but given we don't print that many logs in info severity, would not consider this a problem.

Supersedes #1520. 

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Manually checked logging still works as expected.

Added a throttle message in `system-inspector::Service::GetNext` and checked the output happened every 10 seconds:
image

### DIFF
--- a/collector/lib/Logging.h
+++ b/collector/lib/Logging.h
@@ -3,6 +3,7 @@
 
 #include <chrono>
 #include <cstdint>
+#include <filesystem>
 #include <iomanip>
 #include <iostream>
 #include <sstream>
@@ -54,10 +55,24 @@ void WriteTerminationLog(std::string message);
 
 const size_t LevelPaddingWidth = 7;
 
+/**
+ * Interface for log headers.
+ *
+ * This class defines how log headers should behave and provides some
+ * basic, reusable methods.
+ */
 class ILogHeader {
  public:
   ILogHeader(const char* file, int line, LogLevel level)
-      : file_(file), line_(line), level_(level) {}
+      : line_(line), level_(level) {
+    std::filesystem::path p(file);
+    if (p.has_filename()) {
+      file_ = p.filename().string();
+    } else {
+      file_ = p.string();
+    }
+  }
+
   ILogHeader(const ILogHeader&) = delete;
   ILogHeader(ILogHeader&&) = delete;
   ILogHeader& operator=(const ILogHeader&) = delete;
@@ -81,22 +96,26 @@ class ILogHeader {
 
   void PrintFile() {
     if (CheckLogLevel(LogLevel::DEBUG)) {
-      const char* basename = strrchr(file_, '/');
-      if (basename == nullptr) {
-        basename = file_;
-      } else {
-        ++basename;
-      }
-      std::cerr << "(" << basename << ":" << line_ << ") ";
+      std::cerr << "(" << file_ << ":" << line_ << ") ";
     }
   }
 
  private:
-  const char* file_;
+  std::string file_;
   int line_;
   LogLevel level_;
 };
 
+/**
+ * Basic log header.
+ *
+ * Most log lines will use this header.
+ *
+ * The format when running in debug level will be:
+ * [LEVEL  YYYY/MM/DD HH:mm:SS] (file:line)
+ *
+ * Other log levels exclude the (file:line) part of the header.
+ */
 class LogHeader : public ILogHeader {
  public:
   LogHeader(const char* file, int line, LogLevel level)
@@ -108,6 +127,19 @@ class LogHeader : public ILogHeader {
   }
 };
 
+/**
+ * Throttled log header.
+ *
+ * When the same log message is triggered multiple times, this header
+ * help prevent flooding the logs and instead counts the occurrences
+ * of the message, which will be output after a given time window
+ * expires.
+ *
+ * The format when running in debug level will be:
+ * [LEVEL  YYYY/MM/DD HH:mm:SS] [Throttled COUNT message] (file:line)
+ *
+ * Other log levels exclude the (file:line) part of the header.
+ */
 class ThrottledLogHeader : public ILogHeader {
  public:
   ThrottledLogHeader(const char* file, int line, LogLevel level, std::chrono::duration<unsigned int> interval)
@@ -122,15 +154,15 @@ class ThrottledLogHeader : public ILogHeader {
     PrintFile();
   }
 
-  bool ShouldPrint() {
+  bool Suppress() {
     std::chrono::duration elapsed = std::chrono::steady_clock::now() - last_log_;
     count_++;
     if (elapsed < interval_) {
-      return false;
+      return true;
     }
 
     last_log_ = std::chrono::steady_clock::now();
-    return true;
+    return false;
   }
 
  private:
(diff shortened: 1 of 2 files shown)


ovalenti (approved):
This is much cleaner than the macro version, I like it !

Not too worried by the extra cost of the method calls for the same reasons you expose.

collector/lib/Logging.h
```diff
@@ -71,25 +77,79 @@ class LogMessage {
...
+
+class ThrottledLogHeader : public ILogHeader {
+ public:
+  ThrottledLogHeader(const char* file, int line, LogLevel level, std::chrono::duration<unsigned int> interval)
+      : ILogHeader(file, line, level), interval_(interval) {}
+
+  void PrintHeader() override {
+    PrintPrefix();
+
+    std::cerr << "[Throttled " << count_ << " messages] ";
```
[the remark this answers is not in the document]
Molter73: You are right, but a regular person should also not be confused, this should be an easy fix if we restart `count_` to 1 too or we make `ShouldPrint` always increment it.

Molter73:
Checked memory used on a deployed collector with `kubectl top` and this version has a similar usage to the one that is currently on the master branch of the main repo, so I think we are ok on that front as well.

collector/lib/Logging.h
```diff
@@ -71,25 +77,79 @@ class LogMessage {
...
+  }
 
-    if (throttled_) {
-      std::cerr << "[Throttled] ";
-    }
-
-    if (include_file_) {
+  void PrintFile() {
+    if (CheckLogLevel(LogLevel::DEBUG)) {
       const char* basename = strrchr(file_, '/');
```
erthalion, line 84: It's strictly speaking beyond this PR, but since we have c++17, why don't we use something fancy here, like `std::filesystem::path` instead of a plain string?

collector/lib/Logging.h
```diff
@@ -71,25 +77,79 @@ class LogMessage {
...
+
+    std::cerr << "[Throttled " << count_ << " messages] ";
+    count_ = 0;
+
+    PrintFile();
+  }
+
+  bool ShouldPrint() {
+    std::chrono::duration elapsed = std::chrono::steady_clock::now() - last_log_;
+    count_++;
```
erthalion, line 127: It might cause some surprise in the future (imagine someone else reshuffling this code), that `ShouldPrint` function actually increments the counter. Maybe it would be more obvious to introduce another method `Suppress`, that will increment the count? In that case the macro would look like:
```
if (log_level & cond & CLOG_VAR(stmt).ShouldPrint) {
    LogMessage;
} else {
    CLOG_VAR(stmt).Suppress();
}
```

collector/lib/Logging.h
```diff
@@ -71,25 +77,79 @@ class LogMessage {
...
     }
+  }
+
+ private:
+  const char* file_;
+  int line_;
+  LogLevel level_;
+};
+
+class LogHeader : public ILogHeader {
```
erthalion, line 100: Future us would appreciate a few lines of comments, explaining what's the difference between this and the following classes.

erthalion (approved):
Looks good, thanks. Couple commentaries below, but feel free to decide which of them sound important enough for you to include.

collector/lib/Logging.h
Molter73: I didn't want to do too many changes here outside of counting logs, but I agree that using `std::filesystem::path` makes more sense here, I'll take a look into it.

collector/lib/Logging.h
Molter73: Yeah, I didn't really like this approach but I couldn't think of a better solution. I like your suggestion though, I'll add that.

collector/lib/Logging.h
Molter73: Makes sense, I'll document them.

collector/lib/Logging.h
Molter73: Ok, now I remember why this is a pain, the macro needs to end with `LogMessage(...)` so we can directly `<<` into it. I could do something similar to what you proposed by not using braces around the if/else statements, but that causes a bunch of warnings about a dangling else condition and it's overall ugly.

Best I can think of is renaming `ShouldPrint` to `Suppress`, turn the logic of the returned value around and use that, so at least it's a bit more obvious that `Suppress` is telling the macro to not output the log message and increment the count.

erthalion: Interesting. IIUC currently it does

```
if (log_level && cond && CLOG_VAR(stmt).ShouldPrint())
    LogMessage()
```

How does it return `LogMessage(...)` if the if condition is false?

Molter73: It doesn't, it's a macro, so `LogMessage()` and everything to the `;` is skipped when the condition is false.

image

erthalion: Wouldn't then reversing the if condition solve this conundrum? Like

```
if (!log_level ||!cond || CLOG_VAR(stmt).ShouldPrint()) {
    Suppress()
 } else
    LogMessage(...)
```

Molter73: Problem with that is that missing the log level or having the condition be false would also add to the count of message, I think we need something like:
```cpp
if (log_level && cond) 
  if (!CLOG_VAR(stmt).ShouldPrint()) {
    Suppress()
  } else
    LogMessage(...)
```
But then you have a dangling else and the logic becomes difficult to follow.

You also get this really nice warning a bunch of times when compiling:
```
In file included from /home/mmoltras/go/src/github.com/stackrox/collector/collector/lib/system-inspector/EventExtractor.h:10,
                 from /home/mmoltras/go/src/github.com/stackrox/collector/collector/lib/ContainerMetadata.cpp:5:
/home/mmoltras/go/src/github.com/stackrox/collector/collector/lib/system-inspector/EventExtractor.h: In member function ‘const int64_t* collector::system_inspector::EventExtractor::get_event_rawres(sinsp_evt*)’:
/home/mmoltras/go/src/github.com/stackrox/collector/collector/lib/Logging.h:224:6: warning: suggest explicit braces to avoid ambiguous ‘else’ [-Wdangling-else]
  224 |   if (collector::logging::CheckLogLevel(collector::logging::LogLevel::lvl) && (cond)) \
      |      ^
/home/mmoltras/go/src/github.com/stackrox/collector/collector/lib/Logging.h:230:39: note: in expansion of macro ‘CLOG_THROTTLED_IF’
  230 | #define CLOG_THROTTLED(lvl, interval) CLOG_THROTTLED_IF(true, lvl, interval)
      |                                       ^~~~~~~~~~~~~~~~~
```

erthalion: Well, ok, then just go with whatever is easier + loud commentaries around to explain what this function does :)

PR #1878, opened 2024-10-05
ROX-26516: Collector reads runtime config from configmap

JoukoVirtanen:
## Description

Makes it so that the runtime config can be read from a ConfigMap. The ConfigMap is added to collector here https://github.com/stackrox/stackrox/pull/12910. The ConfigMap is optional and is not needed in order for collector to run. For now the runtime config only enables or disables the external IPs feature. If the ConfigMap is not used the external IPs feature will be controlled by an environment variable.

Json is used for the configuration. This is convenient, because there is a method for converting Json strings to protobuf objects. YAML was considered, but having the ConfigMap use Json makes it easy to copy the ConfigMap into the API, and use it there. Also using a library for YAML, proved to be a pain. 

## Checklist

**Automated testing**
  - [x] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Testing was performed using https://github.com/stackrox/stackrox/pull/12910 which adds the ConfigMap to collector.

Created the following yaml file for a ConfigMap

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: collector-config
  namespace: stackrox
data:
  runtime_config.yaml: |
    networkConnectionConfig:
        enableExternalIps: true
```

Saved the contents in config-map.yml.

Ran
```
kubectl create -f config-map.yml
```

Deployed ACS on a GKE infra cluster using `deploy/deploy-local.sh`.

Set the collector image to one from this PR using `ks edit ds collector`.

The collector logs were checked and the following was found
```
[INFO    2024/10/08 20:46:11] Runtime configuration: network_connection_config {
  enable_external_ips: true
}
```

### enableExternalIps false

The configmap was changed to
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: collector-config
  namespace: stackrox
data:
  runtime_config.yaml: |
    networkConnectionConfig:
        enableExternalIps: false
```

Collector pods were deleted and allowed to come back up
```
ks delete pod -l app=collector
```

```
[INFO    2024/10/08 20:52:24] Runtime configuration: network_connection_config {
}
```

### Invalid config

The configmap was changed to the following
```
apiVersion: v1
data:
  runtime_config.yaml: |
    networkConnectionConfig:
        enableExternalIps: asdf
kind: ConfigMap
metadata:
  creationTimestamp: "2024-10-08T20:51:09Z"
  name: collector-config
  namespace: stackrox
  resourceVersion: "1587598"
  uid: 7594e306-7334-42b5-b439-ed04ae88ffed
```

The collector log had the following
```
[INFO    2024/10/08 20:52:24] Runtime configuration: network_connection_config {
}
```

This is the expected result.

### No config map

The configmap was deleted and collector was restarted. The following was found in the collector log.

There was nothing in the collector logs.

### Different config path

The configmap was changed to the following
```
apiVersion: v1
data:
  collector.yaml: |
    networkConnectionConfig:
        enableExternalIps: true
kind: ConfigMap
metadata:
  creationTimestamp: "2024-10-08T21:00:49Z"
  name: collector-config
  namespace: stackrox
  resourceVersion: "1593991"
  uid: 829de9f9-0f19-4457-927d-b3ea33b8f38f
```

The following were added to the env section of collector

```
        - name: ROX_COLLECTOR_CONFIG_PATH
          value: /etc/stackrox/collector.yaml
```

The following was in the collector log
```
[INFO    2024/10/08 21:05:08] Runtime configuration: network_connection_config {
  enable_external_ips: true
}
```


collector/lib/CollectorConfig.cpp
```diff
@@ -400,6 +406,48 @@ void CollectorConfig::HandleSinspEnvVars() {
   }
 }
 
+void CollectorConfig::HandleConfigMapString(const std::string& jsonString) {
+  sensor::CollectorConfig config;
+  auto status = google::protobuf::util::JsonStringToMessage(jsonString, &config);
+
+  if (!status.ok()) {
+    CLOG(ERROR) << "Failed to parse runtime config. Error: " << status.message();
+    CLOG(ERROR) << "ConfigMap file contents: " << jsonString;
```
JoukoVirtanen, line 415: I have considered making this `FATAL`, but I don't think we want collector to go into a crashloop over this.

collector/lib/CollectorConfig.cpp
```diff
@@ -283,6 +287,8 @@ void CollectorConfig::InitCollectorConfig(CollectorArgs* args) {
   HandleAfterglowEnvVars();
   HandleConnectionStatsEnvVars();
   HandleSinspEnvVars();
+  std::filesystem::path configMapFilePath = runtime_configmap_path.valueOr("/run/collector_runtime_config/config.json");
```
SimonBaeumer, line 290: 1. Why JSON and not YAML like other configurations in Kubernetes and StackRox?

2. nit: You can use refer to the config as `/run/collector/...` instead, this gives the directory more flexibility to be used for different collector configs. The file name could be more explicit and refer to the runtime config.

```suggestion
  std::filesystem::path configMapFilePath = runtime_configmap_path.valueOr("/run/collector/runtime_config.json");
```

collector/lib/CollectorConfig.cpp
SimonBaeumer: It depends, if it is loaded at start-up I think it is okay to crash that users fix the config. 
I agree with you that the config should not crash collector at runtime.

collector/lib/CollectorConfig.cpp
```diff
@@ -400,6 +406,48 @@ void CollectorConfig::HandleSinspEnvVars() {
...
+  }
+}
+
+std::string readJsonFileToString(const std::filesystem::path& filePath) {
+  std::ifstream fileStream(filePath);
+  if (!fileStream.is_open()) {
+    CLOG(WARNING) << "Unable to open ConfigMap file: " << filePath
+                  << ". File may not exist or permissions may be incorrect. "
+                  << "If a ConfigMap was not created intentionally, which is the default, "
+                  << "no action is needed.";
```
Molter73, line 429: This is way too verbose, I propose this function to be re-written to something like this:
```cpp
std::string readJsonFileToString(const std::filesystem::path& filePath) {
  if (!std::filesystem::exists(filePath) {
    CLOG(DEBUG) << "No configuration file found.";
    return "";
  }

  std::ifstream fileStream(filePath);
  if (!fileStream.is_open()) {
    CLOG(WARNING) << "Unable to open configuration file: " << filePath;
    return "";
  }

  std::string content((std::istreambuf_iterator<char>(fileStream)),
                      std::istreambuf_iterator<char>());

  if (content.empty()) {
    CLOG(WARNING) << "Configuration file is empty: " << filePath;
  }
  return content;
}
```

Also, as you may have noticed from the changes, don't refer to the configuration file as a `ConfigMap`, we are configuring it as a configmap in kubernetes, but we are reading a configuration file, not the configmap object.

collector/lib/CollectorConfig.cpp
```diff
@@ -400,6 +406,48 @@ void CollectorConfig::HandleSinspEnvVars() {
...
+  sensor::CollectorConfig config;
+  auto status = google::protobuf::util::JsonStringToMessage(jsonString, &config);
+
+  if (!status.ok()) {
+    CLOG(ERROR) << "Failed to parse runtime config. Error: " << status.message();
+    CLOG(ERROR) << "ConfigMap file contents: " << jsonString;
+  } else {
+    SetRuntimeConfig(config);
+    CLOG(INFO) << "Set the runtime config using a configmap";
+    CLOG(INFO) << config.DebugString();
```
Molter73, line 419: Too verbose IMO, this should be enough and easier to read:
```suggestion
    CLOG(INFO) << "Runtime configuration: " << config.DebugString();
```

collector/lib/CollectorConfig.cpp
Molter73: Alternatively, yaml-cpp can directly read files if we move to use that.

collector/lib/CollectorConfig.cpp
```diff
@@ -400,6 +406,48 @@ void CollectorConfig::HandleSinspEnvVars() {
...
+  if (content.empty()) {
+    CLOG(WARNING) << "ConfigMap file is empty: " << filePath;
+  }
+  return content;
+}
+
+void CollectorConfig::HandleConfigMap(const std::filesystem::path& filePath) {
+  std::string jsonConfig = readJsonFileToString(filePath);
+  if (jsonConfig.empty()) {
+    CLOG(WARNING) << "No runtime configuration specified in ConfigMap.";
```
Molter73, line 445: Doesn't `readJsonFileToString` already warn about this?
```suggestion
```

collector/lib/CollectorConfig.cpp
```diff
@@ -77,6 +79,8 @@ PathEnvVar tls_ca_path("ROX_COLLECTOR_TLS_CA");
 PathEnvVar tls_client_cert_path("ROX_COLLECTOR_TLS_CLIENT_CERT");
 PathEnvVar tls_client_key_path("ROX_COLLECTOR_TLS_CLIENT_KEY");
 
+PathEnvVar runtime_configmap_path("RUNTIME_CONFIGMAP_PATH");
```
Molter73, line 82: Rename this to `ROX_COLLECTOR_CONFIGURATION` or `ROX_COLLECTOR_CONFIGURATION_PATH` and provide a default value directly.
```suggestion
PathEnvVar config_file("ROX_COLLECTOR_CONFIGURATION", "/etc/stackrox/collector.json");
```

collector/lib/CollectorConfig.h
```diff
@@ -188,6 +188,8 @@ class CollectorConfig {
   void HandleAfterglowEnvVars();
   void HandleConnectionStatsEnvVars();
   void HandleSinspEnvVars();
+  void HandleConfigMapString(const std::string& jsonString);
+  void HandleConfigMap(const std::filesystem::path& filePath);
```
Molter73, line 192: Rename this so `ConfigMap` is not part of the name
```suggestion
  void HandleConfigString(const std::string& jsonString);
  void HandleConfig(const std::filesystem::path& filePath);
```

collector/test/CollectorConfigTest.cpp
```diff
@@ -32,6 +34,10 @@ class MockCollectorConfig : public CollectorConfig {
   void MockSetEnableExternalIPs(bool value) {
     SetEnableExternalIPs(value);
   }
+
+  void MockHandleConfigMapString(const std::string& jsonStr) {
+    HandleConfigMapString(jsonStr);
+  }
```
Molter73, line 40: Why do we need a mock method if we are just calling the real thing?
```suggestion
```

collector/test/CollectorConfigTest.cpp
```diff
@@ -145,4 +151,54 @@ TEST(CollectorConfigTest, TestEnableExternalIpsRuntimeConfig) {
...
+           "networkConnectionConfig": {
+           }
+         )";
+
+  MockCollectorConfig config;
+  config.MockHandleConfigMapString(jsonStr);
+
+  EXPECT_FALSE(config.EnableExternalIPs());
+}
+
```
Molter73, line 203: We can probably rewrite all of these as a single test with something like:
```cpp
TEST(CollectorConfigTest, TestConfigurationFile) {
  std::vector<std::pair<std::string, bool>> tests = {
    { R"({
       "networkConnectionConfig": {
         "enableExternalIps": true
       }
     })", true },
    { R"({
           "networkConnectionConfig": {
             "enableExternalIps": false
           }
         })", false },
    // ...
  };
  
  for (const auto& [input, expected] : tests) {
    CollectorConfig config;
    config.HandleConfigMapString(input);
    EXPECT_EQUAL(config.EnableExternalIPs(), expected);
  }
}
```

Molter73: We are also missing a test for an empty string as input.

Molter73 (commented):
Overall looks good, but I still have some comments before I'm ready to approve.

collector/test/CollectorConfigTest.cpp
```diff
@@ -145,4 +151,54 @@ TEST(CollectorConfigTest, TestEnableExternalIpsRuntimeConfig) {
   EXPECT_TRUE(config.EnableExternalIPs());
 }
 
+TEST(CollectorConfigTest, TestConfigMapTrue) {
+  std::string jsonStr = R"({
+           "networkConnectionConfig": {
+             "enableExternalIps": true
```
Molter73, line 157: I also now realize the only place I can actually check what the configuration is going to look like is these tests, which is not great, we need a way to have a formal specification of the configuration that we can properly review, but for now I guess it's fine 🤷🏻‍♂️ .

I do have some objections though:
-  `networkConnectionConfig` is redundant, we know this is a configuration parameter, so it should be `networkConnection`.
- `enableExternalIps` is not extensible.

I would prefer the configuration to look something closer to this:
```yaml
networkConnection:
  externalIPs:
    enable: true
```

JoukoVirtanen: If you want to change these, we need to change the protobuf definitions.

Molter73: Why not? Is not like the feature is GA.

collector/test/CollectorConfigTest.cpp
JoukoVirtanen: Done

collector/lib/CollectorConfig.cpp
JoukoVirtanen: I am now using yaml instead of json.

collector/lib/CollectorConfig.cpp
JoukoVirtanen: For now this is at startup, so I will continue to use `FATAL`.

collector/lib/CollectorConfig.cpp
```diff
@@ -400,6 +405,61 @@ void CollectorConfig::HandleSinspEnvVars() {
...
+  }
+
+  bool enableExternalIps = false;
+  try {
+    if (networkConnectionConfig["enableExternalIps"]) {
+      enableExternalIps = networkConnectionConfig["enableExternalIps"].as<bool>();
+    }
+  } catch (const std::exception& e) {
+    CLOG(FATAL) << "An error occurred getting enableExternalIps";
+  }
```
Molter73, line 426: You should be able to directly provide a default here
```suggestion
  if (networkConnectionConfig["enableExternalIps"]) {
    enableExternalIps = networkConnectionConfig["enableExternalIps"].as<bool>(false);
  }
```

collector/lib/CollectorConfig.cpp
```diff
@@ -400,6 +405,61 @@ void CollectorConfig::HandleSinspEnvVars() {
...
+
+  return true;
+}
+
+void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
+  if (!std::filesystem::exists(filePath)) {
+    CLOG(DEBUG) << "No configuration file found. " << filePath;
+    return;
+  }
+  YAML::Node yamlConfig = YAML::LoadFile(filePath);
```
Molter73, line 442:
```suggestion
```

collector/lib/CollectorConfig.cpp
```diff
@@ -400,6 +405,61 @@ void CollectorConfig::HandleSinspEnvVars() {
...
+    CLOG(FATAL) << "Failed to parse the configuration file: " << filePath
+                << ". Error: " << e.what();
+  } catch (const YAML::Exception& e) {
+    CLOG(FATAL) << "An error occurred while loading the configuration file: " << filePath
+                << ". Error: " << e.what();
+  } catch (const std::exception& e) {
+    CLOG(FATAL) << "An unexpected error occurred while trying to read: " << filePath << e.what();
+  }
+
+  YamlConfigToConfig(yamlConfig);
```
Molter73, line 460:
```suggestion
```

collector/lib/CollectorConfig.cpp
```diff
@@ -4,6 +4,8 @@
 #include <optional>
 #include <sstream>
 
+#include <yaml-cpp/yaml.h>
```
Molter73, line 7: This include is already added to `CollectorConfig.h`, so it is not needed here.
```suggestion
```

collector/test/CollectorConfigTest.cpp
```diff
@@ -32,6 +34,10 @@ class MockCollectorConfig : public CollectorConfig {
   void MockSetEnableExternalIPs(bool value) {
     SetEnableExternalIPs(value);
   }
+
+  bool MockYamlConfigToConfig(YAML::Node& yamlConfig) {
+    return YamlConfigToConfig(yamlConfig);
+  }
```
Molter73, line 40: Same question I asked previously, why do we need a mock method if we are just calling the real method? You should be able to instantiate a `CollectorConfig` in the tests, instead of a `MockCollectorConfig`.
```suggestion
```

collector/test/CollectorConfigTest.cpp
```diff
@@ -145,4 +151,64 @@ TEST(CollectorConfigTest, TestEnableExternalIpsRuntimeConfig) {
   EXPECT_TRUE(config.EnableExternalIPs());
 }
 
+TEST(CollectorConfigTest, TestYamlConfigToConfigMultiple) {
+  std::vector<std::tuple<std::string, bool, bool>> tests = {
```
Molter73, line 155: Having 2 bools in the vector of tests does not make it obvious that one is for the expected value and the other is for checking if the yaml is valid or not. Split invalid yaml to its own test and keep this one for checking the value of valid yaml configuration. This should also simplify the tests overall.

Molter73: I also now see there is just one case for invalid yaml, so splitting it to its own makes even more sense to me.

collector/test/CollectorConfigTest.cpp
```diff
@@ -145,4 +151,64 @@ TEST(CollectorConfigTest, TestEnableExternalIpsRuntimeConfig) {
...
+  }
+}
+
+TEST(CollectorConfigTest, TestYamlConfigToConfigEmpty) {
+  std::string yamlStr = R"()";
+  YAML::Node yamlNode = YAML::Load(yamlStr);
+
+  MockCollectorConfig config;
+
+  EXPECT_DEATH({ config.MockYamlConfigToConfig(yamlNode); }, ".*");
```
Molter73, line 207: Why do we need to check the value of runtime_config after this statement? `EXPECT_DEATH` means the program has already stopped, nothing will be done with the value.

collector/test/CollectorConfigTest.cpp
```diff
@@ -145,4 +151,64 @@ TEST(CollectorConfigTest, TestEnableExternalIpsRuntimeConfig) {
...
+       false, false}};
+
+  for (const auto& [yamlStr, expected, valid] : tests) {
+    YAML::Node yamlNode = YAML::Load(yamlStr);
+
+    MockCollectorConfig config;
+
+    bool result = config.MockYamlConfigToConfig(yamlNode);
+    std::optional<sensor::CollectorConfig> runtime_config;
+    runtime_config = config.GetRuntimeConfig();
```
Molter73, line 187:
```suggestion
    auto runtime_config = config.GetRuntimeConfig();
```

collector/lib/CollectorConfig.cpp
```diff
@@ -77,6 +79,8 @@ PathEnvVar tls_ca_path("ROX_COLLECTOR_TLS_CA");
 PathEnvVar tls_client_cert_path("ROX_COLLECTOR_TLS_CLIENT_CERT");
 PathEnvVar tls_client_key_path("ROX_COLLECTOR_TLS_CLIENT_KEY");
 
+PathEnvVar config_file("ROX_COLLECTOR_CONFIG_PATH", "/etc/stackrox/collector.json");
```
Molter73, line 82: This should be `.yml` or `.yaml` now, shouldn't it?
```suggestion
PathEnvVar config_file("ROX_COLLECTOR_CONFIG_PATH", "/etc/stackrox/collector.yaml");
```

collector/lib/CollectorConfig.cpp
JoukoVirtanen: Done

collector/test/CollectorConfigTest.cpp
JoukoVirtanen: `YamlConfigToConfig` is protected, so I do need to use the mock.

collector/test/CollectorConfigTest.cpp
JoukoVirtanen: Done

collector/test/CollectorConfigTest.cpp
JoukoVirtanen: Removed checks after this.

collector/lib/CollectorConfig.cpp
```diff
@@ -400,6 +403,56 @@ void CollectorConfig::HandleSinspEnvVars() {
...
+void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
+  if (!std::filesystem::exists(filePath)) {
+    CLOG(DEBUG) << "No configuration file found. " << filePath;
+    return;
+  }
+
+  try {
+    YAML::Node yamlConfig = YAML::LoadFile(filePath);
+    YamlConfigToConfig(yamlConfig);
+  } catch (const YAML::BadFile& e) {
```
ovalenti, line 442: As I understand it, at one point, we will call this several times after startup to re-read the configuration.
I just want to underline that in this case, re-reading and consuming the config will likely happen in different threads and we should guard against race condition.

It is probably fine for now if we are calling `HandleConfig()` only at startup. Do we want to implement locking now or commit to do it in a subsequent PR ?

ovalenti (commented):
The multi-threaded angle.

collector/test/CollectorConfigTest.cpp
Molter73: No, you don't, you can just use a `FRIEND_TEST`
http://google.github.io/googletest/reference/testing.html#FRIEND_TEST

You can find an example on how to use it here:
https://github.com/stackrox/collector/blob/0cbc2f0a2846122d203495e66815be8c0dc20832/collector/lib/system-inspector/Service.h#L47
https://github.com/stackrox/collector/blob/0cbc2f0a2846122d203495e66815be8c0dc20832/collector/test/SystemInspectorServiceTest.cpp#L8

collector/lib/CollectorConfig.cpp
Molter73: I've actually discussed this with @Stringy when he was some of these things, we need to start reading configuration from a centralized static `CollectorConfig` and a `std::shared_mutex` will need to be put in place so we can lock it exclusively to update configuration and shared for only reading it. It'd be nice if the mechanism is broad enough to lock the entire configuration, that way we can start moving everything to the configuration file.

For now, I would go with the existing approach, since there is no actual updating going on while running and change it in a follow up, as long as we are all aware this is a thing we need to do we should be fine. 

Good call out though!

collector/test/CollectorConfigTest.cpp
Molter73: I see the whole test file is doing this though, so I'll just resolve the comment and refactor the thing in a follow up.

collector/test/CollectorConfigTest.cpp
```diff
@@ -145,4 +151,71 @@ TEST(CollectorConfigTest, TestEnableExternalIpsRuntimeConfig) {
...
+
+    MockCollectorConfig config;
+
+    bool result = config.MockYamlConfigToConfig(yamlNode);
+    auto runtime_config = config.GetRuntimeConfig();
+
+    EXPECT_TRUE(result);
+    EXPECT_TRUE(runtime_config.has_value());
+
+    if (runtime_config.has_value()) {
```
Molter73, line 187: You don't need the if, you already expected `runtime_config.has_value()` to be true.
```suggestion
    EXPECT_TRUE(runtime_config.has_value());
```

Molter73 (approved):
LGTM!

collector/lib/CollectorConfig.cpp
JoukoVirtanen: Let's implement locking when we actually need it.

docs/references.md
```diff
@@ -104,6 +106,35 @@ seconds. The default value is 30 seconds.
 
 * `logLevel`: Sets logging level. The default is INFO.
 
+### File mount or ConfigMap
```
SimonBaeumer, line 109: Nice to document it!

SimonBaeumer (approved):

collector/test/CollectorConfigTest.cpp
JoukoVirtanen: Done

PR #1879, opened 2024-10-08
ROX-26516: Introspection endpoint for runtime config

JoukoVirtanen:
## Description

Makes it possible to find the current state of the runtime configuration by querying a collector endpoint.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Deployed ACS on a GKE cluster. Changed the collector image to the one for this PR. Also added the following environment variable

```
        - name: ROX_COLLECTOR_INTROSPECTION_ENABLE
          value: "true"
```

Ran the following command
```
kubectl -n stackrox port-forward ds/collector 8080:8080 > /dev/null 2>&1 &
```

Queried the endpoint
```
curl http://localhost:8080/state/config
```
and got this response
```
{"networkConnectionConfig":{"enableExternalIps":true}}
```

### Set enableExternalIps to false

The result in this case was
```
{"networkConnectionConfig":{}}
```

### Not using a ConfigMap

In this case the response was
```
{}
```


ovalenti:
I believe that the builder regeneration logic bases its decision on modifications that are part of the PR, which is incorrect here (the base PR modifies builder). Added the `build-builder-image` with the hope that it fixes the build.

ovalenti (changes requested):

collector/lib/CollectorConfigInspector.cpp
```diff
@@ -0,0 +1,61 @@
...
+const std::string CollectorConfigInspector::kBaseRoute = "/state/config";
+
+CollectorConfigInspector::CollectorConfigInspector(const std::shared_ptr<CollectorConfig> config) : config_(config) {
+}
+
+Json::Value CollectorConfigInspector::configToJson() {
+  Json::Value root;
+  std::optional<sensor::CollectorConfig> runtime_config = config_->GetRuntimeConfig();
+
+  if (runtime_config.has_value()) {
```
Molter73, line 24: Early returns FTW!
```suggestion
  if (!runtime_config.has_value()) {
    return root;
  }
```

collector/lib/CollectorConfigInspector.cpp
```diff
@@ -0,0 +1,61 @@
...
+namespace collector {
+
+const std::string CollectorConfigInspector::kBaseRoute = "/state/config";
+
+CollectorConfigInspector::CollectorConfigInspector(const std::shared_ptr<CollectorConfig> config) : config_(config) {
+}
+
+Json::Value CollectorConfigInspector::configToJson() {
+  Json::Value root;
+  std::optional<sensor::CollectorConfig> runtime_config = config_->GetRuntimeConfig();
```
Molter73, line 22:
```suggestion
  const auto& runtime_config = config_->GetRuntimeConfig();
```
Otherwise, you'll be creating a copy.

collector/lib/CollectorConfigInspector.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "ConnTracker.h"
+
+namespace collector {
+
+const std::string CollectorConfigInspector::kBaseRoute = "/state/config";
+
+CollectorConfigInspector::CollectorConfigInspector(const std::shared_ptr<CollectorConfig> config) : config_(config) {
+}
+
+Json::Value CollectorConfigInspector::configToJson() {
```
Molter73, line 20: There are a few situations in this method that could be considered errors, like failing to convert the configuration to JSON, it'd be nice if we could change the HTTP status to 500 if one of these occur, maybe we can return a tuple with the result and the json? Or use a `std::variant` of sorts?

collector/lib/CollectorConfigInspector.cpp
```diff
@@ -0,0 +1,61 @@
...
+      return Json::Value();
+    }
+
+    Json::CharReaderBuilder readerBuilder;
+    std::string errs;
+    std::istringstream iss(jsonString);
+    if (!Json::parseFromStream(readerBuilder, iss, &root, &errs)) {
+      CLOG(ERROR) << "Failed to parse JSON string: " << errs;
+      return Json::Value();
+    }
```
Molter73, line 40: If the caller to this method is just going to grab the `Json::Value` object and turn it into a string, why are we grabbing a string and turning it into a `Json::Value` here? Why can't we just return the string?

collector/lib/CollectorConfigInspector.h
```diff
@@ -0,0 +1,30 @@
+#ifndef COLLECTOR_CONFIGINSPECTOR_H
+#define COLLECTOR_CONFIGINSPECTOR_H
```
Molter73, line 2: Just for consistency
```suggestion
#ifndef COLLECTOR_CONFIG_INSPECTOR_H
#define COLLECTOR_CONFIG_INSPECTOR_H
```

collector/lib/CollectorConfigInspector.cpp
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  std::string jsonString;
+  const auto& config = runtime_config.value();
+  google::protobuf::util::Status status = google::protobuf::util::MessageToJsonString(config, &jsonString);
+
+  if (!status.ok()) {
+    isError = true;
+    CLOG(WARNING) << "Failed to convert protobuf object to JSON: " << status.ToString();
+    return "{\"error\": \"Failed to convert protobuf object to JSON\"";
```
Molter73, line 29: [nit]
```suggestion
    return R"({"error": "Failed to convert protobuf object to JSON")";
```

collector/lib/CollectorConfigInspector.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+  if (req_info == nullptr) {
+    return ServerError(conn, "unable to read request");
+  }
+
+  bool isError = false;
+  std::string jsonString = configToJson(isError);
+
+  if (isError) {
+    mg_printf(conn, "HTTP/1.1 500 OK\r\nContent-Type: application/json\r\nConnection: close\r\n\r\n");
```
Molter73, line 46:
```suggestion
    mg_printf(conn, "HTTP/1.1 500 Internal Server Error\r\nContent-Type: application/json\r\nConnection: close\r\n\r\n");
```

docs/troubleshooting.md
```diff
@@ -602,3 +602,8 @@ $ curl "http://<collector>:8080/state/network/connection?container=c6f030bc4b42&
   ]
 }
 ```
+
+### Config endpoing
```
Molter73, line 606:
```suggestion
### Config endpoint
```

Molter73 (approved):
Just some minor nitpicks, but otherwise LGTM!

collector/lib/CollectorConfigInspector.cpp
JoukoVirtanen: Done

collector/lib/CollectorConfigInspector.cpp
JoukoVirtanen: Done

PR #1889, opened 2024-10-12
ROX-26637: Reread configmap when it is updated

JoukoVirtanen:
## Description

Makes it so that when the ConfigMap or config file is updated collector reads it and updates the runtime configuration. Also handles cases were the ConfigMap does not exist at collector startup, is created later, and is deleted during runtime.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Deployed ACS on a GKE cluster using `deploy/deploy-local.sh`. Replaced the collector image with one from this PR. The connection reported by collector to sensor where logged. Initially there was no ConfigMap. 

Before deploying ACS the following deployment was created

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: external-dns
spec:
  replicas: 2
  selector:
    matchLabels:
      app: external-dns-app
  template:
    metadata:
      labels:
        app: external-dns-app
    spec:
      containers:
      - name: dns-app-container
        image: alpine:latest
        ports:
        - containerPort: 8080
        env:
        - name: EXTERNAL_IP
          value: "8.8.8.8"  # Google's Public DNS IP
        - name: EXTERNAL_PORT
          value: "53"  # DNS runs on port 53
        command: ["sh", "-c"]
        args:
        - |
          # Connect to Google's DNS IP
          while true; do
            nc -zv $EXTERNAL_IP $EXTERNAL_PORT;
            sleep 60;
          done
---
apiVersion: v1
kind: Service
metadata:
  name: external-dns-service
spec:
  selector:
    app: external-dns-app
  ports:
    - protocol: TCP
      port: 80
      targetPort: 8080
  type: ClusterIP
```

One of the containers of the pods was the following
```
$ kubectl describe pod external-dns-54879dc46d-jfjdp | grep "Container ID"
    Container ID:  containerd://df6e7a3a43d22e448d8fe49ccda1dd5b428bd3c8d9f18b69c67145cb4b87fcdb
```

The following was observed in the collector logs.
```
[INFO    2024/10/15 18:41:31] df6e7a3a43d2: :::0 -> 255.255.255.255/0:53 [tcp] active:1
```
Along with many other connections. No logging involving the runtime configuration or the ConfigMap was found. Note that the normalized connection above corresponds to the one created by the deployment specified above.

The following ConfigMap was then created
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: collector-config
  namespace: stackrox
data:
  runtime_config.yaml: |
    networking:
        externalIps:
          enable: true
```

The following was observed in the collector logs
```
[INFO    2024/10/15 18:44:39] Runtime configuration:
[INFO    2024/10/15 18:44:39] networking {
  external_ips {
    enable: true
  }
}

[INFO    2024/10/15 18:44:39] Waiting for file event
```
and
```
[INFO    2024/10/15 18:45:31] df6e7a3a43d2: :::0 -> 8.8.8.8:53 [tcp] active:1
[INFO    2024/10/15 18:49:31] df6e7a3a43d2: :::0 -> 255.255.255.255/0:53 [tcp] active:0
```

The connection with the un-normalized IP address is reported as active and after the afterglow period is over the normalized connection is reported as inactive.

External IPs was then disabled by editing the ConfigMap to the following

```
apiVersion: v1
data:
  runtime_config.yaml: |
    networking:
        externalIps:
          enable: false
kind: ConfigMap
metadata:
  creationTimestamp: "2024-10-15T18:43:37Z"
  name: collector-config
  namespace: stackrox
  resourceVersion: "1046248"
  uid: 8722e73b-ffb9-45fe-a52d-a13132383779
```

The following was observed in the collector logs

```
[INFO    2024/10/15 18:59:06] Got event event->mask= 1024
[INFO    2024/10/15 18:59:06] event->mask & IN_MODIFY= 0
[INFO    2024/10/15 18:59:06] File was removed
[INFO    2024/10/15 18:59:06] Runtime configuration:
[INFO    2024/10/15 18:59:06] networking {
  external_ips {
  }
}

[INFO    2024/10/15 18:59:06] Got event event->mask= 32768
[INFO    2024/10/15 18:59:06] event->mask & IN_MODIFY= 0
[INFO    2024/10/15 18:59:06] File was removed
[INFO    2024/10/15 18:59:06] Runtime configuration:
[INFO    2024/10/15 18:59:06] networking {
  external_ips {
  }
}

[INFO    2024/10/15 18:59:06] Waiting for file event
```
and

```
[INFO    2024/10/15 18:59:32] df6e7a3a43d2: :::0 -> 255.255.255.255/0:53 [tcp] active:1
[INFO    2024/10/15 19:03:32] df6e7a3a43d2: :::0 -> 8.8.8.8:53 [tcp] active:0
```
The connection with the normalized IP address is reported open and the connection with the un-normalized IP address is reported closed after the afterglow period.

The ConfigMap was deleted and the collector logs were checked

```
[INFO    2024/10/15 21:12:45] Got event event->mask= 1024
[INFO    2024/10/15 21:12:45] event->mask & IN_MODIFY= 0
[INFO    2024/10/15 21:12:45] File was removed
[INFO    2024/10/15 21:12:45] The configuration file has been deleted. Resetting the configuration
[INFO    2024/10/15 21:15:08] 7fbe41e81a75: :::0 -> 255.255.255.255/0:53 [udp] active:0
[INFO    2024/10/15 21:15:08] a6523116184e: :::0 -> 104.16.0.0/13:443 [tcp] active:1
[INFO    2024/10/15 21:17:38] 2b9112868a44: :::0 -> 255.255.255.255/0:53 [udp] active:0
[INFO    2024/10/15 21:20:38] 038e0a736e89: :::0 -> 255.255.255.255/0:53 [udp] active:0
[INFO    2024/10/15 21:21:38] a6523116184e: :::0 -> 104.16.0.0/13:443 [tcp] active:0
``` 

The collector logs show that the configuration was reset. The connection that has been traced above did not change state so did not appear in the logs, but other connections are shown to be normalized.

The original ConfigMap was then recreated

The following was observed in the collector logs

```
[INFO    2024/10/15 21:29:08] Runtime configuration:
[INFO    2024/10/15 21:29:08] networking {
  external_ips {
    enable: true
  }
}

[INFO    2024/10/15 21:29:08] Got event event->mask= 32768
[INFO    2024/10/15 21:29:08] event->mask & IN_MODIFY= 0
[INFO    2024/10/15 21:29:08] File was removed
[INFO    2024/10/15 21:29:08] Runtime configuration:
[INFO    2024/10/15 21:29:08] networking {
  external_ips {
    enable: true
  }
}

[INFO    2024/10/15 21:29:08] Waiting for file event
```

```
[INFO    2024/10/15 21:29:38] df6e7a3a43d2: :::0 -> 8.8.8.8:53 [tcp] active:1
[INFO    2024/10/15 21:33:38] df6e7a3a43d2: :::0 -> 255.255.255.255/0:53 [tcp] active:0
```

The connection with the un-normalized IP address is reported open and after the afterglow period the normalized version is reported as being closed.

All of this is as expected.

### DIFF
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -281,7 +281,6 @@ void CollectorConfig::InitCollectorConfig(CollectorArgs* args) {
   HandleAfterglowEnvVars();
   HandleConnectionStatsEnvVars();
   HandleSinspEnvVars();
-  HandleConfig(config_file.value());
 
   host_config_ = ProcessHostHeuristics(*this);
 }
@@ -400,6 +399,7 @@ void CollectorConfig::HandleSinspEnvVars() {
 }
 
 void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
+  // Don't read the file during a scrape
   std::lock_guard<std::mutex> lock(*mutex_);
 
   if (yamlConfig.IsNull() || !yamlConfig.IsDefined()) {
@@ -455,6 +455,12 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
   }
 }
 
+void WaitForFileToExist(const std::filesystem::path& filePath) {
+  while (!std::filesystem::exists(filePath)) {
+    sleep(5);
+  }
+}
+
 void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
   CLOG(INFO) << "In WatchConfigFile";
   int fd = inotify_init();
@@ -463,6 +469,8 @@ void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
     return;
   }
 
+  WaitForFileToExist(filePath);
+  HandleConfig(filePath);
   int wd = inotify_add_watch(fd, filePath.c_str(), IN_MODIFY | IN_MOVE_SELF | IN_DELETE_SELF);
   if (wd < 0) {
     CLOG(FATAL) << "Failed to add inotify watch for " << filePath;
@@ -483,10 +491,11 @@ void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
       CLOG(INFO) << "Got event event->mask= " << event->mask;
       CLOG(INFO) << "event->mask & IN_MODIFY= " << (event->mask & IN_MODIFY);
       if (event->mask & IN_MODIFY) {
-        CLOG(INFO) << "File was replace";
+        CLOG(INFO) << "File was replaced";
         HandleConfig(filePath);
       } else if ((event->mask & IN_MOVE_SELF) || (event->mask & IN_DELETE_SELF) || (event->mask & IN_IGNORED)) {
         CLOG(INFO) << "File was removed";
+        WaitForFileToExist(filePath);
         wd = inotify_add_watch(fd, filePath.c_str(), IN_MODIFY | IN_MOVE_SELF | IN_DELETE_SELF);
         HandleConfig(filePath);
         if (wd < 0) {
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -400,6 +400,8 @@ void CollectorConfig::HandleSinspEnvVars() {
 }
 
 void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
+  std::lock_guard<std::mutex> lock(*mutex_);
+
   if (yamlConfig.IsNull() || !yamlConfig.IsDefined()) {
     CLOG(FATAL) << "Unable to read config from config file";
     return;
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -485,7 +485,8 @@ void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
   HandleConfig(filePath);
   int wd = inotify_add_watch(fd, filePath.c_str(), IN_MODIFY | IN_MOVE_SELF | IN_DELETE_SELF);
   if (wd < 0) {
-    CLOG(FATAL) << "Failed to add inotify watch for " << filePath;
+    std::string errorMsg = "Failed to add inotify watch for " + filePath.string();
+    LogRuntimeConfigError(errorMsg);
     close(fd);
     return;
   }
@@ -495,7 +496,8 @@ void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
     CLOG(INFO) << "Waiting for file event";
     int length = read(fd, buffer, sizeof(buffer));
     if (length < 0) {
-      CLOG(FATAL) << "Unable to read event for " << filePath;
+      std::string errorMsg = "Unable to read event for " + filePath.string();
+      LogRuntimeConfigError(errorMsg);
     }
 
     for (int i = 0; i < length; i += sizeof(struct inotify_event)) {
@@ -509,12 +511,13 @@ void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
         CLOG(INFO) << "File was removed";
         WaitForFileToExist(filePath);
         wd = inotify_add_watch(fd, filePath.c_str(), IN_MODIFY | IN_MOVE_SELF | IN_DELETE_SELF);
-        HandleConfig(filePath);
         if (wd < 0) {
-          CLOG(FATAL) << "Failed to add inotify watch for " << filePath;
+          std::string errorMsg = "Failed to add inotify watch for " + filePath.string();
+          LogRuntimeConfigError(errorMsg);
           close(fd);
           return;
         }
+        HandleConfig(filePath);
       }
     }
   }
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -5,6 +5,7 @@
 #include <sstream>
 
 #include <sys/inotify.h>
+#include <sys/select.h>
 
 #include <libsinsp/sinsp.h>
 
@@ -473,7 +474,7 @@ int CollectorConfig::WaitForInotifyAddWatch(int fd, const std::filesystem::path&
   while (!thread_.should_stop()) {
     int wd = inotify_add_watch(fd, filePath.c_str(), IN_MODIFY | IN_MOVE_SELF | IN_DELETE_SELF);
     if (wd < 0) {
-      CLOG_THROTTLED(ERROR, std::chrono::seconds(30)) << "Failed to add inotify watch for " << filePath;
+      CLOG_THROTTLED(ERROR, std::chrono::seconds(30)) << "Failed to add inotify watch for " << filePath << ": (" << errno << ") " << StrError();
     } else {
       return wd;
     }
@@ -484,24 +485,52 @@ int CollectorConfig::WaitForInotifyAddWatch(int fd, const std::filesystem::path&
 }
 
 void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
+  fd_set rfds;
+  struct timeval tv = {};
   int fd = inotify_init();
   if (fd < 0) {
     CLOG(ERROR) << "inotify_init() failed: " << StrError();
     CLOG(ERROR) << "Runtime configuration will not be used.";
     return;
   }
+
   WaitForFileToExist(filePath);
   int wd = WaitForInotifyAddWatch(fd, filePath);
+  if (wd == -1) {
+    CLOG(ERROR) << "Failed to add inotify watch, runtime configuration will not be used.";
+    close(fd);
+    return;
+  }
+
   bool success = HandleConfig(filePath);
   if (!success) {
     CLOG(FATAL) << "Unable to parse configuration file: " << filePath;
   }
 
   char buffer[1024];
+
   while (!thread_.should_stop()) {
+    tv.tv_sec = 2;
+    tv.tv_usec = 0;
+    FD_ZERO(&rfds);
+    FD_SET(fd, &rfds);
+
+    int retval = select(FD_SETSIZE, &rfds, nullptr, nullptr, &tv);
+    if (retval < 0) {
+      CLOG(WARNING) << "'select' call failed: (" << errno << ") " << StrError(errno);
+      continue;
+    }
+
+    if (retval == 0) {
+      // select timed out, check if we should be stopping.
+      continue;
+    }
+
+    // Received event from inotify.
     int length = read(fd, buffer, sizeof(buffer));
     if (length < 0) {
       CLOG_THROTTLED(ERROR, std::chrono::seconds(30)) << "Unable to read event for " << filePath;
+      continue;
     }
 
     struct inotify_event* event;
@@ -509,18 +538,17 @@ void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
       event = (struct inotify_event*)&buffer[i];
       if (event->mask & IN_MODIFY) {
         HandleConfig(filePath);
-      } else if ((event->mask & IN_MOVE_SELF) || (event->mask & IN_DELETE_SELF) || (event->mask & IN_IGNORED)) {
-        WaitForFileToExist(filePath);
+      } else if ((event->mask & IN_MOVE_SELF) || (event->mask & IN_DELETE_SELF)) {
         inotify_rm_watch(fd, wd);
-        fd = inotify_init();
-        if (fd < 0) {
-          CLOG(ERROR) << "inotify_init() failed: " << StrError();
-          CLOG(ERROR) << "Runtime configuration will no longer be used. Reverting to default values.";
-          std::unique_lock<std::shared_mutex> lock(mutex_);
-          runtime_config_.reset();
-          return;
-        }
+
+        WaitForFileToExist(filePath);
         wd = WaitForInotifyAddWatch(fd, filePath);
+        if (wd == -1) {
+          // Only situation that could get us here is if collector is
+          // stopping, so we break out and let the thread finish.
+          break;
+        }
+
         HandleConfig(filePath);
       }
     }
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -4,6 +4,8 @@
 #include <optional>
 #include <sstream>
 
+#include <sys/inotify.h>
+
 #include <libsinsp/sinsp.h>
 
 #include "CollectionMethod.h"
@@ -449,6 +451,49 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
   } catch (const std::exception& e) {
     CLOG(FATAL) << "An unexpected error occurred while trying to read: " << filePath << e.what();
   }
+}
+
+void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
+  int fd = inotify_init();
+  if (fd < 0) {
+    CLOG(FATAL) << "inotify_init failed";
+    return;
+  }
+
+  int wd = inotify_add_watch(fd, filePath.c_str(), IN_MODIFY);
+  if (wd < 0) {
+    CLOG(FATAL) << "Failed to add inotify watch for " << filePath;
+    close(fd);
+    return;
+  }
+
+  char buffer[1024];
+  while (true) {
+    int length = read(fd, buffer, sizeof(buffer));
+    if (length < 0) {
+      CLOG(FATAL) << "read error";
+    }
+
+    for (int i = 0; i < length; i += sizeof(struct inotify_event)) {
+      struct inotify_event* event = (struct inotify_event*)&buffer[i];
+      if (event->mask & IN_MODIFY) {
+        CLOG(INFO) << "Configuration file modified. Reloading...";
+        HandleConfig(filePath);  // Reload configuration
+      }
+    }
+  }
+
+  inotify_rm_watch(fd, wd);
+  close(fd);
+}
+
+void CollectorConfig::Start() {
+  thread_.Start([this] { WatchConfigFile(config_file.value()); });
+  CLOG(INFO) << "Watching config file: " << config_file.value();
+}
+
+void CollectorConfig::Stop() {
+  thread_.Stop();
 }
 
 bool CollectorConfig::TurnOffScrape() const {
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -398,12 +398,23 @@ void CollectorConfig::HandleSinspEnvVars() {
   }
 }
 
+void CollectorConfig::LogRuntimeConfigError(const std::string& str) {
+  if (runtime_config_.has_value()) {
+    // If a runtime config already exists and there is an error setting it
+    // log the error and continue with the existing configuration.
+    CLOG(ERROR) << str;
+  } else {
+    // If there is no runtime configuration and there is an error setting it, exit.
+    CLOG(FATAL) << str;
+  }
+}
+
 void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
   // Don't read the file during a scrape
   std::unique_lock<std::shared_mutex> lock(mutex_);
 
   if (yamlConfig.IsNull() || !yamlConfig.IsDefined()) {
-    CLOG(FATAL) << "Unable to read config from config file";
+    LogRuntimeConfigError("Unable to read config from config file");
     return;
   }
   YAML::Node networking = yamlConfig["networking"];
@@ -442,16 +453,17 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
     YAML::Node yamlConfig = YAML::LoadFile(filePath);
     YamlConfigToConfig(yamlConfig);
   } catch (const YAML::BadFile& e) {
-    CLOG(FATAL) << "Failed to open the configuration file: " << filePath
-                << ". Error: " << e.what();
+    std::string errorMsg = "Failed to open the configuration file: " + filePath.string() + ". Error: " + e.what();
+    LogRuntimeConfigError(errorMsg);
   } catch (const YAML::ParserException& e) {
-    CLOG(FATAL) << "Failed to parse the configuration file: " << filePath
-                << ". Error: " << e.what();
+    std::string errorMsg = "Failed to parse the configuration file: " + filePath.string() + ". Error: " + e.what();
+    LogRuntimeConfigError(errorMsg);
   } catch (const YAML::Exception& e) {
-    CLOG(FATAL) << "An error occurred while loading the configuration file: " << filePath
-                << ". Error: " << e.what();
+    std::string errorMsg = "An error occurred while loading the configuration file: " + filePath.string() + ". Error: " + e.what();
+    LogRuntimeConfigError(errorMsg);
   } catch (const std::exception& e) {
-    CLOG(FATAL) << "An unexpected error occurred while trying to read: " << filePath << e.what();
+    std::string errorMsg = "An unexpected error occurred while trying to read: " + filePath.string() + e.what();
+    LogRuntimeConfigError(errorMsg);
   }
 }
 
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -471,7 +471,6 @@ int WaitForInotifyAddWatch(int fd, const std::filesystem::path& filePath) {
     int wd = inotify_add_watch(fd, filePath.c_str(), IN_MODIFY | IN_MOVE_SELF | IN_DELETE_SELF);
     if (wd < 0) {
       CLOG_THROTTLED(ERROR, std::chrono::seconds(30)) << "Failed to add inotify watch for " << filePath;
-      close(fd);
     } else {
       return wd;
     }
@@ -500,12 +499,14 @@ void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
       CLOG(ERROR) << "Unable to read event for " << filePath;
     }
 
-    for (int i = 0; i < length; i += sizeof(struct inotify_event)) {
-      struct inotify_event* event = (struct inotify_event*)&buffer[i];
+    struct inotify_event *event = buffer;
+    for (int i = 0; i < length; i += sizeof(struct inotify_event) + event->len) {
+      event = (struct inotify_event*)&buffer[i];
       if (event->mask & IN_MODIFY) {
         HandleConfig(filePath);
       } else if ((event->mask & IN_MOVE_SELF) || (event->mask & IN_DELETE_SELF) || (event->mask & IN_IGNORED)) {
         WaitForFileToExist(filePath);
+        inotify_rm_watch(fd, wd);
         wd = WaitForInotifyAddWatch(fd, filePath);
         HandleConfig(filePath);
       }
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -400,7 +400,7 @@ void CollectorConfig::HandleSinspEnvVars() {
 
 void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
   // Don't read the file during a scrape
-  std::lock_guard<std::mutex> lock(*mutex_);
+  std::unique_lock<std::shared_mutex> lock(mutex_);
 
   if (yamlConfig.IsNull() || !yamlConfig.IsDefined()) {
     CLOG(FATAL) << "Unable to read config from config file";
@@ -457,7 +457,7 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
 
 void WaitForFileToExist(const std::filesystem::path& filePath) {
   while (!std::filesystem::exists(filePath)) {
-    sleep(5);
+    sleep(1);
   }
 }
 
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -459,14 +459,11 @@ bool CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
 
 void CollectorConfig::WaitForFileToExist(const std::filesystem::path& filePath) {
   int count = 0;
-  while (!std::filesystem::exists(filePath)) {
+  while (!std::filesystem::exists(filePath) && !thread_.should_stop()) {
     sleep(1);
     count++;
     if (count > 45) {
       runtime_config_.reset();
-    }
-    if (thread_.should_stop()) {
-      break;
     }
   }
 }
--- a/collector/lib/CollectorConfig.cpp
+++ b/collector/lib/CollectorConfig.cpp
@@ -398,24 +398,12 @@ void CollectorConfig::HandleSinspEnvVars() {
   }
 }
 
-void CollectorConfig::LogRuntimeConfigError(const std::string& str) {
-  if (runtime_config_.has_value()) {
-    // If a runtime config already exists and there is an error setting it
-    // log the error and continue with the existing configuration.
-    CLOG(ERROR) << str;
-  } else {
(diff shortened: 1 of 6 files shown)


collector/lib/CollectorConfig.cpp
```diff
@@ -447,6 +451,73 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
...
 }
 
+void WaitForFileToExist(const std::filesystem::path& filePath) {
+  while (!std::filesystem::exists(filePath)) {
+    sleep(5);
+  }
+}
+
+void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
+  CLOG(INFO) << "In WatchConfigFile";
```
JoukoVirtanen, line 461: Will remove before merging.

collector/lib/CollectorConfig.cpp
```diff
@@ -447,6 +451,73 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
...
+  int fd = inotify_init();
+  if (fd < 0) {
+    CLOG(FATAL) << "inotify_init failed";
+    return;
+  }
+
+  WaitForFileToExist(filePath);
```
JoukoVirtanen, line 468: It seems that `inotify` doesn't work unless the file exists to begin with, so I think this is needed.

collector/lib/CollectorConfig.cpp
```diff
@@ -447,6 +451,73 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
...
+    CLOG(INFO) << "Waiting for file event";
+    int length = read(fd, buffer, sizeof(buffer));
+    if (length < 0) {
+      CLOG(FATAL) << "Unable to read event for " << filePath;
+    }
+
+    for (int i = 0; i < length; i += sizeof(struct inotify_event)) {
+      struct inotify_event* event = (struct inotify_event*)&buffer[i];
+      CLOG(INFO) << "Got event event->mask= " << event->mask;
+      CLOG(INFO) << "event->mask & IN_MODIFY= " << (event->mask & IN_MODIFY);
```
JoukoVirtanen, line 488: Remove before merging.

collector/lib/CollectorConfig.cpp
```diff
@@ -447,6 +451,73 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
...
+      if (event->mask & IN_MODIFY) {
+        CLOG(INFO) << "File was replaced";
```
JoukoVirtanen, line 490: Remove before merging.

collector/lib/CollectorConfig.cpp
```diff
@@ -447,6 +451,73 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
...
+        HandleConfig(filePath);
+      } else if ((event->mask & IN_MOVE_SELF) || (event->mask & IN_DELETE_SELF) || (event->mask & IN_IGNORED)) {
+        CLOG(INFO) << "File was removed";
```
JoukoVirtanen, line 493: Remove before merging.

collector/lib/CollectorConfig.cpp
```diff
@@ -447,6 +451,73 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
...
+        if (wd < 0) {
+          CLOG(FATAL) << "Failed to add inotify watch for " << filePath;
+          close(fd);
+          return;
+        }
+      }
+    }
+  }
+
+  CLOG(INFO) << "Leaving WatchingConfigFile " << filePath;
```
JoukoVirtanen, line 506: Probably keep this, but change what is logged. I don't know what it should be changed to.

collector/lib/CollectorConfig.h
```diff
@@ -193,12 +200,15 @@ class CollectorConfig {
   std::optional<TlsConfig> tls_config_;
 
   std::optional<sensor::CollectorConfig> runtime_config_;
+  StoppableThread thread_;
+  std::mutex* mutex_ = new std::mutex();
```
JoukoVirtanen, line 204: I don't know where this mutex should live. It is used by `CollectorConfig` and `NetworkStatusNotifier`. It will likely be used by other components in the future.

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -301,6 +303,7 @@ sensor::NetworkConnectionInfoMessage* NetworkStatusNotifier::CreateInfoMessage(c
 
 void NetworkStatusNotifier::AddConnections(::google::protobuf::RepeatedPtrField<sensor::NetworkConnection>* updates, const ConnMap& delta) {
   for (const auto& delta_entry : delta) {
+    CLOG(INFO) << delta_entry.first << " active:" << delta_entry.second.IsActive();
```
JoukoVirtanen, line 306: Change to `DEBUG` before merging.

collector/lib/CollectorConfig.h
JoukoVirtanen: This should probably be a shared pointer.

collector/lib/CollectorConfig.cpp
```diff
@@ -398,6 +399,9 @@ void CollectorConfig::HandleSinspEnvVars() {
 }
 
 bool CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
+  // Don't read the file during a scrape
+  std::lock_guard<std::mutex> lock(*mutex_);
```
Molter73, line 403: With the changes to the type of mutex being used, this needs to be:
```suggestion
  std::unique_lock<std::shared_mutex> lock(mutex_);
```

collector/lib/CollectorConfig.cpp
```diff
@@ -398,6 +399,9 @@ void CollectorConfig::HandleSinspEnvVars() {
...
+
   if (yamlConfig.IsNull() || !yamlConfig.IsDefined()) {
     CLOG(FATAL) << "Unable to read config from config file";
```
Molter73, line 406: This will cause collector to crash loop if the configuration is invalid, this is fine during startup, but not so good if collector was already running. Ideally, we would try to parse the configuration into a copy, check said configuration is valid and then swap it with the configuration being actively used. That way, if the configuration is changed in an invalid way, we could alert of this situation and keep collector running with the old configuration.

For the time being, we can keep the current implementation, but I would try to fix this ASAP.

collector/lib/CollectorConfig.cpp
```diff
@@ -447,6 +451,73 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
   }
 }
 
+void WaitForFileToExist(const std::filesystem::path& filePath) {
+  while (!std::filesystem::exists(filePath)) {
+    sleep(5);
```
Molter73, line 456: Is this stopping the entirety of collector or just the inotify thread?

Molter73: Also, have you tested how long k8s takes to recreate the config file when updated? 5 seconds sounds like a lot, we can probably check every second.
```suggestion
    sleep(1);
```

collector/lib/CollectorConfig.cpp
```diff
@@ -447,6 +451,73 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
...
+    close(fd);
+    return;
+  }
+
+  char buffer[1024];
+  while (true) {
+    CLOG(INFO) << "Waiting for file event";
+    int length = read(fd, buffer, sizeof(buffer));
+    if (length < 0) {
+      CLOG(FATAL) << "Unable to read event for " << filePath;
```
Molter73, line 482: I wouldn't kill collector completely just because of an inotify error, can we go into a state of trying to readd the watch instead?

collector/lib/CollectorConfig.cpp
```diff
@@ -447,6 +451,73 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
...
+      if (event->mask & IN_MODIFY) {
+        CLOG(INFO) << "File was replaced";
+        HandleConfig(filePath);
+      } else if ((event->mask & IN_MOVE_SELF) || (event->mask & IN_DELETE_SELF) || (event->mask & IN_IGNORED)) {
+        CLOG(INFO) << "File was removed";
+        WaitForFileToExist(filePath);
+        wd = inotify_add_watch(fd, filePath.c_str(), IN_MODIFY | IN_MOVE_SELF | IN_DELETE_SELF);
+        HandleConfig(filePath);
+        if (wd < 0) {
+          CLOG(FATAL) << "Failed to add inotify watch for " << filePath;
```
Molter73, line 498: Similar to my previous comment, I wouldn't kill collector completely in this situation.

collector/lib/NetworkStatusNotifier.cpp
Molter73: I would prefer it removed 🙂

collector/lib/CollectorConfig.cpp
JoukoVirtanen: It just stops the inotify thread.

collector/lib/NetworkStatusNotifier.cpp
JoukoVirtanen: Will remove before merging.

collector/lib/CollectorConfig.h
JoukoVirtanen: Done

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -224,6 +224,8 @@ void NetworkStatusNotifier::RunSingle(IDuplexClientWriter<sensor::NetworkConnect
   int64_t time_at_last_scrape = NowMicros();
 
   while (writer->Sleep(next_scrape)) {
+    // We don't want to scrape while reading the config file
+    std::lock_guard<std::mutex> lock(*(config_->GetMutex()));
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

collector/lib/CollectorConfig.cpp
JoukoVirtanen: I have added a WaitForInotifyAddWatch function.

collector/lib/CollectorConfig.cpp
JoukoVirtanen: Now this only fatal is `runtime_config_` is not set.

collector/lib/CollectorConfig.cpp
```diff
@@ -438,17 +453,88 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
...
+    int wd = inotify_add_watch(fd, filePath.c_str(), IN_MODIFY | IN_MOVE_SELF | IN_DELETE_SELF);
+    if (wd < 0) {
+      CLOG_THROTTLED(ERROR, std::chrono::seconds(30)) << "Failed to add inotify watch for " << filePath;
+      close(fd);
+    } else {
+      return;
+    }
+    sleep(1);
+  }
+}
```
Molter73, line 492: You only need to initialize inotify once, this method should look like this:
```suggestion
int WaitForInotifyAddWatch(int fd, const std::filesystem::path& filePath) {
  while (true) {
    int wd = inotify_add_watch(fd, filePath.c_str(), IN_MODIFY | IN_MOVE_SELF | IN_DELETE_SELF);
    if (wd < 0) {
      CLOG_THROTTLED(ERROR, std::chrono::seconds(30)) << "Failed to add inotify watch for " << filePath;
      close(fd);
    } else {
      return wd;
    }
    sleep(1);
  }
}
```

collector/lib/CollectorConfig.cpp
```diff
@@ -438,17 +453,88 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
...
+
+void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
+  CLOG(INFO) << "In WatchConfigFile";
+  int fd, wd;
```
Molter73, line 496: Initialize inotify here, the errors that can occur during this initialization will probably be fatal, since they are due to resource exhaustion.
```suggestion
  int fd = inotify_init();
  if (fd < 0) {
    CLOG(ERROR) << "inotify_init() failed: "<< StrError();
    CLOG(ERROR) << "Runtime configuration will not be used.";
    return;
  }
```

collector/lib/CollectorConfig.cpp
```diff
@@ -438,17 +453,88 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
...
+      struct inotify_event* event = (struct inotify_event*)&buffer[i];
+      CLOG(INFO) << "Got event event->mask= " << event->mask;
+      CLOG(INFO) << "event->mask & IN_MODIFY= " << (event->mask & IN_MODIFY);
+      if (event->mask & IN_MODIFY) {
+        CLOG(INFO) << "File was replaced";
+        HandleConfig(filePath);
+      } else if ((event->mask & IN_MOVE_SELF) || (event->mask & IN_DELETE_SELF) || (event->mask & IN_IGNORED)) {
+        CLOG(INFO) << "File was removed";
+        WaitForFileToExist(filePath);
+        WaitForInotifyAddWatch(fd, wd, filePath);
```
Molter73, line 519: Aren't you leaking `wd` and `fd` here?

collector/lib/CollectorConfig.cpp
```diff
@@ -438,17 +453,88 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
     YAML::Node yamlConfig = YAML::LoadFile(filePath);
     YamlConfigToConfig(yamlConfig);
   } catch (const YAML::BadFile& e) {
-    CLOG(FATAL) << "Failed to open the configuration file: " << filePath
-                << ". Error: " << e.what();
+    std::string errorMsg = "Failed to open the configuration file: " + filePath.string() + ". Error: " + e.what();
+    LogRuntimeConfigError(errorMsg);
```
Molter73, line 457: Instead of this, you can have `HandleConfig` return a boolean for file parsing success.
```cpp
CLOG(ERROR) << "Failed to open the configuration file: " << filePath
                << ". Error: " << e.what();
return false;
```

Then you can use this returned value during initialization to do a `CLOG(FATAL)` and ignore it during regular operation. This also allows dropping the `LogRuntimeConfigError` method, which is not very ergonomic IMO.

collector/lib/CollectorConfig.cpp
```diff
@@ -438,17 +453,88 @@ void CollectorConfig::HandleConfig(const std::filesystem::path& filePath) {
...
+      CLOG(INFO) << "Got event event->mask= " << event->mask;
+      CLOG(INFO) << "event->mask & IN_MODIFY= " << (event->mask & IN_MODIFY);
+      if (event->mask & IN_MODIFY) {
+        CLOG(INFO) << "File was replaced";
+        HandleConfig(filePath);
+      } else if ((event->mask & IN_MOVE_SELF) || (event->mask & IN_DELETE_SELF) || (event->mask & IN_IGNORED)) {
+        CLOG(INFO) << "File was removed";
+        WaitForFileToExist(filePath);
+        WaitForInotifyAddWatch(fd, wd, filePath);
+        HandleConfig(filePath);
```
Molter73, line 520: What happens if the configuration file was removed and is never again recreated? shouldn't we fall back to not using runtime configuration?

collector/lib/CollectorConfig.cpp
JoukoVirtanen: This was intentional so that fd and wd would be available for
```
  inotify_rm_watch(fd, wd);
  close(fd);
```
but with the changes that you suggest, this is not needed.

collector/lib/CollectorConfig.cpp
JoukoVirtanen: I don't think so. If the file is removed, we should go back to waiting for it to be recreated and continue to use current configuration. There could be someone who instead of editing the configmap, deletes it and creates a new one.

Molter73: Then if you enable runtime configuration you are forever tied to it until the next restart? That doesn't sound right.

JoukoVirtanen: You can set the configmap to the default.

JoukoVirtanen: I am open to clearing the configuration when the configmap is deleted, but we should continue to watch for the creation of the configmap.

JoukoVirtanen: Now the runtime configuration is reset when the confgmap is deleted, but it still waits for the configmap to be created.

collector/lib/CollectorConfig.cpp
```diff
@@ -428,27 +431,107 @@ void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
...
+  }
+
+  return true;
+}
+
+void CollectorConfig::WaitForFileToExist(const std::filesystem::path& filePath) {
+  while (!std::filesystem::exists(filePath)) {
+    if (runtime_config_.has_value()) {
+      CLOG(INFO) << "The configuration file has been deleted. Resetting the configuration";
+      runtime_config_.reset();
```
JoukoVirtanen, line 464: The config is reset here, because sometimes inotify will report the file as being removed if it is modified.

collector/lib/CollectorConfig.cpp
```diff
@@ -422,33 +425,106 @@ void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
...
+
+  char buffer[1024];
+  while (true) {
+    int length = read(fd, buffer, sizeof(buffer));
+    if (length < 0) {
+      CLOG(ERROR) << "Unable to read event for " << filePath;
+    }
+
+    for (int i = 0; i < length; i += sizeof(struct inotify_event)) {
+      struct inotify_event* event = (struct inotify_event*)&buffer[i];
```
Molter73, line 504: inotify events have a variable length array in them, you need to adjust for that here
```suggestion
    struct inotify_event *event = buffer;
    for (int i = 0; i < length; i += sizeof(struct inotify_event) + event->len) {
      event = (struct inotify_event*)&buffer[i];
```

Molter73 (commented):
I was looking into the inotify documentation and I'm fairly sure we could add a watch to the parent directory of the configuration file to watch for it to be created instead of polling, but we can do that in a follow up.

collector/lib/CollectorConfig.cpp
```diff
@@ -422,33 +425,106 @@ void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
   externalIpsConfig->set_enable(enableExternalIps);
 
   SetRuntimeConfig(runtime_config);
-  CLOG(INFO) << "Runtime configuration:";
-  CLOG(INFO) << GetRuntimeConfigStr();
+  CLOG(INFO) << "Runtime configuration:\n" + GetRuntimeConfigStr();
```
SimonBaeumer, line 428: C++ related question for me: Is there something like a `string` function which is called automatically when used in a formatter, eg in `prinftf("%s", runtimeConfig)`?

collector/lib/CollectorConfig.cpp
SimonBaeumer: I think when reloading the config it should be cached, and on user error load the already loaded configuration.
E.g. a misconfiguration with invalid YAML or deleting the config by accident.

Not blocking the PR for this, open for discussion for how critical collector team sees these points. (i don't think it is high prio though)

collector/lib/CollectorConfig.cpp
```diff
@@ -422,33 +425,106 @@ void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
...
+  WaitForFileToExist(filePath);
+  int wd = WaitForInotifyAddWatch(fd, filePath);
+  bool success = HandleConfig(filePath);
+  if (!success) {
+    CLOG(FATAL) << "Unable to parse configuration file: " << filePath;
   }
+
+  char buffer[1024];
+  while (true) {
+    int length = read(fd, buffer, sizeof(buffer));
```
SimonBaeumer, line 498: C++ understanding: Is my understanding correct that `inotify` registers a file descripter on `fd`, after registering an infinite loop is started to `read` the fd for the inotify events into the `buffer`.

The buffer is casted to an `inotify_event` and we check if it is the `IN_MODIFY` event.

Basically, how I understand this now, is that the we need to handle the inotify events and wait for `read` continuously,.

Is `read` a blocking call or would the loop be halted until an event is tirggered?

SimonBaeumer (commented):
Thx for adding me to the review.
@Molter73 can review it better, but I love to learn more about C++ and Collector.

collector/lib/CollectorConfig.cpp
Molter73: In C++ we usually just use the output stream operator (`operator<<`) which handles formatted output, you usually overload it for your classes and then use it directly. There are stream manipulators for handling certain aspects of the output like whether `true` and `false` should be integers or alphanumeric (https://en.cppreference.com/w/cpp/io/manip).

C++23 added [std::format](https://en.cppreference.com/w/cpp/io/print) which is actually pretty nice, but because it is in C++23, it won't be used until 2032 in production projects.

collector/lib/CollectorConfig.cpp
Molter73: > C++ understanding: Is my understanding correct that `inotify` registers a file descripter on `fd`, after registering an infinite loop is started to `read` the fd for the inotify events into the `buffer`.
> 
> The buffer is casted to an `inotify_event` and we check if it is the `IN_MODIFY` event.
> 
> Basically, how I understand this now, is that the we need to handle the inotify events and wait for `read` continuously,.

Yes, this is all correct.

> Is `read` a blocking call or would the loop be halted until an event is tirggered?

In this instance, `read` is a blocking call, so the thread will stop until an event occurs, from `man inotify`:
```
   Reading events from an inotify file descriptor
       To determine what events have occurred, an application read(2)s from the inotify file descriptor.  If no events have
       so far occurred, then, assuming a blocking file descriptor, read(2) will block until at least one event occurs  (un‐
       less interrupted by a signal, in which case the call fails with the error EINTR; see signal(7)).
```

Molter73:
> I think when reloading the config it should be cached, and on user error load the already loaded configuration.
E.g. a misconfiguration with invalid YAML or deleting the config by accident.

We had this conversation, I agree that misconfiguration should rollback to the existing one.

Removing the file on the other hand should fallback to the default behavior. This is because the existence of the file is tied to runtime configuration being enabled or not, Jouko proposed having that controlled by a separate feature flag, but adding a feature flag to enable reading a configuration flag that controls other features is... Yeah, I don't like that.

In the future we'll probably have to put some more thought on it, the way I'd like it to work is that we have defaults, not having the file rolls everything to the defaults, those defaults could potentially be set by a separate file. For now I kinda just want the feature merged.

collector/lib/CollectorConfig.cpp
Molter73: Before I forget, you can also just call `printf` from C++, as you can with any C function, it's just not best practices because it leads to writing "C style C++", which is terrible and you should definitely never do :pokerface:

collector/lib/CollectorConfig.cpp
```diff
@@ -422,33 +425,107 @@ void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
...
+  }
+  WaitForFileToExist(filePath);
+  int wd = WaitForInotifyAddWatch(fd, filePath);
+  bool success = HandleConfig(filePath);
+  if (!success) {
+    CLOG(FATAL) << "Unable to parse configuration file: " << filePath;
   }
+
+  char buffer[1024];
+  while (true) {
```
Molter73, line 496: I just realized these `while (true)` will cause the thread to block forever and will most likely cause issues when stopping collector, we should change them to `while (!should_stop())` and ensure the more nested loops in `WaitForFileToExist` and `WaitForInotifyAddWatch` can return and let the thread stop if needed.

collector/lib/CollectorConfig.cpp
```diff
@@ -398,9 +399,11 @@ void CollectorConfig::HandleSinspEnvVars() {
 }
 
 void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
+  // Don't read the file during a scrape
+  std::unique_lock<std::shared_mutex> lock(mutex_);
+
   if (yamlConfig.IsNull() || !yamlConfig.IsDefined()) {
-    CLOG(FATAL) << "Unable to read config from config file";
-    return;
+    throw std::runtime_error("Unable to read config from config file");
```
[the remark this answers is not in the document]
JoukoVirtanen: I have made a ticket to make improvements to this code in 4.7 https://issues.redhat.com/browse/ROX-26643

collector/lib/NetworkStatusNotifier.cpp
JoukoVirtanen: Done

collector/lib/CollectorConfig.cpp
JoukoVirtanen: I fixed all cases of `while (true)`.

collector/lib/CollectorConfig.cpp
JoukoVirtanen: I am now just using `<<`.

collector/test/CollectorConfigTest.cpp
```diff
@@ -223,7 +223,7 @@ TEST(CollectorConfigTest, TestYamlConfigToConfigEmpty) {
 
   MockCollectorConfig config;
 
-  EXPECT_DEATH({ config.MockYamlConfigToConfig(yamlNode); }, ".*");
+  EXPECT_THROW({ config.MockYamlConfigToConfig(yamlNode); }, std::runtime_error);
```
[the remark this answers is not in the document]
JoukoVirtanen: Done

collector/lib/CollectorConfig.cpp
JoukoVirtanen: I am now waiting for 45 seconds before resetting the configuration.

collector/lib/CollectorConfig.cpp
```diff
@@ -422,33 +425,116 @@ void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
...
+  return true;
+}
+
+void CollectorConfig::WaitForFileToExist(const std::filesystem::path& filePath) {
+  int count = 0;
+  while (!std::filesystem::exists(filePath)) {
+    sleep(1);
+    count++;
+    if (count > 45) {
+      runtime_config_.reset();
```
Molter73, line 466:
```suggestion
      std::unique_lock<std::shared_mutex> lock(mutex_);
      runtime_config_.reset();
```

Molter73 (approved):

collector/lib/CollectorConfig.cpp
SimonBaeumer: Thanks Mauro!

SimonBaeumer:
> > I think when reloading the config it should be cached, and on user error load the already loaded configuration.
> > E.g. a misconfiguration with invalid YAML or deleting the config by accident.
> 
> We had this conversation, I agree that misconfiguration should rollback to the existing one.
> 
> Removing the file on the other hand should fallback to the default behavior. This is because the existence of the file is tied to runtime configuration being enabled or not, Jouko proposed having that controlled by a separate feature flag, but adding a feature flag to enable reading a configuration flag that controls other features is... Yeah, I don't like that.
> 
> In the future we'll probably have to put some more thought on it, the way I'd like it to work is that we have defaults, not having the file rolls everything to the defaults, those defaults could potentially be set by a separate file. For now I kinda just want the feature merged.

Makes sense :+1:

collector/lib/CollectorConfig.cpp
```diff
@@ -422,33 +425,120 @@ void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
...
+
+void CollectorConfig::WatchConfigFile(const std::filesystem::path& filePath) {
+  int fd = inotify_init();
+  if (fd < 0) {
+    CLOG(ERROR) << "inotify_init() failed: " << StrError();
+    CLOG(ERROR) << "Runtime configuration will not be used.";
+    return;
+  }
+  WaitForFileToExist(filePath);
+  int wd = WaitForInotifyAddWatch(fd, filePath);
```
Molter73, line 494: With the newer changes for properly stopping the thread, after calling `WaitForFileToExist` and `WaitForInotifyAddWatch` you need to check if we are trying to stop or not. Otherwise, the code will think we got out of those calls because the file exists or the watcher is already set.
```suggestion
  WaitForFileToExist(filePath);
  int wd = WaitForInotifyAddWatch(fd, filePath);
  if (thread_.shoud_stop()) {
    close(fd);
    return;
  }
```

You probably need something similar on line 523, otherwise `HandleConfig` will be called without the file existing and mess up.

collector/lib/CollectorConfig.cpp
```diff
@@ -422,33 +425,120 @@ void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
...
+    count++;
+    if (count > 45) {
+      std::unique_lock<std::shared_mutex> lock(mutex_);
+      runtime_config_.reset();
+    }
+  }
+}
+
+int CollectorConfig::WaitForInotifyAddWatch(int fd, const std::filesystem::path& filePath) {
+  while (!thread_.should_stop()) {
```
Molter73, line 473: Now I'm thinking, why does this methid need a loop? If we already waited for the file to exist, is not like the OS is goning to forbid adding the watch now and randomly decide to allow it the next second, I think we should be able to just remove this:
```suggestion
```

collector/lib/CollectorConfig.cpp
```diff
@@ -422,33 +425,120 @@ void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
...
+    if (length < 0) {
+      CLOG_THROTTLED(ERROR, std::chrono::seconds(30)) << "Unable to read event for " << filePath;
+    }
+
+    struct inotify_event* event;
+    for (int i = 0; i < length; i += sizeof(struct inotify_event) + event->len) {
+      event = (struct inotify_event*)&buffer[i];
+      if (event->mask & IN_MODIFY) {
+        HandleConfig(filePath);
+      } else if ((event->mask & IN_MOVE_SELF) || (event->mask & IN_DELETE_SELF) || (event->mask & IN_IGNORED)) {
```
Molter73, line 512: The infinite loop you are getting is due to `IN_IGNORED` in this condition, from the man page:
```
           IN_IGNORED
                  Watch was removed explicitly (inotify_rm_watch(2)) or automatically (file was deleted, or filesystem  was
                  unmounted).  See also BUGS.
```

That means we are getting a move or delete event, then manually removing the watch, which triggers an IN_IGNORED event, we get that, go back in this place, remove the watch create it again get another IN_IGNORED event, etc...

Just remove `IN_IGNORED`, move or delete should be enough, we don't need to do anything in the ignore case.
```suggestion
      } else if ((event->mask & IN_MOVE_SELF) || (event->mask & IN_DELETE_SELF)) {
```

collector/lib/CollectorConfig.cpp
```diff
@@ -422,33 +425,120 @@ void CollectorConfig::YamlConfigToConfig(YAML::Node& yamlConfig) {
...
+  WaitForFileToExist(filePath);
+  int wd = WaitForInotifyAddWatch(fd, filePath);
+  bool success = HandleConfig(filePath);
+  if (!success) {
+    CLOG(FATAL) << "Unable to parse configuration file: " << filePath;
+  }
+
+  char buffer[1024];
+  while (!thread_.should_stop()) {
+    int length = read(fd, buffer, sizeof(buffer));
```
Molter73, line 502: Another random thought I just had, does this `read` not block collector from stopping correctly? Do we need to use `select` with a timeout on the file descriptor and decide if we want to stop or retry reading events?

collector/lib/CollectorConfig.cpp
JoukoVirtanen: In the future `inotify` will monitor the directory for the creation of the file, so `WaitForFileToExist` will not be needed.

Molter73 (approved):

PR #1890, opened 2024-10-13
Don't read the ConfigMap when runtime config not used

JoukoVirtanen:
## Description

Make it so that the runtime configuration is only read when `ROX_COLLECTOR_RUNTIME_CONFIG_ENABLED` is set to true.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)

### DIFF
--- a/collector/lib/CollectorService.cpp
+++ b/collector/lib/CollectorService.cpp
@@ -58,7 +58,9 @@ void CollectorService::RunForever() {
   prometheus::Exposer exposer("9090");
   exposer.RegisterCollectable(registry);
 
-  config_->Start();
+  if (config_->EnableRuntimeConfig()) {
+    config_->Start();
+  }
   CollectorStatsExporter exporter(registry, config_, &system_inspector_);
 
   std::unique_ptr<NetworkStatusNotifier> net_status_notifier;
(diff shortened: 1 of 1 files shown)


collector/lib/CollectorService.cpp
```diff
@@ -58,7 +58,9 @@ void CollectorService::RunForever() {
   prometheus::Exposer exposer("9090");
   exposer.RegisterCollectable(registry);
 
-  config_->Start();
+  if (config_->EnableRuntimeConfig()) {
```
Molter73, line 61: `EnableRuntimeConfig()` was originally put in place to start capturing namespace information from Falco. If this is no longer needed and/or the purpose of the configuration parameter has been changed, please adjust it accordingly, we don't want to have a performance impact for something we are not using.

JoukoVirtanen: I think that is one of the purposes of `EnableRuntimeConfig()`. I think the main purpose is to control if runtime configuration is enabled or not. That is why it is called what it is and not `EnableNamespaceInfo()`. If runtime configuration is not used then there is no need to obtain namespace information. Also if runtime configuration is not used, the runtime configuration should not be obtained from the ConfigMap.

PR #1898, opened 2024-10-18
Allow to ignore process arguments

erthalion:
## Description

Add an internal option to not collect process arguments, as a workaround for [ROX-25845](https://issues.redhat.com/browse/ROX-25845).

## Checklist

**Automated testing**
  - [x] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Previous evaluation in the field with the debugging image.

### DIFF
--- a/collector/lib/ProcessSignalFormatter.h
+++ b/collector/lib/ProcessSignalFormatter.h
@@ -3,6 +3,8 @@
 
 #include <memory>
 
+#include <gtest/gtest_prod.h>
+
 #include "api/v1/signal.pb.h"
 #include "internalapi/sensor/signal_iservice.pb.h"
 #include "storage/process_indicator.pb.h"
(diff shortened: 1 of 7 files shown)


collector/lib/ProcessSignalFormatter.h
```diff
@@ -52,6 +56,8 @@ class ProcessSignalFormatter : public ProtoSignalFormatter<sensor::SignalStreamM
   const EventNames& event_names_;
   std::unique_ptr<system_inspector::EventExtractor> event_extractor_;
   ContainerMetadata container_metadata_;
+
+  const CollectorConfig* config_;
```
Molter73, line 60: Could we store this as a reference to make it clear it can never be null and it won't be reassigned? (at least for now).
```suggestion
  const CollectorConfig& config_;
```

collector/lib/ProcessSignalHandler.h
```diff
@@ -34,6 +42,8 @@ class ProcessSignalHandler : public SignalHandler {
   ProcessSignalFormatter formatter_;
   system_inspector::Stats* stats_;
   RateLimitCache rate_limiter_;
+
+  const CollectorConfig* config_;
```
Molter73, line 46: Same question here.

collector/test/ProcessSignalFormatterTest.cpp
```diff
@@ -12,14 +12,34 @@ namespace collector {
...
+};
+
+class MockProcessSignalFormatter : public ProcessSignalFormatter {
+ public:
+  MockProcessSignalFormatter(sinsp* inspector, const CollectorConfig& config) : ProcessSignalFormatter(inspector, config) {};
+
+  ProcessSignal* MockCreateProcessSignal(sinsp_evt* event) {
+    return CreateProcessSignal(event);
+  }
+};
```
Molter73, line 33: Could we use `FRIEND_TEST` definitions here instead of mocks? Ideally, we would have an easy way to create a config object with any configuration we want without any of these methods, but for now, I think I prefer not having mocks when all we are using them for is access to private/protected fields, WDYT?

http://google.github.io/googletest/reference/testing.html#FRIEND_TEST

Molter73 (approved):
LGTM! Just a couple of minor comments, but they are nothing major, so feel free to ignore them.

PR #1899, opened 2024-10-18
Jv integration tests for runtime config

JoukoVirtanen:
## Description

Integration tests for runtime configuration. The tests create various versions of the runtime configuration file and check that the configuration introspection endpoint returns the correct expected configuration.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [x] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Ran the integration tests locally 2,000 times without a failure using the following script
```
#!/usr/bin/env bash
set -eou pipefail

for ((i = 0; i < 2000; i=i+1)); do
        echo "i= $i"
        sudo -E make -C integration-tests TestRuntimeConfigFile
done
```

### DIFF
--- a/collector/lib/ConfigLoader.cpp
+++ b/collector/lib/ConfigLoader.cpp
@@ -55,12 +55,13 @@ bool ConfigLoader::LoadConfiguration(CollectorConfig& config) {
 bool ConfigLoader::LoadConfiguration(CollectorConfig& config, const YAML::Node& node) {
   const auto& config_file = CONFIG_FILE.value();
 
-  if (node.IsNull() || !node.IsDefined()) {
+  if (node.IsNull() || !node.IsDefined() || !node.IsMap()) {
     CLOG(ERROR) << "Unable to read config from " << config_file;
     return false;
   }
+
   YAML::Node networking_node = node["networking"];
-  if (!networking_node) {
+  if (!networking_node || networking_node.IsNull()) {
     CLOG(DEBUG) << "No networking in " << config_file;
     return true;
   }
--- a/collector/test/ConfigLoaderTest.cpp
+++ b/collector/test/ConfigLoaderTest.cpp
@@ -66,14 +66,22 @@ TEST(CollectorConfigTest, TestYamlConfigToConfigInvalid) {
   }
 }
 
-TEST(CollectorConfigTest, TestYamlConfigToConfigEmpty) {
-  YAML::Node yamlNode = YAML::Load("");
-  CollectorConfig config;
-  ASSERT_FALSE(ConfigLoader::LoadConfiguration(config, yamlNode));
+TEST(CollectorConfigTest, TestYamlConfigToConfigEmptyOrMalformed) {
+  std::vector<std::string> tests = {
+      R"(
+                  asdf
+               )",
+      R"()"};
 
-  auto runtime_config = config.GetRuntimeConfig();
+  for (const auto& yamlStr : tests) {
+    YAML::Node yamlNode = YAML::Load(yamlStr);
+    CollectorConfig config;
+    ASSERT_FALSE(ConfigLoader::LoadConfiguration(config, yamlNode));
 
-  EXPECT_FALSE(runtime_config.has_value());
+    auto runtime_config = config.GetRuntimeConfig();
+
+    EXPECT_FALSE(runtime_config.has_value());
+  }
 }
 
 TEST(CollectorConfigTest, TestPerContainerRateLimit) {
--- a/integration-tests/pkg/assert/assert.go
+++ b/integration-tests/pkg/assert/assert.go
@@ -1,7 +1,6 @@
 package assert
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"strings"
@@ -60,7 +59,6 @@ func AssertRepeated(t *testing.T, tickTime time.Duration, timeout time.Duration,
 	}
 }
 
->>>>>>> 66ca3aa0 (X-Smart-Squash: Squashed 41 commits:)
 func ElementsMatchFunc[N any](expected []N, actual []N, equal func(a, b N) bool) bool {
 	if len(expected) != len(actual) {
 		return false
--- a/integration-tests/pkg/assert/assert.go
+++ b/integration-tests/pkg/assert/assert.go
@@ -1,13 +1,66 @@
 package assert
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/davecgh/go-spew/spew"
 	"github.com/stretchr/testify/assert"
+
+	"github.com/stackrox/collector/integration-tests/pkg/collector"
+	"github.com/stackrox/collector/integration-tests/pkg/log"
+	"github.com/stackrox/collector/integration-tests/pkg/types"
 )
 
+func AssertExternalIps(t *testing.T, enable bool, collectorIP string) {
+	tickTime := 1 * time.Second
+	timeout := 3 * time.Minute
+	AssertRepeated(t, tickTime, timeout, func() bool {
+		body, err := collector.IntrospectionQuery(collectorIP, "/state/config")
+		assert.NoError(t, err)
+		var response types.RuntimeConfig
+		err = json.Unmarshal(body, &response)
+		assert.NoError(t, err)
+
+		return response.Networking.ExternalIps.Enable == enable
+	})
+}
+
+func AssertNoRuntimeConfig(t *testing.T, collectorIP string) {
+	tickTime := 1 * time.Second
+	timeout := 3 * time.Minute
+	AssertRepeated(t, tickTime, timeout, func() bool {
+		body, err := collector.IntrospectionQuery(collectorIP, "/state/config")
+		assert.NoError(t, err)
+		return strings.TrimSpace(string(body)) == "{}"
+	})
+}
+
+func AssertRepeated(t *testing.T, tickTime time.Duration, timeout time.Duration, condition func() bool) {
+	tick := time.NewTicker(tickTime)
+	timer := time.After(timeout)
+
+	for {
+		select {
+		case <-tick.C:
+			if condition() {
+				// Condition has been met
+				return
+			}
+
+		case <-timer:
+			// TODO: This message should be passed in rather than hard coded here
+			log.Error("Timeout reached: Runtime configuration was not updated")
+			t.FailNow()
+		}
+	}
+}
+
+>>>>>>> 66ca3aa0 (X-Smart-Squash: Squashed 41 commits:)
 func ElementsMatchFunc[N any](expected []N, actual []N, equal func(a, b N) bool) bool {
 	if len(expected) != len(actual) {
 		return false
--- a/integration-tests/pkg/assert/assert.go
+++ b/integration-tests/pkg/assert/assert.go
@@ -19,7 +19,7 @@ var (
 	runtimeConfigErrorMsg = "Runtime configuration was not updated"
 )
 
-func AssertExternalIps(t *testing.T, enable bool, collectorIP string) {
+func AssertExternalIps(t *testing.T, enabled bool, collectorIP string) {
 	tickTime := 1 * time.Second
 	timeout := 3 * time.Minute
 	AssertRepeated(t, tickTime, timeout, runtimeConfigErrorMsg, func() bool {
@@ -29,7 +29,7 @@ func AssertExternalIps(t *testing.T, enable bool, collectorIP string) {
 		err = json.Unmarshal(body, &response)
 		assert.NoError(t, err)
 
-		return response.Networking.ExternalIps.Enable == enable
+		return response.Networking.ExternalIps.Enable == enabled
 	})
 }
 
--- a/integration-tests/pkg/assert/assert.go
+++ b/integration-tests/pkg/assert/assert.go
@@ -15,10 +15,14 @@ import (
 	"github.com/stackrox/collector/integration-tests/pkg/types"
 )
 
+var (
+	runtimeConfigErrorMsg = "Runtime configuration was not updated"
+)
+
 func AssertExternalIps(t *testing.T, enable bool, collectorIP string) {
 	tickTime := 1 * time.Second
 	timeout := 3 * time.Minute
-	AssertRepeated(t, tickTime, timeout, func() bool {
+	AssertRepeated(t, tickTime, timeout, runtimeConfigErrorMsg, func() bool {
 		body, err := collector.IntrospectionQuery(collectorIP, "/state/config")
 		assert.NoError(t, err)
 		var response types.RuntimeConfig
@@ -32,14 +36,14 @@ func AssertExternalIps(t *testing.T, enable bool, collectorIP string) {
 func AssertNoRuntimeConfig(t *testing.T, collectorIP string) {
 	tickTime := 1 * time.Second
 	timeout := 3 * time.Minute
-	AssertRepeated(t, tickTime, timeout, func() bool {
+	AssertRepeated(t, tickTime, timeout, runtimeConfigErrorMsg, func() bool {
 		body, err := collector.IntrospectionQuery(collectorIP, "/state/config")
 		assert.NoError(t, err)
 		return strings.TrimSpace(string(body)) == "{}"
 	})
 }
 
-func AssertRepeated(t *testing.T, tickTime time.Duration, timeout time.Duration, condition func() bool) {
+func AssertRepeated(t *testing.T, tickTime time.Duration, timeout time.Duration, msg string, condition func() bool) {
 	tick := time.NewTicker(tickTime)
 	timer := time.After(timeout)
 
@@ -52,8 +56,7 @@ func AssertRepeated(t *testing.T, tickTime time.Duration, timeout time.Duration,
 			}
 
 		case <-timer:
-			// TODO: This message should be passed in rather than hard coded here
-			log.Error("Timeout reached: Runtime configuration was not updated")
+			log.Error("Timeout reached: " + msg)
 			t.FailNow()
 		}
 	}
--- a/integration-tests/pkg/collector/collector_docker.go
+++ b/integration-tests/pkg/collector/collector_docker.go
@@ -41,6 +41,7 @@ func NewDockerCollectorManager(e executor.Executor, name string) *DockerCollecto
 		"/host/etc:ro":              "/etc",
 		"/host/usr/lib:ro":          "/usr/lib",
 		"/host/sys/kernel/debug:ro": "/sys/kernel/debug",
+		"/etc/stackrox:ro":          "/tmp/collector-test",
 	}
 
 	return &DockerCollectorManager{
--- a/integration-tests/pkg/collector/introspection.go
+++ b/integration-tests/pkg/collector/introspection.go
@@ -8,8 +8,8 @@ import (
 	"github.com/stackrox/collector/integration-tests/pkg/log"
 )
 
-func (k *K8sCollectorManager) IntrospectionQuery(endpoint string) ([]byte, error) {
-	uri := fmt.Sprintf("http://%s:8080%s", k.IP(), endpoint)
+func IntrospectionQuery(collectorIP string, endpoint string) ([]byte, error) {
+	uri := fmt.Sprintf("http://%s:8080%s", collectorIP, endpoint)
 	resp, err := http.Get(uri)
 	if err != nil {
 		return nil, err
--- a/integration-tests/pkg/mock_sensor/expect_conn.go
+++ b/integration-tests/pkg/mock_sensor/expect_conn.go
@@ -90,7 +90,7 @@ func (s *MockSensor) ExpectSameElementsConnections(t *testing.T, containerID str
 	}
 
 	connections := s.Connections(containerID)
-	if collectorAssert.ElementsMatchFunc(connections, expected, equal) {
+	if collectorAssert.ElementsMatchFunc(expected, connections, equal) {
 		return true
 	}
 
@@ -100,13 +100,13 @@ func (s *MockSensor) ExpectSameElementsConnections(t *testing.T, containerID str
 		select {
 		case <-timer:
 			connections := s.Connections(containerID)
-			return collectorAssert.AssertElementsMatchFunc(t, connections, expected, equal)
+			return collectorAssert.AssertElementsMatchFunc(t, expected, connections, equal)
 		case conn := <-s.LiveConnections():
 			if conn.GetContainerId() != containerID {
 				continue
 			}
 			connections := s.Connections(containerID)
-			if collectorAssert.ElementsMatchFunc(connections, expected, equal) {
+			if collectorAssert.ElementsMatchFunc(expected, connections, equal) {
 				return true
 			}
 		}
--- a/integration-tests/pkg/mock_sensor/server.go
+++ b/integration-tests/pkg/mock_sensor/server.go
@@ -432,8 +432,8 @@ func (m *MockSensor) pushConnection(containerID string, connection *sensorAPI.Ne
 		CloseTimestamp: connection.GetCloseTimestamp().String(),
 	}
 
-	if _, ok := m.connections[containerID]; ok {
-		m.connections[containerID] = append(m.connections[containerID], conn)
+	if c, ok := m.connections[containerID]; ok {
+		m.connections[containerID] = append(c, conn)
 	} else {
 		m.connections[containerID] = []types.NetworkInfo{conn}
 	}
--- a/integration-tests/pkg/types/runtime_config.go
+++ b/integration-tests/pkg/types/runtime_config.go
@@ -16,7 +16,7 @@ func (n *RuntimeConfig) Equal(other RuntimeConfig) bool {
 	return n.Networking.ExternalIps.Enable == other.Networking.ExternalIps.Enable
 }
 
-func (s *RuntimeConfig) GetRuntimeConfigStr() (string, error) {
+func (n *RuntimeConfig) GetRuntimeConfigStr() (string, error) {
 	yamlBytes, err := yaml.Marshal(s)
 
 	if err != nil {
--- a/integration-tests/pkg/types/runtime_config.go
+++ b/integration-tests/pkg/types/runtime_config.go
@@ -1,5 +1,9 @@
 package types
 
+import (
+	"gopkg.in/yaml.v3"
+)
+
 type RuntimeConfig struct {
 	Networking struct {
 		ExternalIps struct {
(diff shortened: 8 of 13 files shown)


collector/lib/ConfigLoader.cpp
```diff
@@ -47,6 +47,10 @@ bool ConfigLoader::LoadConfiguration(CollectorConfig& config) {
   } catch (const YAML::ParserException& e) {
     CLOG(ERROR) << "Failed to parse the configuration file: " << config_file << ". Error: " << e.what();
     return false;
+  } catch (const YAML::Exception& e) {
+    CLOG(ERROR) << "Failed to parse the configuration file: " << config_file << ". Error: " << e.what();
+  } catch (const std::exception& e) {
+    CLOG(ERROR) << "An unknown error occured while loading the configuration file: " << config_file << ". Error: " << e.what();
```
JoukoVirtanen, line 53: Remove

collector/lib/ConfigLoader.cpp
```diff
@@ -55,12 +59,13 @@ bool ConfigLoader::LoadConfiguration(CollectorConfig& config) {
...
 
-  if (node.IsNull() || !node.IsDefined()) {
+  if (node.IsNull() || !node.IsDefined() || !node.IsMap()) {
     CLOG(ERROR) << "Unable to read config from " << config_file;
     return false;
   }
+
   YAML::Node networking = node["networking"];
-  if (!networking) {
+  if (!node["networking"] || node["networking"].IsNull()) {
```
Molter73, line 68:
```suggestion
  if (!networking || networking.IsNull()) {
```

collector/test/ConfigLoaderTest.cpp
```diff
@@ -67,12 +67,20 @@ TEST(CollectorConfigTest, TestYamlConfigToConfigInvalid) {
 }
 
 TEST(CollectorConfigTest, TestYamlConfigToConfigEmpty) {
```
Molter73, line 69: Please update this test name.

integration-tests/pkg/collector/collector_docker.go
```diff
@@ -41,6 +41,7 @@ func NewDockerCollectorManager(e executor.Executor, name string) *DockerCollecto
 		"/host/etc:ro":              "/etc",
 		"/host/usr/lib:ro":          "/usr/lib",
 		"/host/sys/kernel/debug:ro": "/sys/kernel/debug",
+		"/etc/stackrox:ro":          "/tmp",
```
Molter73, line 44: Consider creating a separate directory for this like `/tmp/collector-test` and use that instead of the full /tmp directory.

integration-tests/pkg/collector/introspection.go
```diff
@@ -23,3 +30,50 @@ func (k *K8sCollectorManager) IntrospectionQuery(endpoint string) ([]byte, error
...
+				return
+			}
+
+		case <-timer:
+			// TODO: This message should be passed in rather than hard coded here
+			t.Log("Timeout reached: Runtime configuration was not updated")
+			t.FailNow()
+		}
+	}
+}
```
Molter73, line 79: We probably want to move all assertions to a separate `pkg/assert` or similar.

integration-tests/pkg/collector/introspection.go
```diff
@@ -23,3 +30,50 @@ func (k *K8sCollectorManager) IntrospectionQuery(endpoint string) ([]byte, error
...
 	return body, err
 }
+
+func QueryConfig(t *testing.T, collectorIP string) []byte {
+	log.Info("Querying: /state/config")
+	body, err := IntrospectionQuery(collectorIP, "/state/config")
+	assert.NoError(t, err)
+	log.Info("Response: %q", body)
+	return body
+}
```
Molter73, line 40: I'm not 100% sold this is something we need, from a test we could just do:
```go
body, err := IntrospectionQuery(s.Collector().IP(), "/state/config")
s.Require().NoError(err)
```
This is simple enough and you can add the logging if you deem it necessary.

Molter73 (commented):
I'm less than half way through, I'll finish the review tomorrow.

collector/lib/ConfigLoader.cpp
JoukoVirtanen: Removed

collector/test/ConfigLoaderTest.cpp
JoukoVirtanen: Done

integration-tests/pkg/collector/collector_docker.go
JoukoVirtanen: Done

integration-tests/pkg/collector/introspection.go
JoukoVirtanen: Done

integration-tests/pkg/collector/introspection.go
JoukoVirtanen: Done

integration-tests/pkg/mock_sensor/expect_conn.go
```diff
@@ -33,7 +33,7 @@ loop:
 		case <-timer:
 			// we know they don't match at this point, but by using
 			// ElementsMatch we get much better logging about the differences
-			return assert.ElementsMatch(t, expected, s.Connections(containerID), "timed out waiting for networks")
+			return assert.ElementsMatch(t, expected, s.Connections(containerID), "timed out waiting for network connections")
```
JoukoVirtanen, line 36: The changes in this file were made here https://github.com/stackrox/collector/pull/1902/files

integration-tests/pkg/mock_sensor/server.go
```diff
@@ -32,7 +32,6 @@ const (
 // us to use any comparable type as the key)
 type ProcessMap map[types.ProcessInfo]interface{}
 type LineageMap map[types.ProcessLineage]interface{}
-type ConnMap map[types.NetworkInfo]interface{}
```
JoukoVirtanen, line 35: The changes in this file were made here https://github.com/stackrox/collector/pull/1902/files

integration-tests/suites/runtime_config_file.go
```diff
@@ -0,0 +1,172 @@
...
+	s.execShellCommand(cmd)
+}
+
+func (s *RuntimeConfigFileTestSuite) setRuntimeConfig(runtimeConfigFile string, configStr string) {
+	cmd := "echo '" + configStr + "' > " + runtimeConfigFile
+	s.execShellCommand(cmd)
+}
+
+func (s *RuntimeConfigFileTestSuite) setExternalIpsEnable(runtimeConfigFile string, enable bool) {
+	configStr := "networking:\n  externalIps:\n    enable: " + strconv.FormatBool(enable)
```
Molter73, line 36:
```suggestion
	configStr := `networking:
  externalIps:
    enable: ` + strconv.FormatBool(enable)
```

Or even better, use the `RuntimeConfig` type defined in `integration-tests/pkg/types/runtime_config.go` and serialize it to yaml.

integration-tests/suites/runtime_config_file.go
```diff
@@ -0,0 +1,172 @@
...
+	externalIp := "8.8.8.8"
+	serverPort := 53
+	externalUrl := fmt.Sprintf("http://%s:%d", externalIp, serverPort)
+	image_store := config.Images()
+	containerID, err := s.Executor().StartContainer(
+		config.ContainerStartConfig{
+			Name:    "external-connection",
+			Image:   image_store.QaImageByKey("qa-alpine-curl"),
+			Command: []string{"sh", "-c", "while true; do curl " + externalUrl + "; sleep 1; done"},
+		})
```
Molter73, line 59: [nit]
```suggestion
	containerID, err := s.Executor().StartContainer(
		config.ContainerStartConfig{
			Name:    "external-connection",
			Image:   config.Images().QaImageByKey("qa-alpine-curl"),
			Command: []string{"sh", "-c", "while true; do curl " + externalUrl + "; sleep 1; done"},
		})
```

integration-tests/suites/runtime_config_file.go
```diff
@@ -0,0 +1,172 @@
...
+	// expect a normalized connection
+	expectedConnections := []types.NetworkInfo{activeNormalizedConnection}
+	connectionSuccess := s.Sensor().ExpectSameElementsConnections(s.T(), clientContainer, 10*time.Second, expectedConnections...)
+	s.Require().True(connectionSuccess)
+
+	s.setExternalIpsEnable(runtimeConfigFile, true)
+	assert.AssertExternalIps(s.T(), true, collectorIP)
+	// When external IPs is enabled the normalized connection is reported as inactive and the unnormalized connection
+	// is reported as active
+	expectedConnections = append(expectedConnections, activeUnnormalizedConnection, inactiveNormalizedConnection)
```
Molter73, line 112: Shouldn't we be cleaning `expectedConnections` between each test? Isn't this just saying "you should see the previous ones and these ones"? How does that work when we turn external IPs back off?

integration-tests/suites/runtime_config_file.go
```diff
@@ -0,0 +1,172 @@
...
+	connectionSuccess = s.Sensor().ExpectSameElementsConnections(s.T(), clientContainer, 10*time.Second, expectedConnections...)
+	s.Require().True(connectionSuccess)
+
+	invalidConfig := "asdf"
+	s.setRuntimeConfig(runtimeConfigFile, invalidConfig)
+	assert.AssertExternalIps(s.T(), false, collectorIP)
+	// Testing an invalid configuration. There should not be a change in the configuration or reported connections
+	common.Sleep(3 * time.Second) // Sleep so that collector has a chance to report connections
+	connectionSuccess = s.Sensor().ExpectSameElementsConnections(s.T(), clientContainer, 10*time.Second, expectedConnections...)
+	s.Require().True(connectionSuccess)
```
Molter73, line 162: Couldn't this be split into multiple test cases? Reviewing this test is borderline impossible.

integration-tests/pkg/assert/assert.go
```diff
@@ -0,0 +1,53 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/stackrox/collector/integration-tests/pkg/types"
+)
+
+func AssertExternalIps(t *testing.T, enable bool, collectorIP string) {
+	AssertRepeated(t, func() bool {
+		body, err := collector.IntrospectionQuery(collectorIP, "/state/config")
```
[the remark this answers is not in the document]
JoukoVirtanen: I think it can be done in another PR. Maybe we should have an `Introspection` struct that could be a member of a `Collector` interface. The introspection object would get the IP address from the collector object.

integration-tests/suites/runtime_config_file.go
JoukoVirtanen: I don't think we should clean `expectConnections` the mock server does not clear out the connections that it sees after asserts, so `expectConnections` should not be cleared out either.

> Isn't this just saying "you should see the previous ones and these ones"? 
Yes

 > How does that work when we turn external IPs back off?
 When external IPs is turned back off the connection with the unnormalized connection is reported closed and the normalized connection is reported open. We get two reported connection events in that case.

integration-tests/suites/runtime_config_file.go
JoukoVirtanen: It could be, but I think this is better. This tests for case when external IPs is enabled after the config file is deleted as well as the case when external IPs is disabled after the config file is deleted. It tests for deleting the config file when external IPs is enabled and it tests for deleting the config file with external IPs is disabled. These cases could be split into different tests, but it would take longer to run. Would better comments make it easier to understand what is going on in the test?

Molter73: > Couldn't this be split into multiple test cases? Reviewing this test is borderline impossible.

It's not a problem if they take a couple more seconds, the problem is that you have 8, highly condensed, nearly identical blocks of code. This makes it really hard to understand what is actually changing from one block to the next, because I can't see what the difference is. This, in turn, makes it really hard for me to give meaningful feedback on the test, because I can't understand it.

The intent of the test should be obvious from a glance at the code.

> Would better comments make it easier to understand what is going on in the test?

If anything, the existing comments are making the code even denser and harder to understand. Try doing shorter comments at the start of each block, don't go into deep details of what the block is doing, just give a hint of what it'll do and try to read the block of code from the perspective of a person that is going through it for the first time, ask yourself:
- Is it easy to see what has changed in this part of the test?
- Can I easily spot the intent of the block?
- Is it distinct enough from a previous, similar block?

JoukoVirtanen: I have split the test into three tests.

Molter73 (commented):
Thanks for splitting the tests up, as silly as it may sound, it's a lot easier to understand what is going on now.

The comments I left now are mostly on moving some comments around to leave the code blocks on their own. This should help reduce clutter a bit more, since the description of what will be done is provided up front and the code is left as clean as possible.

JoukoVirtanen:
CI results

Run 1
Many Konflux integration tests failed. rhcos failed with 
```
=== RUN   TestProcfsScraperDisableFeatureFlag/TestProcfsScraper
    expect_conn.go:181: 
        	Error Trace:	/tests/pkg/mock_sensor/expect_conn.go:181
        	            				/tests/suites/procfs_scraper.go:72
        	Error:      	elements differ
        	            	
        	            	extra elements in list A:
        	            	([]interface ***) (len=1) ***
        	            	 (types.EndpointInfo) ***
        	            	  Protocol: (string) (len=15) "L4_PROTOCOL_TCP",
        	            	  Address: (types.ListenAddress) ***
        	            	   AddressData: (string) (len=4) "\x00\x00\x00\x00",
        	            	   Port: (int) 80,
        	            	   IpNetwork: (string) (len=5) "\x00\x00\x00\x00 "
        	            	  ***,
        	            	  CloseTimestamp: (string) (len=5) "<nil>",
        	            	  Originator: (types.ProcessOriginator) ***
        	            	   ProcessName: (string) "",
        	            	   ProcessExecFilePath: (string) "",
        	            	   ProcessArgs: (string) ""
        	            	  ***
        	            	 ***
        	            	***
        	            	
        	            	
        	            	listA:
        	            	([]types.EndpointInfo) (len=1) ***
        	            	 (types.EndpointInfo) ***
        	            	  Protocol: (string) (len=15) "L4_PROTOCOL_TCP",
        	            	  Address: (types.ListenAddress) ***
        	            	   AddressData: (string) (len=4) "\x00\x00\x00\x00",
        	            	   Port: (int) 80,
        	            	   IpNetwork: (string) (len=5) "\x00\x00\x00\x00 "
        	            	  ***,
        	            	  CloseTimestamp: (string) (len=5) "<nil>",
        	            	  Originator: (types.ProcessOriginator) ***
        	            	   ProcessName: (string) "",
        	            	   ProcessExecFilePath: (string) "",
        	            	   ProcessArgs: (string) ""
        	            	  ***
        	            	 ***
        	            	***
        	            	
        	            	
        	            	listB:
        	            	([]types.EndpointInfo) ***
        	            	***
        	Test:       	TestProcfsScraperDisableFeatureFlag/TestProcfsScraper
        	Messages:   	timed out waiting for endpoints
```
The other konflux integration tests failed with either problems pulling the image. Both konflux and non-konflux cos integration tests failed with verifier errors.

integration-tests/suites/runtime_config_file.go
```diff
@@ -0,0 +1,197 @@
...
+	s.cleanupContainers("external-connection")
+	s.WritePerfResults()
+	s.deleteFile(runtimeConfigFile)
+}
+
+func (s *RuntimeConfigFileTestSuite) TestRuntimeConfigFileEnable() {
+	// The runtime config file was deleted before starting collector so there should not be any config
+	assert.AssertNoRuntimeConfig(s.T(), collectorIP)
+	// Since there is no config the default is used, which means external IPs is disabled and we should
+	// expect a normalized connection
```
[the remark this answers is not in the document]
Molter73: This was not applied to all tests.

Molter73 (commented):
Changes look ok, but #1902 still has comments on it that haven't been addressed, if your intent is to merge the changes from that PR with this one, please close it so I can move my comments to this PR.

integration-tests/pkg/assert/assert.go
```diff
@@ -0,0 +1,53 @@
...
+		assert.NoError(t, err)
+		return strings.TrimSpace(string(body)) == "{}"
+	})
+}
+
+func AssertRepeated(t *testing.T, condition func() bool) {
+	tick := time.NewTicker(1 * time.Second)
+	timer := time.After(3 * time.Minute)
```
Stringy, line 37: Can these timeouts be configurable via some parameters?

JoukoVirtanen: Done

JoukoVirtanen:
> Changes look ok, but #1902 still has comments on it that haven't been addressed, if your intent is to merge the changes from that PR with this one, please close it so I can move my comments to this PR.

I have addressed the PR review comments in the other PR.

integration-tests/suites/runtime_config_file.go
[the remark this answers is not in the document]
JoukoVirtanen: It has now been applied to all tests.

integration-tests/pkg/assert/assert.go
```diff
@@ -0,0 +1,93 @@
...
+		select {
+		case <-tick.C:
+			if condition() {
+				// Condition has been met
+				return
+			}
+
+		case <-timer:
+			// TODO: This message should be passed in rather than hard coded here
+			t.Log("Timeout reached: Runtime configuration was not updated")
```
Molter73, line 56: We probably want to use our log package for consistency
```suggestion
			log.Error("Timeout reached: Runtime configuration was not updated")
```

integration-tests/pkg/mock_sensor/server.go
```diff
@@ -432,11 +432,10 @@ func (m *MockSensor) pushConnection(containerID string, connection *sensorAPI.Ne
 		CloseTimestamp: connection.GetCloseTimestamp().String(),
 	}
 
-	if connections, ok := m.connections[containerID]; ok {
-		connections[conn] = true
+	if _, ok := m.connections[containerID]; ok {
+		m.connections[containerID] = append(m.connections[containerID], conn)
```
Molter73, line 436: [nit] Not 100% sure this will work.
```suggestion
	if c, ok := m.connections[containerID]; ok {
		m.connections[containerID] = append(c, conn)
```

Molter73 (commented):
LGTM, but holding approval until #1902 is merged, so this can be rebased on top of that.

integration-tests/pkg/assert/assert.go
```diff
@@ -1,13 +1,64 @@
...
+		select {
+		case <-tick.C:
+			if condition() {
+				// Condition has been met
+				return
+			}
+
+		case <-timer:
+			// TODO: This message should be passed in rather than hard coded here
+			log.Error("Timeout reached: Runtime configuration was not updated")
```
Molter73, line 56: This `TODO` should be relatively simple to solve, add a format and args parameter to the function like this:
```go
func AssertRepeated(t *testing.T, tickTime time.Duration, timeout time.Duration, condition func() bool, msg string, args ...interface{}) {
```
Then here you can just call `log.Error` like this:
```go
log.Error("Timeout reached: " + msg, args...)
```

If this ends up being too verbose on the call sites with the function being defined inline, you can always define the function and assign it to a variable, then pass that in.

Molter73 (approved):
LGTM!

Just a small comment on a TODO that I think we could solve quickly before merging, but we can also address it on the next PR that needs to use the helper function.

integration-tests/pkg/assert/assert.go
```diff
@@ -1,13 +1,67 @@
...
+	"github.com/stackrox/collector/integration-tests/pkg/collector"
+	"github.com/stackrox/collector/integration-tests/pkg/log"
+	"github.com/stackrox/collector/integration-tests/pkg/types"
+)
+
+var (
+	runtimeConfigErrorMsg = "Runtime configuration was not updated"
 )
 
+func AssertExternalIps(t *testing.T, enable bool, collectorIP string) {
```
Stringy, line 22: [nit] should probably be `enabled` rather than `enable` (the latter implies we're turning it on)

integration-tests/pkg/types/runtime_config.go
```diff
@@ -0,0 +1,13 @@
+package types
+
+type RuntimeConfig struct {
+	Networking struct {
+		ExternalIps struct {
+			Enable bool
+		}
+	}
```
Stringy, line 8: just a note that we'll need to add `PerContainerRateLimit` (probably in a follow up PR)

integration-tests/suites/runtime_config_file.go
```diff
@@ -0,0 +1,195 @@
...
+)
+
+type RuntimeConfigFileTestSuite struct {
+	IntegrationTestSuiteBase
+	ClientContainer string
+}
+
+func (s *RuntimeConfigFileTestSuite) createDirectory(dir string) {
+	cmd := "mkdir " + dir
+	s.execShellCommand(cmd)
```
Stringy, line 66: Can't we do this via the standard library? same for deleteFile, setRuntimeConfig

integration-tests/suites/runtime_config_file.go
```diff
@@ -0,0 +1,195 @@
...
+func (s *RuntimeConfigFileTestSuite) setRuntimeConfig(runtimeConfigFile string, configStr string) {
+	cmd := "echo '" + configStr + "' > " + runtimeConfigFile
+	s.execShellCommand(cmd)
+}
+
+func (s *RuntimeConfigFileTestSuite) setExternalIpsEnable(runtimeConfigFile string, enable bool) {
+	configStr := `networking:
+  externalIps:
+    enable: ` + strconv.FormatBool(enable)
+	s.setRuntimeConfig(runtimeConfigFile, configStr)
```
Stringy, line 83: It would be better for to construct the object and serialize it to a string, and in fact it may be best to have the caller pass the structure directly to setRuntimeConfig

```
func (s *RuntimeConfigFileTestSuite) setRuntimeConfig(path string, config types.RuntimeConfig) error
```

integration-tests/pkg/assert/assert.go
JoukoVirtanen: Done

integration-tests/pkg/types/runtime_config.go
JoukoVirtanen: We should not be defining a RuntimeConfig struct here. We should be importing the go code generated from the protobuf definitions. That should be done in a different PR.

integration-tests/suites/runtime_config_file.go
JoukoVirtanen: This way we get the logging that comes with `execShellCommand`. Also if there are changes to how commands need to be executed, using the standard library might not be the best method. E.g we might want to run the commands in a VM, locally, in a container, or somewhere else and changes to those things might result in the standard library not working. However, you did remind me that this is not the correct place for these functions and I have moved them to `base.go`.

integration-tests/suites/runtime_config_file.go
JoukoVirtanen: Done

integration-tests/suites/runtime_config_file.go
```diff
@@ -0,0 +1,189 @@
...
+	IntegrationTestSuiteBase
+	ClientContainer string
+}
+
+func (s *RuntimeConfigFileTestSuite) setRuntimeConfig(runtimeConfigFile string, configStr string) {
+	cmd := "echo '" + configStr + "' > " + runtimeConfigFile
+	s.execShellCommand(cmd)
+}
+
+func (s *RuntimeConfigFileTestSuite) setExternalIpsEnable(runtimeConfigFile string, enable bool) {
```
Stringy, line 69: This might sound like a nit, but from a separation of concerns perspective, this function shouldn't do the marshalling - it should be handled by `setRuntimeConfig` 

This function should just construct `types.RuntimeConfig` object and pass it to `setRuntimeConfig`. Otherwise, when we inevitably come to adding another field to these tests, we'll have to duplicate the marshaling logic.

integration-tests/suites/runtime_config_file.go
Stringy: We've gone to great lengths to remove the various contexts in which things run, and today we're always running on the same host with the test containers so I don't think we should care about how commands run, or whether to restrict ourselves to using the shell instead of the stdlib. If this ever changes, then we'd have to overhaul the way the tests work again, so a few file system operations would be the least of our worries.

If these specific operations need to run in various contexts, then `execShellCommand` is still inappropriate because it is intended for running a local shell command (via `exec.Command`) and we'd have to change these functions anyway. 

It's just that a shell command feels very expensive for an operation as simple as making a directory or deleting a file, and if we need logging then we can add logging here.

Stringy: Also, we would have a better handle on the errors if we do it via the stdlib. We wouldn't need to worry about failures in exec'ing the shell or the command, for example.

we're also just suppressing all errors at the moment -- we should probably return them to the caller

JoukoVirtanen: Done

integration-tests/suites/runtime_config_file.go
JoukoVirtanen: Done

integration-tests/pkg/types/runtime_config.go
```diff
@@ -0,0 +1,27 @@
...
+			Enable bool `yaml:"enable"`
+		} `yaml:"externalIps"`
+	} `yaml:"networking"`
+}
+
+func (n *RuntimeConfig) Equal(other RuntimeConfig) bool {
+	return n.Networking.ExternalIps.Enable == other.Networking.ExternalIps.Enable
+}
+
+func (s *RuntimeConfig) GetRuntimeConfigStr() (string, error) {
```
JoukoVirtanen, line 19:
```suggestion
func (n *RuntimeConfig) GetRuntimeConfigStr() (string, error) {
```

integration-tests/suites/runtime_config_file.go
```diff
@@ -0,0 +1,192 @@
...
+)
+
+type RuntimeConfigFileTestSuite struct {
+	IntegrationTestSuiteBase
+	ClientContainer string
+}
+
+func (s *RuntimeConfigFileTestSuite) setRuntimeConfig(runtimeConfigFile string, configStr string) {
+	cmd := "echo '" + configStr + "' > " + runtimeConfigFile
+	s.execShellCommand(cmd)
```
JoukoVirtanen, line 65: Use os.write.

Stringy: https://pkg.go.dev/os#example-WriteFile

Stringy (approved):
As discussed, there's a last couple of minor changes but otherwise LGTM!

PR #1900, opened 2024-10-21
Wait for a container to be running

erthalion:
## Description

TestProcessListeningOnPort has became flaky in a nasty way, leaving no logs
from the flask container. Most likely reason is that it takes some time to
start the container, while the client method ContainerCreate doesn't wait for
"running" status.

Make the test more robust by waiting until it reports running status. It's been
done using ContainerExecInspect, not ContainerWait, because the latter one
isn't suitable despite the name. ContainerWait [[1]] could only wait for a
non-running state, and is designed to wait until a container has finished it's
job.

[1]: https://pkg.go.dev/github.com/docker/docker/client#Client.ContainerWait

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Manual testing, running the flaky test.

### DIFF
--- a/integration-tests/pkg/executor/executor_docker_api.go
+++ b/integration-tests/pkg/executor/executor_docker_api.go
@@ -130,6 +130,31 @@ func (d *dockerAPIExecutor) StartContainer(startConfig config.ContainerStartConf
 		return "", errors.Wrapf(err, "start %s", startConfig.Name)
 	}
 
+	RetryWithTimeout(func() (output string, err error) {
+		inspect, err := d.client.ContainerExecInspect(ctx, resp.ID)
+
+		if err != nil {
+			// The client doesn't expose any specific error type in this
+			// case, but "no exec session" is a general error we will see
+			// when a container is not started yet.
+			prefix := "Error response from daemon: no exec session with ID"
+			if strings.HasPrefix(fmt.Sprint(err), prefix) {
+				return NoOutput, err
+			}
+
+			// If it's something outstanding, log it
+			log.Warn("Failed to inspect %s: %+v", startConfig.Name, err)
+			return NoOutput, err
+		}
+
+		if inspect.Running == true {
+			return NoOutput, nil
+		} else {
+			return NoOutput, fmt.Errorf("Container %s is not running",
+				startConfig.Name)
+		}
+	}, fmt.Errorf("Container %s didn't start in time", startConfig.Name))
+
 	log.Info("start %s with %s (%s)\n",
 		startConfig.Name, startConfig.Image, common.ContainerShortID(resp.ID))
 	return resp.ID, nil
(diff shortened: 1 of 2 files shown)


integration-tests/pkg/executor/executor_docker_api.go
```diff
@@ -130,6 +130,28 @@ func (d *dockerAPIExecutor) StartContainer(startConfig config.ContainerStartConf
 		return "", errors.Wrapf(err, "start %s", startConfig.Name)
 	}
 
+	tickSeconds := time.Second
+	timeout := 30 * time.Second
+
+	tick := time.Tick(tickSeconds)
+	timer := time.After(timeout)
+
+	for {
```
Molter73, line 139: We have a `Retry` method that is currently unused, maybe we can move it to the `common` package and use that here?
https://github.com/stackrox/collector/blob/19eafba2eaa3bc48a3d059ca8f190a22fcc25b09/integration-tests/pkg/executor/retry.go#L19

integration-tests/pkg/executor/executor_docker_api.go
```diff
@@ -130,6 +130,28 @@ func (d *dockerAPIExecutor) StartContainer(startConfig config.ContainerStartConf
...
+		select {
+		case <-tick:
+			inspect, err := d.client.ContainerExecInspect(ctx, resp.ID)
```
Molter73, line 142: We might want to directly check if there was an error here:
```suggestion
			inspect, err := d.client.ContainerExecInspect(ctx, resp.ID)
			if err != nil {
				log.Info("Failed to inspect %s: %s", startConfig.Name, err)
				continue
			}
```

Then we can remove it from `err` from the log message on line 148, since it's currently causing some noise:
```
2024/10/21 15:41:38 INFO: Wait for container container-stats to start, %!w(errdefs.errNotFound=***0xc000616a08***)
2024/10/21 15:41:39 INFO: Wait for container container-stats to start, %!w(errdefs.errNotFound=***0xc000616b58***)
2024/10/21 15:41:40 INFO: Wait for container container-stats to start, %!w(errdefs.errNotFound=***0xc000616d08***)
2024/10/21 15:41:41 INFO: Wait for container container-stats to start, %!w(errdefs.errNotFound=***0xc000616fa8***)
```

Molter73 (commented):
Looking at the errors on CI, it seems the proposed changes are locking themselves and timing out on all containers being spin up.

integration-tests/pkg/executor/executor_docker_api.go
Molter73: Actually, just noticed the changes are inside the executor package, so it could be used directly if we wanted to.

erthalion: The original Retry was number based, but in this context time based retry is more appropriate. I've moved the ticker implementation into the retry module.

Molter73 (approved):
LGTM!

Could you edit the PR description a bit before merging? So we can have some context there if we ever need it.

PR #1901, opened 2024-10-21
ROX-26711: Collector mock server should handle external ips

JoukoVirtanen:
## Description

This makes it so that the mock server for collector integration tests can handle external IP addresses and CIDR blocks. Currently the mock server can only handle the `AddressData` field of the `NetworkAddress` struct. However, for external IP addressed and CIDR blocks the `IpNetwork` field is used. This change is needed for testing runtime configuration and the external IPs feature. See https://github.com/stackrox/collector/pull/1899

The function modified here is used to convert objects of type `NetworkConnection` to `NetworkInfo`. In the future `NetworkInfo` will not be used and `NetworkConnection` will be used directly. The same will be done for all of the types in ` integration-tests/pkg/types`. See https://issues.redhat.com/browse/ROX-26716

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

The code change here was tested by using it in integration tests here https://github.com/stackrox/collector/pull/1899

### DIFF
--- a/integration-tests/pkg/mock_sensor/server.go
+++ b/integration-tests/pkg/mock_sensor/server.go
@@ -489,33 +489,27 @@ func (m *MockSensor) pushEndpoint(containerID string, endpoint *sensorAPI.Networ
 // translateAddress is a helper function for converting binary representations
 // of network addresses (in the signals) to usable forms for testing
 func (m *MockSensor) translateAddress(addr *sensorAPI.NetworkAddress) string {
-	address := utils.IPAddress{}
-	ipNetwork := utils.IPNetwork{}
-
+	peerId := utils.NetworkPeerID{Port: uint16(addr.GetPort())}
 	addressData := addr.GetAddressData()
 	if len(addressData) > 0 {
-		address = utils.IPFromBytes(addressData)
+		peerId.Address = utils.IPFromBytes(addressData)
+		return peerId.String()
+	}
+
+	// If there is no address data, IpNetwork should be set and represent
+	// a CIDR block or external IP address.
+	ipNetworkData := addr.GetIpNetwork()
+	if len(ipNetworkData) == 0 {
+		return peerId.String()
+	}
+
+	// If the prefix length is 32 this is a regular IP address
+	// and not a CIDR block
+	ipNetwork := utils.IPNetworkFromCIDRBytes(ipNetworkData)
+	if ipNetwork.PrefixLen() == byte(32) {
+		peerId.Address = ipNetwork.IP()
 	} else {
-		// If there is no address data IpNetwork should be set and represent
-		// a CIDR block or external IP address.
-		ipNetworkData := addr.GetIpNetwork()
-		if len(ipNetworkData) > 0 {
-			ipNetwork = utils.IPNetworkFromCIDRBytes(ipNetworkData)
-			// If the prefix length is 32 this is a regular IP address
-			// and not a CIDR block
-			if ipNetwork.PrefixLen() == byte(32) {
-				address = ipNetwork.IP()
-				ipNetwork = utils.IPNetwork{}
-			}
-		}
-
+		peerId.IPNetwork = ipNetwork
 	}
-
-	ipPortPair := utils.NetworkPeerID{
-		Address:   address,
-		IPNetwork: ipNetwork,
-		Port:      uint16(addr.GetPort()),
-	}
-
-	return ipPortPair.String()
+	return peerId.String()
 }
--- a/integration-tests/pkg/mock_sensor/server.go
+++ b/integration-tests/pkg/mock_sensor/server.go
@@ -504,11 +504,11 @@ func (m *MockSensor) translateAddress(addr *sensorAPI.NetworkAddress) string {
 	}
 
 	ipNetwork := utils.IPNetworkFromCIDRBytes(ipNetworkData)
-	numBytes := len(ipNetworkData)
 	prefixLen := ipNetwork.PrefixLen()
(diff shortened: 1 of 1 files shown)


integration-tests/pkg/mock_sensor/server.go
```diff
@@ -489,9 +489,33 @@ func (m *MockSensor) pushEndpoint(containerID string, endpoint *sensorAPI.Networ
...
+	if len(addressData) > 0 {
+		address = utils.IPFromBytes(addressData)
+	} else {
+		// If there is no address data IpNetwork should be set and represent
+		// a CIDR block or external IP address.
+		ipNetworkData := addr.GetIpNetwork()
+		if len(ipNetworkData) > 0 {
+			ipNetwork = utils.IPNetworkFromCIDRBytes(ipNetworkData)
+			// If the prefix length is 32 this is a regular IP address
+			// and not a CIDR block
```
Molter73, line 505: So what happens if we get an IPv6 address?

integration-tests/pkg/mock_sensor/server.go
```diff
@@ -489,9 +489,33 @@ func (m *MockSensor) pushEndpoint(containerID string, endpoint *sensorAPI.Networ
...
 func (m *MockSensor) translateAddress(addr *sensorAPI.NetworkAddress) string {
+	address := utils.IPAddress{}
+	ipNetwork := utils.IPNetwork{}
+
+	addressData := addr.GetAddressData()
+	if len(addressData) > 0 {
+		address = utils.IPFromBytes(addressData)
+	} else {
+		// If there is no address data IpNetwork should be set and represent
+		// a CIDR block or external IP address.
```
Molter73, line 500: What happens if we IpNetwork is not set either? Should we panic and stop the test?

integration-tests/pkg/mock_sensor/server.go
```diff
@@ -489,9 +489,33 @@ func (m *MockSensor) pushEndpoint(containerID string, endpoint *sensorAPI.Networ
...
 	ipPortPair := utils.NetworkPeerID{
-		Address: utils.IPFromBytes(addr.GetAddressData()),
-		Port:    uint16(addr.GetPort()),
+		Address:   address,
+		IPNetwork: ipNetwork,
+		Port:      uint16(addr.GetPort()),
 	}
+
 	return ipPortPair.String()
 }
```
Molter73, line 521: This would be more readable if you use early returns, something like this:
```go
func (m *MockSensor) translateAddress(addr *sensorAPI.NetworkAddress) string {
	peerId := utils.NetworkPeerID{Port: uint16(addr.GetPort())}
	addressData := addr.GetAddressData()
	if len(addressData) > 0 {
		peerId.Address = utils.IPFromBytes(addressData)
		return peerId.String()
	}

	// If there is no address data, IpNetwork should be set and represent
	// a CIDR block or external IP address.
	ipNetworkData := addr.GetIpNetwork()
	if len(ipNetworkData) == 0 {
		return peerId.String()
	}

	// If the prefix length is 32 this is a regular IP address
	// and not a CIDR block
	ipNetwork := utils.IPNetworkFromCIDRBytes(ipNetworkData)
	if ipNetwork.PrefixLen() == byte(32) {
		peerId.Address = ipNetwork.IP()
	} else {
		peerId.IPNetwork = ipNetwork
	}
	return peerId.String()
}
```

JoukoVirtanen: Done

integration-tests/pkg/mock_sensor/server.go
JoukoVirtanen: No. The comment wasn't accurate for the local address both the address data and IpNetwork can be blank. I have corrected the comment.

integration-tests/pkg/mock_sensor/server.go
JoukoVirtanen: The code now hands IPv6.

integration-tests/pkg/mock_sensor/server.go
```diff
@@ -489,9 +489,29 @@ func (m *MockSensor) pushEndpoint(containerID string, endpoint *sensorAPI.Networ
...
+	ipNetwork := utils.IPNetworkFromCIDRBytes(ipNetworkData)
+	numBytes := len(ipNetworkData)
+	prefixLen := ipNetwork.PrefixLen()
+	// If this is IPv4 and the prefix length is 32 or this is IPv6 and the prefix length
+	// is 128 this is a regular IP address and not a CIDR block
+	if (numBytes == 5 && prefixLen == byte(32)) || (numBytes == 17 && prefixLen == byte(128)) {
+		peerId.Address = ipNetwork.IP()
+	} else {
+		peerId.IPNetwork = ipNetwork
+	}
```
Molter73, line 515: I was just looking through the IPNetwork definition on the main repo and it has a `Family()` method, any particualr reason why we couldn't do something like this?
```suggestion
	ipNetwork := utils.IPNetworkFromCIDRBytes(ipNetworkData)
	prefixLen := ipNetwork.PrefixLen()
	// If this is IPv4 and the prefix length is 32 or this is IPv6 and the prefix length
	// is 128 this is a regular IP address and not a CIDR block
	if (ipNetwork.Family() == utils.IPv4 && prefixLen == byte(32)) ||
		(ipNetwork.Family() == utils.IPv6 && prefixLen == byte(128)) {
		peerId.Address = ipNetwork.IP()
	} else {
		peerId.IPNetwork = ipNetwork
	}
```

Molter73 (approved):

PR #1902, opened 2024-10-23
ROX-26726: Collector integration tests should record all connections

JoukoVirtanen:
## Description

Currently the mock server only maintains a map where the key is a networking connection, instead of a full list of all connections that it sees, this means that if a connection is opened and closed multiple times it will only save the opening event once. For some tests all of the opening events will need to be recorded. For the runtime configuration tests the same connection will be opened and closed multiple times, so the changes here are needed for it. See https://github.com/stackrox/collector/pull/1899

This PR is built on another PR that makes it so that the mock server can properly handle connections involving external IPs and CIDR blocks. https://github.com/stackrox/collector/pull/1901

Other changes are also made. 

1. The addition of a function that requires that the elements of the list of expected connections matches the elements of the list of observed. This differs from the existing function, which only requires that each expected connection be seen. The old function would tolerate having extra observed connections that were not expected.

2. Connections with different last active timestamps are considered equivalent. 

In the future similar changes will be made for networking endpoints and processes.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

The changes in this PR were used to add more integration tests here https://github.com/stackrox/collector/pull/1899

### DIFF
--- a/integration-tests/pkg/assert/assert.go
+++ b/integration-tests/pkg/assert/assert.go
@@ -25,12 +25,10 @@ func ElementsMatchFunc[N any](expected []N, actual []N, equal func(a, b N) bool)
 
 func ListsToAssertMsg[N any](expected []N, actual []N) string {
 	var expectedBuf, actualBuf bytes.Buffer
-	spew.Fdump(&expectedBuf, expected)
-	spew.Fdump(&actualBuf, actual)
 	return fmt.Sprintf(
 		"Expected elements:\n%s\n\nActual elements:\n%s\n",
-		expectedBuf.String(),
-		actualBuf.String(),
+		spew.Sdump(expected),
+		spew.Sdump(actual),
 	)
 }
 
--- a/integration-tests/pkg/assert/assert.go
+++ b/integration-tests/pkg/assert/assert.go
@@ -36,7 +36,7 @@ func AssertElementsMatchFunc[N any](t *testing.T, expected []N, actual []N, equa
 	match := ElementsMatchFunc(expected, actual, equal)
 	if !match {
 		assertMsg := ListsToAssertMsg(expected, actual)
-		assert.True(t, match, assertMsg)
+		assert.Fail(t, assertMsg)
 	}
 	return match
 }
--- a/integration-tests/pkg/assert/assert.go
+++ b/integration-tests/pkg/assert/assert.go
@@ -1,10 +1,12 @@
 package assert
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"testing"
 
+	"github.com/davecgh/go-spew/spew"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -22,18 +24,13 @@ func ElementsMatchFunc[N any](expected []N, actual []N, equal func(a, b N) bool)
 }
 
 func ListsToAssertMsg[N any](expected []N, actual []N) string {
-	formatList := func(list []N) string {
-		data, err := json.MarshalIndent(list, "", "  ")
-		if err != nil {
-			return fmt.Sprintf("Failed to format list: %v", err)
-		}
-		return string(data)
-	}
-
+	var expectedBuf, actualBuf bytes.Buffer
+	spew.Fdump(&expectedBuf, expected)
+	spew.Fdump(&actualBuf, actual)
 	return fmt.Sprintf(
 		"Expected elements:\n%s\n\nActual elements:\n%s\n",
-		formatList(expected),
-		formatList(actual),
+		expectedBuf.String(),
+		actualBuf.String(),
 	)
 }
 
--- a/integration-tests/pkg/assert/assert.go
+++ b/integration-tests/pkg/assert/assert.go
@@ -1,0 +1,47 @@
+package assert
+
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func ElementsMatchFunc[N any](expected []N, actual []N, equal func(a, b N) bool) bool {
+	if len(expected) != len(actual) {
+		return false
+	}
+
+	for i := range expected {
+		if !equal(expected[i], actual[i]) {
+			return false
+		}
+	}
+	return true
+}
+
+func ListsToAssertMsg[N any](expected []N, actual []N) string {
+	formatList := func(list []N) string {
+		data, err := json.MarshalIndent(list, "", "  ")
+		if err != nil {
+			return fmt.Sprintf("Failed to format list: %v", err)
+		}
+		return string(data)
+	}
+
+	return fmt.Sprintf(
+		"Expected elements:\n%s\n\nActual elements:\n%s\n",
+		formatList(expected),
+		formatList(actual),
+	)
+}
+
+func AssertElementsMatchFunc[N any](t *testing.T, expected []N, actual []N, equal func(a, b N) bool) bool {
+	match := ElementsMatchFunc(expected, actual, equal)
+	if !match {
+		assertMsg := ListsToAssertMsg(expected, actual)
+		assert.True(t, match, assertMsg)
+	}
+	return match
+}
--- a/integration-tests/pkg/assert/assert.go
+++ b/integration-tests/pkg/assert/assert.go
@@ -1,8 +1,6 @@
 package assert
 
 import (
-	"bytes"
-	"encoding/json"
 	"fmt"
 	"testing"
 
@@ -24,7 +22,6 @@ func ElementsMatchFunc[N any](expected []N, actual []N, equal func(a, b N) bool)
 }
 
 func ListsToAssertMsg[N any](expected []N, actual []N) string {
-	var expectedBuf, actualBuf bytes.Buffer
 	return fmt.Sprintf(
 		"Expected elements:\n%s\n\nActual elements:\n%s\n",
 		spew.Sdump(expected),
--- a/integration-tests/pkg/mock_sensor/expect_conn.go
+++ b/integration-tests/pkg/mock_sensor/expect_conn.go
@@ -97,6 +97,21 @@ func (s *MockSensor) checkIfConnectionsMatchExpected(t *testing.T, connections [
 	return false
 }
 
+// getConnectionsAndCompare gets the connections for a container, sorts them if order is set to true, and compares with a set of expected
+// connections. If asssertMismatch is true and the set of observed connections does not match the set of expected connections an assert is
+// triggered. If assertMismatch is not set then just return if the observed and expected connections match.
+func (s *MockSensor) getConnectionsAndCompare(t *testing.T, containerID string, order bool, assertMismatch bool, expected ...types.NetworkInfo) bool {
+	connections := s.Connections(containerID)
+	if order {
+		types.SortConnections(connections)
+	}
+	success := s.checkIfConnectionsMatchExpected(t, connections, expected)
+	if assertMismatch && !success {
+		return assert.ElementsMatch(t, expected, connections, "networking connections do not match")
+	}
+	return success
+}
+
 // CompareConnections compares a list of expected connections to the observed connections. This comparison is done at the beginning, when a new
 // connection arrives, and after a timeout period. The number of connections must match and it can be specified if the order of the connections
 // must match or not. The difference between this function and ExpectConnections is that ExpectConnections tolerates extra observed connections
@@ -105,28 +120,25 @@ func (s *MockSensor) CompareConnections(t *testing.T, containerID string, timeou
 	if order {
 		types.SortConnections(expected)
 	}
-	timer := time.NewTimer(timeout)
-	ticker := time.NewTicker(100 * time.Millisecond)
-	defer ticker.Stop()
+
+	success := s.getConnectionsAndCompare(t, containerID, order, false, expected...)
+	if success {
+		return true
+	}
+
+	timer := time.After(timeout)
 
 	for {
 		select {
-		case <-timer.C:
-			connections := s.Connections(containerID)
-			if order {
-				types.SortConnections(connections)
+		case <-timer:
+			return s.getConnectionsAndCompare(t, containerID, order, true, expected...)
+		case conn := <-s.LiveConnections():
+			if conn.GetContainerId() != containerID {
+				continue
 			}
-			return s.checkIfConnectionsMatchExpected(t, connections, expected)
-		case <-ticker.C:
-			connections := s.Connections(containerID)
-			if order {
-				types.SortConnections(connections)
-			}
-			success := s.checkIfConnectionsMatchExpected(t, connections, expected)
+			success := s.getConnectionsAndCompare(t, containerID, order, false, expected...)
 			if success {
 				return true
-			} else if len(connections) >= len(expected) {
-				return assert.ElementsMatch(t, expected, connections, "networking connections do not match")
 			}
 		}
 	}
@@ -135,13 +147,13 @@ func (s *MockSensor) CompareConnections(t *testing.T, containerID string, timeou
 // ExpectExactConnections requires that within a timeout period the networking connections involving containerID match a list of expected
 // netwoking connections.
 func (s *MockSensor) ExpectExactConnections(t *testing.T, containerID string, timeout time.Duration, expected ...types.NetworkInfo) bool {
-	return s.CompareConnections(t, containerID, timeout, true, expected...)
+	return s.CompareConnections(t, containerID, timeout, false, expected...)
 }
 
 // ExpectSameElementsConnections requires that within a timeout period the networking connections involving containerID match a list of expected
 // netwoking connections, but the order of those connections do not have to match.
 func (s *MockSensor) ExpectSameElementsConnections(t *testing.T, containerID string, timeout time.Duration, expected ...types.NetworkInfo) bool {
-	return s.CompareConnections(t, containerID, timeout, false, expected...)
+	return s.CompareConnections(t, containerID, timeout, true, expected...)
 }
 
 // ExpectEndpoints waits up to the timeout for the gRPC server to receive
--- a/integration-tests/pkg/mock_sensor/expect_conn.go
+++ b/integration-tests/pkg/mock_sensor/expect_conn.go
@@ -86,7 +86,6 @@ func (s *MockSensor) checkIfConnectionsMatchExpected(t *testing.T, connections [
 	}
 
 	if len(connections) == len(expected) {
-		types.SortConnections(connections)
 		for i := range expected {
 			if !expected[i].Equal(connections[i]) {
 				return assert.ElementsMatch(t, expected, connections, "networking connections do not match")
--- a/integration-tests/pkg/mock_sensor/expect_conn.go
+++ b/integration-tests/pkg/mock_sensor/expect_conn.go
@@ -103,8 +103,7 @@ func (s *MockSensor) ExpectSameElementsConnections(t *testing.T, containerID str
 	types.SortConnections(expected)
 
 	connections := s.SortedConnections(containerID)
-	success := s.checkIfConnectionsMatchExpected(t, connections, expected)
-	if success {
+	if s.checkIfConnectionsMatchExpected(t, connections, expected) {
 		return true
 	}
 
--- a/integration-tests/pkg/mock_sensor/expect_conn.go
+++ b/integration-tests/pkg/mock_sensor/expect_conn.go
@@ -33,7 +33,7 @@ loop:
 		case <-timer:
 			// we know they don't match at this point, but by using
 			// ElementsMatch we get much better logging about the differences
-			return assert.ElementsMatch(t, expected, s.Connections(containerID), "timed out waiting for networks")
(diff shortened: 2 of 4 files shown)


integration-tests/pkg/mock_sensor/expect_conn.go
```diff
@@ -72,14 +72,78 @@ loop:
...
+// CompareConnections compares a list of expected connections to the observed connections. This comparison is done at the beginning, when a new
+// connection arrives, and after a timeout period. The number of connections must match and it can be specified if the order of the connections
+// must match or not. The difference between this function and ExpectConnections is that ExpectConnections tolerates extra observed connections
+// that are not expected.
+func (s *MockSensor) CompareConnections(t *testing.T, containerID string, timeout time.Duration, order bool, expected ...types.NetworkInfo) bool {
+	if order {
+		types.SortConnections(expected)
+	}
+	timer := time.NewTimer(timeout)
+	ticker := time.NewTicker(100 * time.Millisecond)
```
JoukoVirtanen, line 109: Instead of using a ticker get the connections when a new connection comes in, like the other functions.

integration-tests/pkg/mock_sensor/expect_conn.go
```diff
@@ -72,14 +72,90 @@ loop:
...
+		types.SortConnections(connections)
+		for i := range expected {
+			if !expected[i].Equal(connections[i]) {
+				return assert.ElementsMatch(t, expected, connections, "networking connections do not match")
+			}
+		}
+		return true
+	}
+	return false
+}
```
Molter73, line 98: Isn't this whole method the same as  just calling `assert.ElementsMatch(t, expected, connections, "networking connections do not match")`? From the testify docs:
```
ElementsMatch asserts that the specified listA(array, slice...) is equal to specified listB(array, slice...) ignoring the order of the elements. If there are duplicate elements, the number of appearances of each of them in both lists should match. 
```
https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch

integration-tests/pkg/mock_sensor/expect_conn.go
```diff
@@ -72,14 +72,90 @@ loop:
...
+
+// getConnectionsAndCompare gets the connections for a container, sorts them if order is set to true, and compares with a set of expected
+// connections. If asssertMismatch is true and the set of observed connections does not match the set of expected connections an assert is
+// triggered. If assertMismatch is not set then just return if the observed and expected connections match.
+func (s *MockSensor) getConnectionsAndCompare(t *testing.T, containerID string, order bool, assertMismatch bool, expected ...types.NetworkInfo) bool {
```
Molter73, line 103: If `checkIfConnectionsMatchExpected` always sorts the container connections, what is the point of ordering them here based on the `order` parameter?

Molter73: Also, why do we need `assertMismatch`? Couldn't the caller just check if the returned value is false?

Molter73: On a separate note about `order`, the way collector handles connections is inherently asynchronous, so even with extreme care, if the implementation preserved the order, expecting a given order of connections to always be in the same order seems unrealistic and a great way to add flakes to our tests.

Molter73 (commented):
I'm half way through the review and I've decided not to finish it. While I see value in having the mock sensor store all the connections seen, the added checks in `expect_conn.go` add complexity where it isn't needed (see my comments on that file for more details). Please revisit the implementation to see if we can get rid of most of the methods in there and reach out to me if you want to discuss potential alternatives.

integration-tests/pkg/mock_sensor/server.go
```diff
@@ -47,7 +46,7 @@ type MockSensor struct {
 	processLineages map[string]LineageMap
 	processMutex    sync.Mutex
 
-	connections  map[string]ConnMap
+	connections  map[string][]types.NetworkInfo
```
Stringy, line 49: it might be worth retaining the type alias for consistency with the other maps (but renamed to something else like `ConnList`)

integration-tests/pkg/mock_sensor/expect_conn.go
Stringy: +1 to @Molter73's point about avoiding reliance on connection order. We should probably always sort both lists (from collector and the expected info from the tests themselves)

Stringy: It might even make sense for `s.Connections()` to return a sorted list

integration-tests/pkg/mock_sensor/server.go
```diff
@@ -171,8 +166,11 @@ func (m *MockSensor) HasConnection(containerID string, conn types.NetworkInfo) b
 	defer m.networkMutex.Unlock()
 
 	if conns, ok := m.connections[containerID]; ok {
-		_, exists := conns[conn]
-		return exists
+		for _, connection := range conns {
+			if connection.Equal(conn) {
+				return true
+			}
+		}
```
Stringy, line 173: I think we can use [slices.ContainsFunc](https://pkg.go.dev/slices#ContainsFunc) here:
```suggestion
        return slices.ContainsFunc(conns, func(c Connection) bool {
            return c.Equal(conn)
        })
```

integration-tests/pkg/mock_sensor/expect_conn.go
JoukoVirtanen: Done

integration-tests/pkg/mock_sensor/expect_conn.go
JoukoVirtanen: No. There is a difference. This uses the `Equal` method, which is not what `ElementsMatch` uses. The `Equal` method views timestamps as being equal if they are both non-nil or both nil, while `ElementsMatch` checks that the timestamps are actually the same. This is similar to what other existing parts of the code do. E.g. `ExpectEndpoints` checks if an endpoint is in the list of expected endpoints using `HasEndpoint`, which uses the `Equal` method for endpoints. If this is not the case when the timer runs out, `ElementsMatch` is called to get better logging about the differences. This is the same reason that `ElementsMatch` is called in the code that I wrote here.

integration-tests/pkg/mock_sensor/expect_conn.go
JoukoVirtanen: We may expect a certain order when all of the connections are reported in different scrapes. The connections within a scrape are not ordered, but the scrapes themselves are ordered.

integration-tests/pkg/mock_sensor/server.go
JoukoVirtanen: I think other PRs should change the other maps to lists as well. I think naming a list `ConnList` will just obfuscate things.

JoukoVirtanen: I think other PRs should change the other maps to lists as well. I think naming a list `ConnList` will just obfuscate things.

integration-tests/pkg/mock_sensor/expect_conn.go
JoukoVirtanen: > If checkIfConnectionsMatchExpected always sorts the container connections, what is the point of ordering them here based on the order parameter?

That was a mistake. The connections should not be sorted in `checkIfConnectionsMatchExpected`. I have removed that sort.

JoukoVirtanen: > Also, why do we need assertMismatch? Couldn't the caller just check if the returned value is false?

We don't. I did that to simplify `CompareConnections`.

integration-tests/pkg/mock_sensor/server.go
JoukoVirtanen: Done. Though I didn't commit this exact change, because there is an error in it.

JoukoVirtanen:
> I'm half way through the review and I've decided not to finish it. While I see value in having the mock sensor store all the connections seen, the added checks in `expect_conn.go` add complexity where it isn't needed (see my comments on that file for more details). Please revisit the implementation to see if we can get rid of most of the methods in there and reach out to me if you want to discuss potential alternatives.

I didn't think that the existing functions such as `ExpectConnectionsN` and `ExpectConnections` do what I wanted for external IPs integration tests. The problem with `ExpectConnections` is that it passes if there are extra observed connections that are not expected. That is why I decided to create my own function to check if the observed connections match the expected connections. The simplest thing to do would have been to call `ExpectConnectionsN` in a new function that would check if the connections returned by `ExpectConnectionsN` are correct if `ExpectConnectionsN` passes. The reason I didn't do that is that if `ExpectConnectionsN` finds the wrong number of connections it fails without logging what the observed and expected connections are.

I am open to simplifications such as always sorting the endpoints and not using `assertMismatch`.

integration-tests/pkg/mock_sensor/expect_conn.go
JoukoVirtanen: > It might even make sense for s.Connections() to return a sorted list

I don't like this idea as we might want to not sort the connections in the future. However, I have created `SortConnections` which does what you propose.

Stringy: > We may expect a certain order when all of the connections are reported in different scrapes. The connections within a scrape are not ordered, but the scrapes themselves are ordered.

Perhaps this is reason enough to store a full network message ([NetworkConnectionInfo](https://github.com/stackrox/stackrox/blob/master/proto/internalapi/sensor/network_connection_info.proto#L13)) from Collector rather than extracting individual connections or endpoints. That way we can more accurately test behaviour for a given scrape (represented by a single message.) Another reason is that we currently have no way of deterministically waiting for a scrape to happen and the existing tests that do this simply make a best guess, but if we could wait for the mock sensor to receive a network message then we have a very neat way of doing that wait.

At that point the notion of whether we should or shouldn't sort Connections becomes a non-issue.

> I don't like this idea as we might want to not sort the connections in the future.

We can always undo it in the future if it becomes a problem

integration-tests/pkg/mock_sensor/expect_conn.go
```diff
@@ -72,14 +72,66 @@ loop:
...
+// connection arrives, and after a timeout period. The number of connections must match and the expected and observed connections must match, but the order
+// does not matter.
+func (s *MockSensor) ExpectSameElementsConnections(t *testing.T, containerID string, timeout time.Duration, expected ...types.NetworkInfo) bool {
+	types.SortConnections(expected)
+
+	connections := s.SortedConnections(containerID)
+	success := s.checkIfConnectionsMatchExpected(t, connections, expected)
+	if success {
+		return true
+	}
```
Stringy, line 109: [nit] 
```suggestion
	if s.checkIfConnectionsMatchExpected(t, connections, expected) {
		return true
	}
```

integration-tests/pkg/mock_sensor/server.go
```diff
@@ -155,24 +155,28 @@ func (m *MockSensor) Connections(containerID string) []types.NetworkInfo {
 	defer m.networkMutex.Unlock()
 
 	if connections, ok := m.connections[containerID]; ok {
-		keys := make([]types.NetworkInfo, 0, len(connections))
-		for k := range connections {
-			keys = append(keys, k)
-		}
-		return keys
+		return connections
```
Stringy, line 158: I think this is unsafe, we're returning a reference to the connections slice, and that slice may be modified by both the sensor thread (more connections come in for this containerID) and the main thread (sorting, asserting, etc)

```suggestion
        conns := make([]types.NetworkInfo, len(connections))
        copy(conns, connections)
		return conns
```

integration-tests/pkg/mock_sensor/expect_conn.go
Stringy: But then we have a situation here where the `Equal` method returns false, but `ElementsMatch` considers those elements to be equal, so we neither get the desired result in the test, nor the additional logging that we require. We may need to write our own which lets us use a custom Equals function - something like:

```go
func ElementsMatchFunc[N](t *testing.T, expected []N, actual []N, equal func(a, b N) bool) bool {
    // ... 
} 
```

The main difficulty I think will be constructing the appropriate logging message for the failure but we can probably use [go-spew](https://github.com/davecgh/go-spew) or [litter](https://github.com/sanity-io/litter) to help with this

Stringy (changes requested):

integration-tests/pkg/mock_sensor/expect_conn.go
JoukoVirtanen: I am now doing the sort in `Connections` and have removed `SortedConnections`.

JoukoVirtanen:
>  But then we have a situation here where the Equal method returns false, but ElementsMatch considers those elements to be equal, so we neither get the desired result in the test, nor the additional logging that we require. We may need to write our own which lets us use a custom Equals function - something like:

> func ElementsMatchFunc[N](t *testing.T, expected []N, actual []N, equal func(a, b N) bool) bool {
    // ... 
} 
The main difficulty I think will be constructing the appropriate logging message for the failure but we can probably use go-spew or litter to help with this

I have created the `ElementsMatchFunc`, but I am not using either of the libraries you suggested for logging. I can change that.

integration-tests/pkg/mock_sensor/expect_conn.go
```diff
@@ -72,14 +72,67 @@ loop:
...
-
 			if len(s.Connections(containerID)) == n {
 				return s.Connections(containerID)
 			}
 		}
 	}
 }
 
+// checkIfConnectionsMatchExpected compares a list of expected and observed connection match exactly.
+func (s *MockSensor) checkIfConnectionsMatchExpected(t *testing.T, connections []types.NetworkInfo, expected []types.NetworkInfo) bool {
```
Stringy, line 83: I think this function is redundant now because the extra logic is already handled by `collectorAssert.ElementsMatchFunc`, so in `ExpectSameElementsConnections` below, we should be able to just do:

```
collectorAssert.AssertElementsMatchFunc(t, expected, connections, func(a, b types.NetworkInfo) bool {
    return a.Equals(b)
})
```

If we also make the receiver for types.NetworkInfo.Equal a value rather than a pointer, we can further simplify:

```
collectorAssert.AssertElementsMatchFunc(t, expected, connections, types.NetworkInfo.Equal)
```

we'd get a copy I think, but that's almost certainly fine for the scale we have in the tests

integration-tests/pkg/assert/assert.go
```diff
@@ -0,0 +1,44 @@
...
+		expectedBuf.String(),
+		actualBuf.String(),
+	)
+}
+
+func AssertElementsMatchFunc[N any](t *testing.T, expected []N, actual []N, equal func(a, b N) bool) bool {
+	match := ElementsMatchFunc(expected, actual, equal)
+	if !match {
+		assertMsg := ListsToAssertMsg(expected, actual)
+		assert.True(t, match, assertMsg)
```
Stringy, line 41: `assert.FailNow` would be more appropriate here, and would have us an additional confusing log message about false not being true

integration-tests/pkg/assert/assert.go
```diff
@@ -0,0 +1,44 @@
...
+	spew.Fdump(&expectedBuf, expected)
+	spew.Fdump(&actualBuf, actual)
+	return fmt.Sprintf(
+		"Expected elements:\n%s\n\nActual elements:\n%s\n",
+		expectedBuf.String(),
+		actualBuf.String(),
+	)
+}
+
+func AssertElementsMatchFunc[N any](t *testing.T, expected []N, actual []N, equal func(a, b N) bool) bool {
```
Stringy, line 37: Based on the name of this function, I would not expect it to return anything, and panic (via `assert.FailNow`) when the condition is false. If we need another that does return success or failure, then it should be `ExpectElementsMatchFunc`

integration-tests/pkg/assert/assert.go
```diff
@@ -0,0 +1,44 @@
...
+
+	for i := range expected {
+		if !equal(expected[i], actual[i]) {
+			return false
+		}
+	}
+	return true
+}
+
+func ListsToAssertMsg[N any](expected []N, actual []N) string {
```
Stringy, line 26: I don't think this needs to be exposed outside this package - all message handling should happen internally

integration-tests/pkg/assert/assert.go
```diff
@@ -0,0 +1,44 @@
...
+	var expectedBuf, actualBuf bytes.Buffer
+	spew.Fdump(&expectedBuf, expected)
+	spew.Fdump(&actualBuf, actual)
+	return fmt.Sprintf(
+		"Expected elements:\n%s\n\nActual elements:\n%s\n",
+		expectedBuf.String(),
+		actualBuf.String(),
```
Stringy, line 33: [nit]

```suggestion
	return fmt.Sprintf(
		"Expected elements:\n%s\n\nActual elements:\n%s\n",
		spew.Sdump(expected),
		spew.Sdump(actual),
```

integration-tests/pkg/mock_sensor/expect_conn.go
Stringy: In fact, we already copy because the function passed to `ElementsMatchFunc` takes the variables by value anyway, so changing types.NetworkInfo.Equal would really just make things a little simpler

integration-tests/pkg/assert/assert.go
Stringy: or `assert.Fail` if we want to return

integration-tests/pkg/mock_sensor/expect_conn.go
```diff
@@ -72,14 +72,67 @@ loop:
...
+	if len(connections) > len(expected) {
+		assert.FailNow(t, collectorAssert.ListsToAssertMsg(expected, connections))
+	}
+
+	// If the number of observed connections equals the number of expected connections, but the elements are
+	// different, they will never be the same. Exit early
+	if len(connections) == len(expected) && !collectorAssert.ElementsMatchFunc(connections, expected, equal) {
+		assert.FailNow(t, collectorAssert.ListsToAssertMsg(expected, connections))
+	}
+	return collectorAssert.ElementsMatchFunc(connections, expected, equal)
```
Molter73, line 98:
```suggestion
	// We haven't gotten all connections yet, we'll try again later
	if len(connection) < len(expected) {
		return false
	}

	return collectorAssert.ElementsMatchFunc(connections, expected, func(c1, c2 types.NetworkInfo) bool {
		return c1.Equal(c2)
	})

```

integration-tests/pkg/mock_sensor/expect_conn.go
```diff
@@ -72,14 +72,67 @@ loop:
...
+
+	for {
+		select {
+		case <-timer:
+			connections := s.Connections(containerID)
+			success := s.checkIfConnectionsMatchExpected(t, connections, expected)
+			if !success {
+				return assert.ElementsMatch(t, expected, connections, "networking connections do not match")
+			}
+			return success
```
Molter73, line 122: [nit]
```suggestion
			if !s.checkIfConnectionsMatchExpected(t, connections, expected) {
				return assert.ElementsMatch(t, expected, connections, "networking connections do not match")
			}
			return true
```

integration-tests/pkg/mock_sensor/expect_conn.go
```diff
@@ -72,14 +72,67 @@ loop:
...
+		case conn := <-s.LiveConnections():
+			if conn.GetContainerId() != containerID {
+				continue
+			}
+			connections := s.Connections(containerID)
+			success := s.checkIfConnectionsMatchExpected(t, connections, expected)
+			if success {
```
Molter73, line 129: [nit]
```suggestion
			if s.checkIfConnectionsMatchExpected(t, connections, expected) {
```

integration-tests/pkg/mock_sensor/expect_conn.go
JoukoVirtanen: I am now just doing
```
return collectorAssert.AssertElementsMatchFunc(t, connections, expected, equal)
```

integration-tests/pkg/mock_sensor/expect_conn.go
```diff
@@ -72,14 +72,47 @@ loop:
...
 	}
 }
 
+// ExpectSameElementsConnections compares a list of expected connections to the observed connections. This comparison is done at the beginning, when a new
+// connection arrives, and after a timeout period. The number of connections must match and the expected and observed connections must match, but the order
+// does not matter.
+func (s *MockSensor) ExpectSameElementsConnections(t *testing.T, containerID string, timeout time.Duration, expected ...types.NetworkInfo) bool {
+	types.SortConnections(expected)
+
+	equal := func(c1 types.NetworkInfo, c2 types.NetworkInfo) bool {
```
Molter73, line 88: [nit] This can be made a bit less verbose
```suggestion
	equal := func(c1, c2 types.NetworkInfo) bool {
```

Molter73 (approved):
LGTM! Just a minor nit that you are free to ignore if you want to.

Stringy (approved):

PR #1904, opened 2024-10-23
Use inotify for reloading configuration in all cases

Molter73:
## Description

Stop polling for the runtime configuration file and instead fully rely on inotify for reloading. In order to achieve this, 2 new classes are introduced:

- Inotify: handles operations involving inotify, including:
  - Opening and closing the inotify file descriptor.
  - Adding, removing and keeping track of file/directory watchers.
  - Reading events and forwarding them to a caller in a non-blocking fashion.
- ConfigLoader: uses an Inotify object for tracking changes to the runtime configuration file and update it accordingly.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [x] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Tested manually on a kind cluster.

Two k8s integration tests were added using the configuration introspection endpoint:
- First test:
  - Launch collector with a config map already created.
  - Check the external IPs feature has the correct setting.
  - Edit the configmap and check collector reloads the configuration correctly.
- Second test:
  - Launch collector with no configmap and check runtime configuration is empty.
  - Create the configuration map and check collector starts using runtime configuration with the correct value.
  - Edit the configmap and check collector reloads the configuration correctly.
  - Remove the configmap and check collector drops runtime configuration.


collector/lib/Inotify.cpp
```diff
@@ -0,0 +1,132 @@
...
+int Inotify::RemoveWatcher(WatcherIterator it) {
+  int res = 0;
+  if (it != watchers_.end()) {
+    res = inotify_rm_watch(fd_, it->wd);
+    watchers_.erase(it, it + 1);
+  }
+  return res;
+}
+
+Inotify::WatcherIterator Inotify::FindWatcher(const std::filesystem::path& needle) {
```
JoukoVirtanen, line 64: Why name the variable `needle`.

JoukoVirtanen: This probably doesn't matter very much since we won't have many watchers, but maybe use a map to avoid iterating.

Molter73: > Why name the variable `needle`.

Because you could think of the watcher store as a haystack, this is something I've seen before and kinda stuck with me. As an example, from `man strstr`:
```
char *strstr(const char *haystack, const char *needle);
```


> This probably doesn't matter very much since we won't have many watchers, but maybe use a map to avoid iterating.

I left a comment w.r.t. this:
https://github.com/stackrox/collector/blob/ed1540635dcb1b8d1bb2af0130764b9ccb2d5af7/collector/lib/Inotify.h#L103-L108
TL;DR we would need 2 maps, one order by path and another by wd, so instead I just used a vector to KISS.

integration-tests/suites/k8s/base.go
```diff
@@ -0,0 +1,189 @@
+package k8s
```
JoukoVirtanen, line 1: Thanks for moving things from `k8s_namespace.go` to here.

integration-tests/pkg/collector/introspection.go
```diff
@@ -0,0 +1,26 @@
...
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/stackrox/collector/integration-tests/pkg/log"
+)
+
+func (k *K8sCollectorManager) IntrospectionQuery(endpoint string) ([]byte, error) {
+	uri := fmt.Sprintf("http://%s:8080%s", k.IP(), endpoint)
+	body := []byte{}
```
JoukoVirtanen, line 13: Unneeded declaration of body.

integration-tests/pkg/collector/introspection.go
```diff
@@ -0,0 +1,26 @@
...
+	if err != nil {
+		return body, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return body, log.Error("IntrospectionQuery failed with %s", resp.Status)
+	}
+
+	defer resp.Body.Close()
+	body, err = io.ReadAll(resp.Body)
```
JoukoVirtanen, line 24:
```suggestion
	body, err := io.ReadAll(resp.Body)
```
If you are removing the declaration of body above then this change is needed.

integration-tests/pkg/collector/introspection.go
```diff
@@ -0,0 +1,26 @@
...
+	resp, err := http.Get(uri)
+	if err != nil {
+		return body, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return body, log.Error("IntrospectionQuery failed with %s", resp.Status)
+	}
+
+	defer resp.Body.Close()
```
JoukoVirtanen, line 23: This should be moved to after `resp, err := http.Get(uri)`

integration-tests/pkg/collector/collector_k8s.go
```diff
@@ -179,12 +200,28 @@ func (k *K8sCollectorManager) IsRunning() (bool, error) {
...
+			Namespace: TEST_NAMESPACE,
+		}
+
+		k.containerID = k.executor.PodContainerID(cf)
 	}
+	return k.containerID
+}
 
-	return k.executor.PodContainerID(cf)
+func (k *K8sCollectorManager) IP() string {
```
JoukoVirtanen, line 214: This could be moved to k8s/base.go and it could take the pod name as a parameter rather than hard coding "collector". Then here we would just have a one line function that calls IP("collector").

collector/lib/Inotify.h
```diff
@@ -0,0 +1,292 @@
...
+        os << "\n";
+      }
+
+      os << w;
+    }
+    return os;
+  }
+
+  // Taken from <sys/inotify.h>
+  static constexpr std::array<unsigned int, 17> BIT_MASKS{
```
JoukoVirtanen, line 247: Why copy from there?

integration-tests/suites/k8s/config_reload.go
```diff
@@ -0,0 +1,165 @@
...
+	k.AssertExternalIps(true)
+
+	log.Info("Checking external IPs is disabled")
+	configMap.Data["runtime_config.yaml"] = EXT_IP_DISABLE
+	k.updateConfigMap(&configMap)
+	k.AssertExternalIps(false)
+
+	log.Info("Checking runtime configuration is not in use")
+	k.deleteConfigMap(CONFIG_MAP_NAME)
+	k.assertNoRuntimeConfig()
```
JoukoVirtanen, line 87: Would be good to create it one more time and make sure that the ConfigMap can be read after being deleted.

integration-tests/suites/k8s/config_reload.go
```diff
@@ -0,0 +1,165 @@
...
+			}
+
+		case <-timer:
+			k.FailNow("Runtime configuration was not updated")
+		}
+	}
+}
+
+func (k *K8sConfigReloadTestSuite) assertNoRuntimeConfig() {
+	tick := time.Tick(10 * time.Second)
```
JoukoVirtanen, line 144: I think the tick can be much faster. I don't think it costs us anything to have a faster tick and should make the test faster. Waiting for 3 minutes also seems excessive.

integration-tests/suites/k8s/config_reload.go
```diff
@@ -0,0 +1,165 @@
...
+	})
+	k.AssertExternalIps(true)
+
+	log.Info("Checking external IPs is disabled")
+	configMap.Data["runtime_config.yaml"] = EXT_IP_DISABLE
+	k.updateConfigMap(&configMap)
+	k.AssertExternalIps(false)
+}
+
+func (k *K8sConfigReloadTestSuite) AssertExternalIps(enable bool) {
```
JoukoVirtanen, line 116: Feels like `AssertExternalIps` and `assertNoRuntimeConfig` should be combined.

integration-tests/suites/k8s/config_reload.go
```diff
@@ -0,0 +1,165 @@
...
+	EXT_IP_ENABLE = `
+networking:
+  externalIps:
+    enable: true
+`
+
+	EXT_IP_DISABLE = `
+networking:
+  externalIps:
+    enable: false
```
JoukoVirtanen, line 25: I also tested invalid input in my PR, but maybe that can be done here later in another PR.

JoukoVirtanen: Also I had a function that accepted a bool, converted it to a string, and added it to the config string. I think that seems a bit better than this.

collector/lib/Inotify.h
Molter73: Because this way I can have an iterable array, rather than unrelated `#define` statements that mean nothing on their own.

integration-tests/pkg/collector/collector_k8s.go
Molter73: We need an executor to do this, so rather than `pkg/k8s` I would consider moving it to the k8s specific executor.

integration-tests/suites/k8s/base.go
Molter73: This file mirrors `integration-tests/suites/base.go`, we can consider moving more things from these files to a different package in a follow up PR, since setting up the testing environment for k8s is not the focus of this one.

integration-tests/suites/k8s/config_reload.go
Molter73: I will consider doing these on follow up PRs, right now the tests are minimal and focused on correct functionality, not error handling that has already been tested in unit tests.

Molter73: > Also I had a function that accepted a bool, converted it to a string, and added it to the config string. I think that seems a bit better than this.

There's probably a way to define the full structure and serialize it to yaml that would be easier than this described approach, for now, the current approach is good enough for the implemented tests IMO.

integration-tests/suites/k8s/config_reload.go
Molter73: If the tests started with no configuration file and it was created afterwards, why would that be different to re-adding it here?

integration-tests/suites/k8s/config_reload.go
Molter73: K8s take the better part of a minute to propagate the configmap to pods, relative to that, this tick is fast and waiting 3 minutes is not excessive IMO.

integration-tests/suites/k8s/config_reload.go
Molter73: Dmitrii added a retry with timeout method, I will try using that instead to unify the assertions.

collector/lib/Inotify.cpp
```diff
@@ -0,0 +1,132 @@
...
+  FD_ZERO(&rfds);
+  FD_SET(fd_, &rfds);
+
+  int retval = select(fd_ + 1, &rfds, nullptr, nullptr, &tv);
+  if (retval < 0) {
+    return InotifyError("'select' call failed: (" + std::to_string(errno) + ") " + StrError());
+  }
+
+  if (retval == 0) {
+    return InotifyTimeout("'select' timedout");
```
Stringy, line 105: [nit]
```suggestion
    return InotifyTimeout("'select' timed out");
```

Stringy: also, another nit, the 'timed out' part of the message is a little redundant due to the type, so perhaps it can be dropped (or added to the class itself as a base template for the error)

Stringy (approved):
Looks good! just a couple of comments, but nothing especially contentious.

integration-tests/pkg/collector/collector_k8s.go
JoukoVirtanen: Sounds good.

integration-tests/pkg/collector/introspection.go
JoukoVirtanen: When there is an error you can return nil instead of body. That is the way that errors are usually handled in the stackrox repo.

integration-tests/suites/k8s/base.go
JoukoVirtanen: Doing it in a follow up PR is fine.

integration-tests/suites/k8s/config_reload.go
JoukoVirtanen: Okay

integration-tests/suites/k8s/config_reload.go
JoukoVirtanen: It is certainly possible that  a bug could lead to incorrect behavior in that case. E.g. collector might be listening to file creation events on startup, but not after the file is deleted.

collector/lib/Inotify.h
JoukoVirtanen: I thought this exact code was in `<sys/inotify.h>`. I checked `<sys/inotify.h> and that is not the case. I think a map should be used here, so that you don't get a mismatch between `BIT_MASK` and `MASK_STR`. I think it is easy to mess up the order here.

JoukoVirtanen: I thought you had copied `BIT_MASK` from `<sys/inotify.h>`, but actually looking at the file, I can see that is not the case.

integration-tests/pkg/collector/introspection.go
Molter73: I didn't know `nil` could be used with arrays and maps TIL, thanks!

collector/lib/Inotify.h
Molter73: I can now see how the comment may be misleading, I'll tweak it a bit to make it clearer we are creating the array from the defines in that header.

collector/lib/Inotify.h
```diff
@@ -0,0 +1,292 @@
...
+std::ostream& operator<<(std::ostream& os, const Watcher& w);
+
+/**
+ * Abstraction over the inotify subsystem.
+ */
+class Inotify {
+ public:
+  Inotify() : fd_(inotify_init()) {
+    if (fd_ < 0) {
+      CLOG(ERROR) << "Failed to initialize inotify: (" << errno << ") " << StrError();
```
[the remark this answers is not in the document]
Molter73: Correct, if the configuration is there at startup we'll read it and notify we are not using configuration reloading here:
https://github.com/stackrox/collector/blob/31d27e6bf7e843186a518bcfe924d307c6ceee84/collector/lib/ConfigLoader.cpp#L88-L91

collector/lib/ConfigLoader.cpp
```diff
@@ -0,0 +1,225 @@
...
+    return false;
+  } catch (const YAML::ParserException& e) {
+    CLOG(ERROR) << "Failed to parse the configuration file: " << config_file << ". Error: " << e.what();
+    return false;
+  }
+
+  return LoadConfiguration(config, node);
+}
+
+bool ConfigLoader::LoadConfiguration(CollectorConfig& config, const YAML::Node& node) {
```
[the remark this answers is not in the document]
Molter73: We've talked about this offline, but leaving it here for completeness.

Fully agree that reading a file and updating our configuration is something we want to abstract as much as possible, so far, we've used this handcrafted approach because we are only filling one field, but it is really clear this will escalate out of proportion very fast as soon as we start adding some more configuration with different types, etc..

I like the idea of doing schema validation, but I think at some point we are going to want to send this configuration over the wire, (at least that was the original intent and I don't see why it wouldn't resurface in the near future). This was the motivation for using protobuf for the definition and I think we can rely on it for versioning and compatibility. Going one step further, protobuf provides a reflection API that we could leverage to fill out the fields in an abstracted way, I have a PoC library for testing this, it's not feature complete, but it proves that it can be done: https://github.com/Molter73/config-much

integration-tests/suites/k8s/config_reload.go
Molter73: Jokes on me, there was a bug with this case.

integration-tests/suites/k8s/config_reload.go
Molter73: Turned out to be more of a hassle to use the retry method, did some refactoring to unify logic though.

JoukoVirtanen (approved):

PR #1911, opened 2024-10-25
Update gRPC to v1.67.0

Molter73:
## Description

Dependencies of gRPC updated:
- abseil: 20240722.0
- protobuf: v28.3

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [ ] Run e2e tests on the main repo.
- [ ] Run multi-arch builds.
- [ ] Check new version of protobuf deals with non-UTF8 characters correctly.


collector/test/UtilityTest.cpp
```diff
@@ -117,15 +117,20 @@ TEST(ExtractContainerIDFromCgroupTest, TestExtractContainerIDFromCgroup) {
...
-      "ab\x80"
-      "cd");
-  std::string expected_output("ab?cd");
+  using test_case = std::pair<std::string, std::string>;
+  std::vector<test_case> tests = {
+      {"ab\200cd", "ab?cd"},
+      {"ab\200", "ab?"},
+      {"\200ab\200", "?ab?"},
+      {"\200\200ab\200", "??ab?"},
+  };
```
ovalenti, line 126: [nit] I suggest that we keep a valid ascii-only string and a valid string with expanded codes in the test.

ovalenti (approved):
I had anticipated much more things to fix, and I am glad that I was wrong :tada:

collector/test/UtilityTest.cpp
Molter73: There already is a separate test for a valid ASCII string, I added a second case to that one with a text in Japanase, hopefully that is good enough.

PR #1913, opened 2024-10-25
build: Prevent OOM kills around Collector TAs

msugakov:
## Description

Occasionally, I saw `determine-image-tag` failing at `use-trusted-artifact` with OOM. I don't have a working link handy but I think this was the reason for https://console.redhat.com/application-pipeline/workspaces/rh-acs/applications/acs/pipelineruns/collector-build-gw9zk to fail because https://github.com/stackrox/collector/runs/31840930570 shows this:

image

Note the following in the failure snippet

> Using token for quay.io

it is what gets logged in `use-trusted-artifact` step, e.g.
image

There's something special about collector repo that it requires more memory for dealing with TAs.

Failures happen sporadically which I attribute to clusters getting loaded by other tenants and then containers get less memory than `limits` (because OOTB in Konflux `requests` < `limits`).

It's a bit of a question to @tommartensen how he figured that 3Gi is the right value in <https://github.com/stackrox/collector/pull/1843/files#diff-664d74698b59313a5b32d967d3257fc5beb60db5b7c7bc9c08877bddf3701540R62-R65>.

I ran `watch ... kubectl top pod ...` for running pods and saw `create-trusted-artifact` memory usage going over 2Gi.

image

I couldn't catch `use-trusted-artifact` memory needs this way so just assume 3Gi is not going to be too much.

Otherwise this change pretty straightforward: I applied the same settings to all `create-trusted-artifact`/`use-trusted-artifact` steps that currently exist in the pipeline.

## Checklist

**Automated testing**

No changes to automated testing.

## Testing Performed

1. Green Konflux CI.
2. I peeked at pods on the cluster to make sure TA step memory limits and requests are indeed 3Gi.

### DIFF
--- a/.tekton/collector-build.yaml
+++ b/.tekton/collector-build.yaml
@@ -55,14 +55,7 @@ spec:
       secretName: '{{ git_auth_secret }}'
 
   taskRunSpecs:
-  - pipelineTaskName: clone-repository
-    stepSpecs:
-    - name: create-trusted-artifact
-      computeResources:
-        limits:
-          memory: 3Gi
-        requests:
-          memory: 3Gi
+
   # Only adjusting computeResources for amd64 build because
   # multi-arch builds happen off cluster
   - pipelineTaskName: build-container-amd64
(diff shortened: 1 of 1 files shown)


tommartensen (approved):
Pointers! 

I cannot recall why I chose 3Gi - but if it works, it works.

msugakov:
Well, fedora-coreos is broken for both main ci and Konflux so I assume it's not me who broke it. I'll merge.

erthalion:
> fedora-coreos is broken for both main ci and Konflux

Right, it's a newish kernel version causes a verifier issue. Has nothing to do with this change.

PR #1942, opened 2024-11-05
Enable building with vcpkg

Molter73:
## Description

This enables building collector outside of the builder container by leveraging the [vcpkg](https://vcpkg.io) package manager. This is not ideal for CI, since vcpkg pulls from the internet so isolated builds are not possible, but it is more convenient than having to install all libraries to the system.

A simple CMakePresets.json file was added to enable even easier use. Command for using this build is: `cmake --preset=vcpkg`

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Manually checked building collector with vcpkg.
CI should check the regular build for collector is not broken.


Stringy (approved):
One minor non-required comment about the set up instructions, but otherwise looks great! I've tested this locally and aside from a long first build, it all works beautifully

PR #1945, opened 2024-11-07
ROX-24945: adds rate limiting of network events per container

Stringy:
## Description

This adds new rate limiting of network connections per container, on top of the existing afterglow and batching methods in order to provide a safety mechanism (upper limit) to the number of reported network connections for a given workload. When External IPs is enabled, the volume of connections will naturally be much higher, so this limit is primarily in service to controlling that volume.

Some caveats and decisions:
- The rate limiting is implemented at the last possible moment, as the batch message is being constructed. This is means we do not rate limit connections that would otherwise be dropped by afterglow. It also means that we're rate limiting the delta between the old state and the new state.
- Close events (those with a close timestamp) are not rate limited. Otherwise we could end up with a situation where a connection exists in Sensor's model, but its close event is never sent so that connection will exist forever.
- The time window for rate limiting is pegged to the scrape interval, to provide a logical and well-understood time frame for scale. This means we can also think of this rate limit as a constraint on the maximum size of batch we'll ever send to Sensor.
- The rate limiting is enabled regardless of external IPs. I think this makes the overall behaviour of network event reporting much clearer and more predictable, but I am not married to this notion and should we want to retain old behaviour for the vast majority of cases, we can change this.

Depends on: https://github.com/stackrox/stackrox/pull/13228

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


collector/lib/NetworkSignalHandler.cpp
```diff
@@ -1,10 +1,13 @@
 #include "NetworkSignalHandler.h"
 
+#include <chrono>
```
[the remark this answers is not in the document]
Stringy: We don't - removed

collector/lib/NetworkSignalHandler.cpp
```diff
@@ -147,6 +150,7 @@ SignalHandler::Result NetworkSignalHandler::HandleSignal(sinsp_evt* evt) {
     return SignalHandler::IGNORED;
   }
 
+  CLOG(DEBUG) << "Live event: " << result.value();
```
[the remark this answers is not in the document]
Stringy: Yeah this was just for manual verification - removed now.

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -1,10 +1,13 @@
 #include "NetworkStatusNotifier.h"
 
+#include <sstream>
```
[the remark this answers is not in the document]
Stringy: Removed

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -300,12 +303,40 @@ sensor::NetworkConnectionInfoMessage* NetworkStatusNotifier::CreateInfoMessage(c
 }
 
 void NetworkStatusNotifier::AddConnections(::google::protobuf::RepeatedPtrField<sensor::NetworkConnection>* updates, const ConnMap& delta) {
+  // Update the rate limiter with new bounds, in case the runtime
+  // configuration has been updated
+  connections_rate_limiter_.SetBurstSize(config_.PerContainerRateLimit());
+  connections_rate_limiter_.SetCapacity(config_.PerContainerRateLimit());
```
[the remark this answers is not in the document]
Stringy: Done, and I've also switched this around to create a new rate limiter directly in this function - due to the way we're rate limiting, we actually don't need the caching part of the RateLimitCache, so a new instance per call to AddConnections should be sufficient. There's certainly an argument that we could create a non-caching version of RateLimitCache, and I'm not completely opposed to that if it would be clearer. WDYT?

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -300,12 +303,40 @@ sensor::NetworkConnectionInfoMessage* NetworkStatusNotifier::CreateInfoMessage(c
...
+      // the case where we send a close event for a connection that it doesn't
+      // know about.
+      //
+      if (!connections_rate_limiter_.Allow(delta_entry.first.container())) {
+        CLOG_THROTTLED(INFO, std::chrono::seconds(5)) << "Rate limiting connections reported from container " << delta_entry.first.container();
+        continue;
+      }
     }
+
+    CLOG(DEBUG) << delta_entry.first;
```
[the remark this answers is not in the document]
Stringy: Removed

collector/lib/NetworkStatusNotifier.h
```diff
@@ -31,7 +32,8 @@ class NetworkStatusNotifier : protected ProtoAllocator<sensor::NetworkConnection
         config_(config),
         comm_(comm),
         connections_total_reporter_(connections_total_reporter),
-        connections_rate_reporter_(connections_rate_reporter) {
+        connections_rate_reporter_(connections_rate_reporter),
+        connections_rate_limiter_(RateLimitCache(config.PerContainerRateLimit(), config.PerContainerRateLimit(), config.ScrapeInterval())) {
```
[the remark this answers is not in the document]
Stringy: This was due to my ignorance about how these constructors worked - my bad! I've removed the connections_rate_limiter_ member now anyway so this problem has also been removed.

integration-tests/suites/network_stress.go
```diff
@@ -0,0 +1,76 @@
...
+	berserker_img := image_store.QaImageByKey("performance-berserker")
+
+	images := []string{
+		berserker_img,
+	}
+
+	for _, image := range images {
+		err := s.executor.PullImage(image)
+		s.Require().NoError(err)
+	}
```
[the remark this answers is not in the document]
Stringy: I've (temporarily) removed this test suite now, in favour of unit tests. I'll work on it on the side and add it back in in another PR.

collector/lib/NetworkStatusNotifier.cpp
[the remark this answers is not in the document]
Stringy: I kinda did it anyway, since it's pretty simple: https://github.com/stackrox/collector/pull/1945/commits/1ddc8aad57fad13aee87d511f8259c878fdcdb74

collector/test/NetworkStatusNotifierTest.cpp
```diff
@@ -337,6 +348,71 @@ TEST(NetworkStatusNotifier, UpdateIPnoAfterglow) {
...
+      {conn1, ConnStatus(1234, true)},
+      {conn2, ConnStatus(1234, true)},
+      {conn3, ConnStatus(1234, false)},
+      {conn4, ConnStatus(12345, false)},
+  };
+
+  google::protobuf::RepeatedPtrField<sensor::NetworkConnection>
+      updatesSingle,
+      updatesDuo,
+      updatesClose;
```
Molter73, line 403: [nitpick] I had a bit of problem understanding these are 3 separate variables of type `google::protobuf::RepeatedPtrField<sensor::NetworkConnection>`, probably a skill issue, so feel free to ignore this, but you could define a shorter name for the type and declare all three variables separately:
```suggestion
  using UpdatedConnections = google::protobuf::RepeatedPtrField<sensor::NetworkConnection>;
  UpdatedConnections updatesSingle;
  UpdatedConnections updatesDuo;
  UpdatedConnections updatesClose;
```

Molter73: Also, you could use an array here
```suggestion
  using UpdatedConnections = google::protobuf::RepeatedPtrField<sensor::NetworkConnection>;
  std::array<UpdatedConnections, 3> updatedConnections;
```

Molter73 (approved):
LGTM! Left a few comments but nothing major.

collector/test/NetworkStatusNotifierTest.cpp
```diff
@@ -337,6 +348,71 @@ TEST(NetworkStatusNotifier, UpdateIPnoAfterglow) {
...
+  // tracker delta, we expect only two to be returned
+  config.SetPerContainerRateLimit(2);
+  config.SetScrapeInterval(2);
+
+  NetworkStatusNotifier netStatusNotifier(conn_scraper, conn_tracker, comm, config);
+
+  Connection conn1("containerId", Endpoint(Address(10, 0, 1, 32), 1024), Endpoint(Address(192, 168, 0, 1), 1000), L4Proto::TCP, true);
+  Connection conn2("containerId", Endpoint(Address(10, 0, 1, 32), 1024), Endpoint(Address(192, 168, 0, 2), 1001), L4Proto::TCP, true);
+  Connection conn3("containerId", Endpoint(Address(10, 0, 1, 32), 1024), Endpoint(Address(192, 168, 0, 3), 1002), L4Proto::TCP, true);
+  Connection conn4("containerId", Endpoint(Address(10, 0, 1, 32), 1024), Endpoint(Address(192, 168, 0, 4), 1003), L4Proto::TCP, true);
```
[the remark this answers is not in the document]
Stringy: I've tried this out but I think I prefer the clarity of statically defining the connections and the deltas - manipulating vectors hides a bit of the intention behind each one. To make it all a bit clearer, though, I've statically define active and closed statuses so there's no more duplication in the deltas

collector/lib/CollectorConfig.h
```diff
@@ -196,6 +206,7 @@ class CollectorConfig {
   std::vector<double> connection_stats_quantiles_;
   double connection_stats_error_;
   unsigned int connection_stats_window_;
+  int64_t per_container_rate_limit_ = 1024;
```
Stringy, line 209: maybe make this a constant / env var

collector/lib/RateLimit.cpp
```diff
@@ -74,4 +74,19 @@ bool RateLimitCache::Allow(std::string key) {
...
 }
 
+CountLimiter::CountLimiter()
+    : count_(4096) {}
+
+CountLimiter::CountLimiter(int64_t count)
+    : count_(count) {}
+
+bool CountLimiter::Allow(std::string key) {
+  auto pair = cache_.emplace(std::make_pair(key, TokenBucket()));
```
JoukoVirtanen, line 84:
```suggestion
  auto pair = cache_.try_emplace(key);
```
This might be slightly more efficient as it only creates the `TokenBucket` if it needs to. `try_emplace` was introduced in C++ 17 so the other similar functions use `emplace` instead of `try_emplace`.

collector/lib/CollectorConfig.h
ovalenti: I wonder how to determine a good default value here. At first, 1024 seemed a small value to me, however, after afterglow, it already takes a quite specific workload to reach to that value under 30s.

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -300,12 +302,36 @@ sensor::NetworkConnectionInfoMessage* NetworkStatusNotifier::CreateInfoMessage(c
...
+      //
+      // We explicitly do not rate limit close events to avoid creation of
+      // zombie connections that have been reported to Sensor. Sensor can handle
+      // the case where we send a close event for a connection that it doesn't
+      // know about.
+      //
+      if (!rate_limiter.Allow(delta_entry.first.container())) {
+        CLOG_THROTTLED(INFO, std::chrono::seconds(5)) << "Rate limiting connections reported from container " << delta_entry.first.container();
+        continue;
+      }
```
ovalenti, line 332: What about a prometheus counter here ? Also, I wonder if we could display more info about the process, but then we would probably need to limit per container+process, which is not the plan :thinking:

Stringy: Good idea, I'll add a counter

collector/test/NetworkStatusNotifierTest.cpp
```diff
@@ -337,6 +348,75 @@ TEST(NetworkStatusNotifier, UpdateIPnoAfterglow) {
...
+  MockCollectorConfig config;
+  std::shared_ptr<MockConnScraper> conn_scraper = std::make_shared<MockConnScraper>();
+  auto conn_tracker = std::make_shared<ConnectionTracker>();
+  auto comm = std::make_shared<MockNetworkConnectionInfoServiceComm>();
+
+  // maximum of 2 connections per scrape interval
+  // if we throw four connections from the same container into the conn
+  // tracker delta, we expect only two to be returned
+  config.SetPerContainerRateLimit(2);
+  config.SetScrapeInterval(2);
```
ovalenti, line 361: I don't understand why we have to set the scrape-interval here.

ovalenti (approved):
Thanks for the stats counter. Overall, this looks good !

I tend to agree with @Molter73 on the fact that it would be easier to make sure that the runtime config always contains the value to use (even if it requires us to write the default values there).

collector/proto/third_party/stackrox
ovalenti, line 1: We also probably want to merge https://github.com/stackrox/stackrox/pull/13228 prior to merge this one.

Stringy:
> it would be easier to make sure that the runtime config always contains the value to use (even if it requires us to write the default values there).

Yes, I agree, though I don't think we can define the defaults in the proto itself (due to limitations with protobufs) so I think we'll need to enforce them somewhere. It might make sense to encapsulate that in Collector, but outside of the existing CollectorConfig class. I'll look into it in a follow up PR.

collector/lib/CollectorConfig.h
JoukoVirtanen: I think this is Okay for now. We can adjust after doing performance tests.

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -300,12 +302,37 @@ sensor::NetworkConnectionInfoMessage* NetworkStatusNotifier::CreateInfoMessage(c
 }
 
 void NetworkStatusNotifier::AddConnections(::google::protobuf::RepeatedPtrField<sensor::NetworkConnection>* updates, const ConnMap& delta) {
+  auto per_container_limit = config_.PerContainerRateLimit();
```
JoukoVirtanen, line 305: Nit. Use `inf64_t` instead of `auto`. I feel like `auto` obfuscates things.

JoukoVirtanen (approved):
LGTM!

PR #1954, opened 2024-11-12
Build bpftool from source

Molter73:
## Description

Downstream builds with ubi 8, which ships a bpftool that is old and doesn't work with the latest Falco changes on Arm and Z. In order to circumvent this, we can simply build from source.

## Checklist
- [x] Investigated and inspected CI test results
- [ ] Updated documentation accordingly

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

If any of these don't apply, please comment below.

## Testing Performed
CI is enough.

PR #1957, opened 2024-11-12
ROX-26784: Only trigger Konflux builds against master

kylape:
This is primarily intended to stop triggering builds in Konflux for "master" components when there are PRs or pushes on release branches.

.

### DIFF
--- a/.tekton/collector-build.yaml
+++ b/.tekton/collector-build.yaml
@@ -8,8 +8,7 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "500"
-    pipelinesascode.tekton.dev/on-cel-expression: |
-      (event == "push" && target_branch.matches("^(master|release-.*)$")) || event == "pull_request"
+    pipelinesascode.tekton.dev/on-cel-expression: target_branch == "master" && event in ["push", "pull_request"]
   labels:
     appstudio.openshift.io/application: acs
     appstudio.openshift.io/component: collector


tommartensen (approved):

Molter73:
Any way we could still have konflux pipelines on stacked PRs? I'm currently testing one and I need konflux to run (#1954)

msugakov:
+1 to Mauro's request. I don't see a reason to not allow for all PRs.

msugakov (changes requested):
+1 to Mauro's request. I don't see a reason to not allow for all PRs.

tommartensen:
> +1 to Mauro's request. I don't see a reason to not allow for all PRs.

it shouldn't run on PRs to release-* branches though, right?

msugakov:
> it shouldn't run on PRs to release-* branches though, right?

CEL is a [filter, not a routing mechanism](https://redhat-internal.slack.com/archives/C05TS9N0S7L/p1730824115817529?thread_ts=1730206231.947449&cid=C05TS9N0S7L). What happens in PRs to release branches is a function of what is in the commit that's being built: application and component for one, CEL filter defining whether to build or not build is another.

kylape:
Based on your comment @msugakov, I think we can simplify all the CEL expressions in the 6 PRs I currently have open. I'll try to do that now.

.tekton/collector-build.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "500"
-    pipelinesascode.tekton.dev/on-cel-expression: target_branch == "master" && event in ["push", "pull_request"]
+    pipelinesascode.tekton.dev/on-cel-expression: event in ["push", "pull_request"]
```
msugakov, line 11: Unfortunately, this needs one more iteration.
The reason is that we still need to have `target_branch` filter for `push` events. If we don't pushes to PR branches cause two builds: on `push` and one `pull_request`.
Builds should be identical because it's the same commit, the same pipeline and because we don't have any conditional logic on the even type. There could be potential issues from builds overwriting and/or appending duplicate data to the final artifacts but my concern is that it's just not the right thing to do.

It's hard to coordinate with that many PRs around. Let's sync on CEL expressions during today's meeting.

.tekton/collector-build.yaml
```diff
@@ -8,8 +8,14 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "500"
+    pipelinesascode.tekton.dev/on-cel-expression: event in ["push", "pull_request"]
     pipelinesascode.tekton.dev/on-cel-expression: |
```
msugakov, line 12: `pipelinesascode.tekton.dev/on-cel-expression` is duplicate now.

msugakov (changes requested):

msugakov (approved):

PR #1958, opened 2024-11-13
Remove krb5-libs from the final image.

ovalenti:
## Description

krb5 is not used by Collector and triggers vulnerability scanners. Remove it.

## Checklist


ovalenti:
Actually, krb5 is required :sweat_smile:

PR #1959, opened 2024-11-13
Add Rust toolchains and example rust library to Collector

Stringy:
## Description

Adds a Rust toolchain into Collector, allowing us to write Rust crates and call into them from C++. The idea is that we can gradually increase the amount of Rust in the project and take advantage of its helpful features including memory / thread safety.

### DIFF
--- a/collector/lib/rust/host/src/info.rs
+++ b/collector/lib/rust/host/src/info.rs
@@ -3,7 +3,8 @@ use log::{debug, info, warn};
 use std::io::Error;
 
 use libbpf_rs::libbpf_sys::{
-    libbpf_num_possible_cpus, libbpf_probe_bpf_prog_type, BPF_PROG_TYPE_TRACING,
+    libbpf_num_possible_cpus, libbpf_probe_bpf_map_type, libbpf_probe_bpf_prog_type,
+    BPF_MAP_TYPE_RINGBUF, BPF_PROG_TYPE_TRACING,
 };
 
 use crate::KernelVersion;
@@ -236,7 +237,20 @@ impl HostInfo {
     }
 
     pub fn has_bpf_ringbuf_support(&self) -> bool {
-        false
+        let res = unsafe { libbpf_probe_bpf_map_type(BPF_MAP_TYPE_RINGBUF, std::ptr::null()) };
+
+        if res == 0 {
+            info!(
+                "BPF ringbuffer map type is not available (errno = {:?})",
+                Error::last_os_error()
+            );
+        }
+
+        if res < 0 {
+            warn!("Unable to check for the BPF ringbuffer availability. Assuming it is available.");
+        }
+
+        res != 0
     }
 }
 


cmake/FindRust.cmake
Stringy, line 1: This is a slightly modified version of the cmake module taken from here: https://github.com/micahsnyder/cmake-rust-demo/blob/main/cmake/FindRust.cmake

The main addition is an include directory to pass to the build commands so we can centralise cbindgen headers

collector/lib/rust/test/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+#[no_mangle]
+pub extern "C" fn external_add(left: u64, right: u64) -> u64 {
+    left + right
```
Molter73, line 3:
```suggestion
    add(left, right)
```
This way would also test doing a call into a separate non-`extern "C"` method.

Stringy: Good idea - done!

collector/lib/rust/host/src/info.rs
```diff
@@ -0,0 +1,80 @@
...
+        self.os_release_key("ID")
+    }
+
+    pub fn build_id(&self) -> String {
+        self.os_release_key("BUILD_ID")
+    }
+
+    pub fn distro(&self) -> String {
+        self.os_release_key("PRETTY_NAME")
+    }
```
Molter73, line 57: There are the only 3 fields gathered from `/etc/os-release`, right? Wonder if we could change the HostInfo definition to something like this:
```rust
pub struct HostInfo {
    os: String,
    build: String,
    distro: Stirng,
}
```

Then, instead of reading the file every time the struct is accessed, you could read it once in a constructor method like:
```rust
    pub fn new<T: Into<PathBuf>>(path: T) -> HostInfo {
        let mut os = String::new();
        let mut build = String::new();
        let mut distro = String::new();

        if let Ok(file) = File::open(path.into()) {
            let reader = BufReader::new(file);

            for line in reader.lines() {
                match lines {
                    Ok(s) if s.starts_with("ID") => os = s["ID".len()..].to_owned(),
                    Ok(s) if s.starts_with("BUILD_ID") => build = s["BUILD_ID".len()..].to_owned(),
                    Ok(s) if s.starts_with("PRETTY_NAME") => distro = s["PRETTY_NAME".len()..].to_owned(),
                    Ok(_) => {} // Nothing to do with this line.
                    Err(_) => break,
                }
            }
        }
        HostInfo{os, build, distro}
    }

File: collector/lib/rust/host/src/info.rs [hunk 3]
```diff
@@ -0,0 +1,80 @@
...
+
+    fn os_release_key(&self, name: &str) -> String {
+        self.filter_by_key("/etc/os-release", name)
+    }
+
+    fn filter_by_key<T: Into<PathBuf>>(&self, path: T, name: &str) -> String {
+        if let Ok(file) = File::open(path.into()) {
```
Molter73, line 64: You can probably remove one level of indentation with a let-else here
```suggestion
    let Ok(file) = File::open(path.into()) else { return Default::default(); };
```

collector/lib/rust/host/src/kernel.rs
```diff
@@ -0,0 +1,79 @@
...
+                    captures[4].parse::<u64>().unwrap()
+                } else {
+                    0
+                },
+                release: release.to_owned(),
+                version: version.to_owned(),
+                machine: machine.to_owned(),
+            },
+            None => Default::default(),
+        }
```
Molter73, line 53: Just an idea here, maybe this will be a bit more readable
```rust
    pub fn new(release: String, version: String, machine: String) -> Self {
        // Regex definition goes here...
        let Some(captures) = re.captures(release} else { return Default::default(); };

        let kernel = captures[1].parse::<u64>().unwrap();
        let major = captures[2].parse::<u64>().unwrap();
        let minor = captures[3].parse::<u64>().unwrap();

        let build_id = if captures.len() > 4 {
            captures[4].parse::<u64>().unwrap()
        } else {
            0
        };

        KernelVersion{
            kernel,
            major,
            minor,
            build_id,
            release,
            version,
            machine,
        }
    }
```

collector/lib/rust/host/src/kernel.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    pub fn from_host() -> Box<KernelVersion> {
+        let kv = match uname() {
+            Ok(info) => KernelVersion::new(&info.release, &info.version, &info.machine),
```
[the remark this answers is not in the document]
Stringy: Neat! I forgot you could do this

collector/lib/rust/host/src/info.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    pub fn secure_boot_status(&self) -> SecureBootStatus {
+        SecureBootStatus::Unset
+    }
+}
+
+/// Constructs a path joined from the host root (usually /host)
+fn host_path<P: Into<PathBuf>>(path: P) -> PathBuf {
+    let root = PathBuf::from(std::env::var("COLLECTOR_HOST_PATH").unwrap_or("/host".to_string()));
+    root.join(path.into())
```
Molter73, line 218:
```suggestion
    let root = std::env::var("COLLECTOR_HOST_PATH").unwrap_or("/host".to_string());
    PathBuf::from(root).join(path.into())
```

collector/lib/rust/host/src/info.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+impl HostInfo {
+    pub fn new() -> Self {
+        let mut os = String::new();
+        let mut build = String::new();
+        let mut distro = String::new();
+        let mut hostname = String::new();
+
+        if let Ok(file) = File::open("/etc/os-release") {
```
Molter73, line 43: Do we need to `host_path` this as well?
```suggestion
        if let Ok(file) = File::open(host_path("/etc/os-release")) {
```

collector/lib/rust/host/src/info.rs
```diff
@@ -0,0 +1,224 @@
...
+            let reader = BufReader::new(file);
+
+            for line in reader.lines() {
+                match line {
+                    Ok(s) if s.starts_with("ID") => os = s["ID ".len()..].to_owned(),
```
Molter73, line 48: Now I'm wondering, do we need to add the `=` to these?
```suggestion
                    Ok(s) if s.starts_with("ID=") => os = s["ID=".len()..].to_owned(),
```

collector/lib/rust/host/src/info.rs
```diff
@@ -0,0 +1,224 @@
...
+                    Err(_) => break,
+                };
+            }
+        }
+
+        let hostname_paths = vec![
+            host_path("/etc/hostname"),
+            host_path("/proc/sys/kernel/hostname"),
+        ];
+        for path in hostname_paths {
```
Molter73, line 63: Do we need a vector? I would assume you can just use an array here
```suggestion
        for path in [
            host_path("/etc/hostname"),
            host_path("/proc/sys/kernel/hostname"),
        ] {
```

collector/lib/rust/host/src/info.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /// Whether or not this host machine is COS
+    pub fn is_cos(&self) -> bool {
+        self.os_id().eq("cos") && !self.build_id().is_empty()
+    }
+
+    /// Whether or not this host machine is CORE-OS
+    pub fn is_coreos(&self) -> bool {
+        self.os_id().eq("coreos")
```
Molter73, line 87: Could we use standard `==` in these checks?
```suggestion
        self.os_id() == "coreos"
```

collector/lib/rust/host/src/info.rs
```diff
@@ -0,0 +1,224 @@
...
+            ),
+            (
+                format!("/usr/lib/debug/boot/vmlinux-{}.debug", self.kernel.release),
+                true,
+            ),
+            (
+                format!("/usr/lib/debug/lib/modules/{}/vmlinux", self.kernel.release),
+                true,
+            ),
+        ];
```
Molter73, line 184:
```suggestion
        let release = self.kernel.release;
        let locations = vec![
            // try canonical vmlinux BTF through sysfs first
            ("/sys/kernel/btf/vmlinux".to_string(), false),
            // fall back to trying to find vmlinux on disk otherwise
            (format!("/boot/vmlinux-{release}"), false),
            (format!("/lib/modules/{release}/vmlinux-{release}"), false),
            (format!("/lib/modules/{release}/build/vmlinux"), false),
            (format!("/usr/lib/modules/{release}/kernel/vmlinux"), true),
            (format!("/usr/lib/debug/boot/vmlinux-{release}"), true,),
            (format!("/usr/lib/debug/boot/vmlinux-{release}.debug"), true),
            (format!("/usr/lib/debug/lib/modules/{release}/vmlinux"), true),
        ];
```

collector/lib/rust/host/src/info.rs
```diff
@@ -0,0 +1,224 @@
...
+            let path = if location.1 {
+                host_path(location.0)
+            } else {
+                PathBuf::from(location.0)
+            };
+
+            if path.exists() {
+                return true;
+            }
+        }
```
Molter73, line 196:
```suggestion
        for (path, on_host) in locations {
            let path = if on_host {
                host_path(path)
            } else {
                PathBuf::from(path)
            };

            if path.exists() {
                return true;
            }
        }
```

collector/lib/rust/host/src/info.rs
Molter73: You could also directly call `host_path` on the ones that need to be mapped and drop from a vector of tuples to a vector of paths.

collector/lib/rust/host/src/kernel.rs
```diff
@@ -0,0 +1,93 @@
...
+            build_id: 0,
+            release: "".to_string(),
+            version: "".to_string(),
+            machine: "".to_string(),
+        }
+    }
+}
+
+impl KernelVersion {
+    pub fn new(release: &str, version: &str, machine: &str) -> Self {
```
Molter73, line 34: Do we need these to be &str? Can't we use straight up String? Most places it looks like we already have those, it would save creating/copying strings.

PR #1963, opened 2024-11-18
ROX-18384: remove slim images from upstream

Stringy:
## Description

Since we have deprecated *full* images from Collector, we now want to drop the -slim tag for all images, since all image flavours are now inherently slim.

This PR removes upstream slim image tagging from Collector.

There are sibling PRs, which I will list here as they become available, that remove slim image use in other areas (downstream builds and the main repo)

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI should be enough.

### DIFF
--- a/ansible/ci-build-collector.yml
+++ b/ansible/ci-build-collector.yml
@@ -63,13 +63,6 @@
         push: true
         source: local
 
-    - name: Push slim to quay.io/stackrox-io
-      community.docker.docker_image:
-        name: "{{ collector_image }}-{{ arch }}"
-        repository: "{{ collector_image }}-{{ arch }}-slim"
-        push: true
-        source: local
-
     - name: Push base to quay.io/stackrox-io
       community.docker.docker_image:
         name: "{{ collector_image }}-{{ arch }}"
@@ -97,13 +90,6 @@
         push: true
         source: local
 
-    - name: Push slim to quay.io/rhacs-eng
-      community.docker.docker_image:
-        name: "{{ rhacs_eng_image }}-{{ arch }}"
-        repository: "{{ rhacs_eng_image }}-{{ arch }}-slim"
-        push: true
-        source: local
-
     - name: Push base to quay.io/rhacs-eng
       community.docker.docker_image:
         name: "{{ rhacs_eng_image }}-{{ arch }}"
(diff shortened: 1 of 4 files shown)


Molter73 (approved):
Small nit, we probably want to rename this job by dropping `slim` from it
image

Other than that, LGTM! Godspeed full images, you will not be missed.

ansible/ci-build-collector.yml
```diff
@@ -123,11 +109,9 @@
         msg:
         - "Pushed the following images:"
         - "  {{ collector_image }}-{{ arch }}"
-        - "  {{ collector_image }}-{{ arch }}-slim"
         - "  {{ collector_image }}-{{ arch }}-base"
         - "  {{ collector_image }}-{{ arch }}-latest"
```
msugakov, line 113: Not a blocking comment and not aiming to be delay the PR by becoming a reviewer.

I recalled this our conversation https://redhat-internal.slack.com/archives/CFMQ5C2TT/p1733309587336659

Your changes in https://github.com/stackrox/stackrox/pull/13350/files#diff-9f1c5124b529a407693ba184291c9797252978fe4ce8770e1ddec20d811efd69L104 allow getting rid of publishing `-latest`.

Also, I believe `-base` wasn't used by StackRox. IIRC, it was created by the Collector build pipelines as something intermediate before drivers are injected and it becomes "the full collector".

docs/how-to-start.md
```diff
@@ -25,7 +25,7 @@ $ make image
```
msugakov, line 25: (not about this line)
I suppose what some of my comments allude to is the need for more testing in different version combinations.
Such as:
- New Central old `roxctl`, can `roxctl` generate sensor manifests and can that be successfully deployed? In both versions of slim=true/false.
- Old Central, new `roxctl`, can sensor manifests be generated and successfully deployed?
- Get old manifest-based secured cluster connected to new Central (probably start both old and then upgrade only Central). Does the UI display all things as expected? Can the secured cluster download probes through Central?
- Get old Helm-based and Operator-based secured clusters connected to new Central. UI fine? Functioning ok?
- Get old Central and connect new Helm-based and Operator-based secured clusters. All works well, including UI?

PR #1967, opened 2024-11-20
Rate limiting fixes (config parsing, better logging)

Stringy:
## Description

So this is embarrassing. The rate limit wasn't parsed from the config and copied over to the local runtime config object, which meant that the rate limit was set to zero and no network connections were sent from Collector.

This fixes that. It also adds some more useful logging around rate limiting.

## Testing Performed

Tested locally on OCP 4 cluster.

### DIFF
--- a/collector/lib/ConfigLoader.cpp
+++ b/collector/lib/ConfigLoader.cpp
@@ -72,11 +72,18 @@ bool ConfigLoader::LoadConfiguration(CollectorConfig& config, const YAML::Node& 
   }
 
   bool enable_external_ips = external_ips_node["enable"].as<bool>(false);
+  int64_t per_container_rate_limit = networking["per_container_rate_limit"].as<int64_t>(1024);
+
   sensor::CollectorConfig runtime_config;
   runtime_config
       .mutable_networking()
       ->mutable_external_ips()
       ->set_enable(enable_external_ips);
+
+  runtime_config
+      .mutable_networking()
+      ->set_per_container_rate_limit(per_container_rate_limit);
+
   config.SetRuntimeConfig(std::move(runtime_config));
 
   CLOG(DEBUG) << "Runtime configuration:\n"
--- a/collector/lib/ConfigLoader.cpp
+++ b/collector/lib/ConfigLoader.cpp
@@ -59,29 +59,27 @@ bool ConfigLoader::LoadConfiguration(CollectorConfig& config, const YAML::Node& 
     CLOG(ERROR) << "Unable to read config from " << config_file;
     return false;
   }
-  YAML::Node networking = node["networking"];
-  if (!networking) {
+  YAML::Node networking_node = node["networking"];
+  if (!networking_node) {
     CLOG(DEBUG) << "No networking in " << config_file;
     return true;
   }
 
-  YAML::Node external_ips_node = networking["externalIps"];
+  YAML::Node external_ips_node = networking_node["externalIps"];
   if (!external_ips_node) {
     CLOG(DEBUG) << "No external IPs in " << config_file;
     return true;
   }
 
   bool enable_external_ips = external_ips_node["enable"].as<bool>(false);
-  int64_t per_container_rate_limit = networking["per_container_rate_limit"].as<int64_t>(1024);
+  int64_t per_container_rate_limit = networking_node["per_container_rate_limit"].as<int64_t>(1024);
 
   sensor::CollectorConfig runtime_config;
-  runtime_config
-      .mutable_networking()
+  auto* networking = runtime_config.mutable_networking();
+  networking
       ->mutable_external_ips()
       ->set_enable(enable_external_ips);
-
-  runtime_config
-      .mutable_networking()
+  networking
       ->set_per_container_rate_limit(per_container_rate_limit);
 
   config.SetRuntimeConfig(std::move(runtime_config));
--- a/collector/lib/NetworkStatusNotifier.cpp
+++ b/collector/lib/NetworkStatusNotifier.cpp
@@ -305,7 +305,16 @@ void NetworkStatusNotifier::AddConnections(::google::protobuf::RepeatedPtrField<
   int64_t per_container_limit = config_.PerContainerRateLimit();
   CountLimiter rate_limiter(per_container_limit);
 
+  CLOG(DEBUG) << "Per container rate limit: " << per_container_limit;
+
+  UnorderedMap<std::string, int> rate_limited_containers;
+
+  int64_t total_events = 0;
+  int64_t added_events = 0;
+
   for (const auto& delta_entry : delta) {
+    total_events++;
+
     auto* conn_proto = ConnToProto(delta_entry.first);
     if (!delta_entry.second.IsActive()) {
       *conn_proto->mutable_close_timestamp() = google::protobuf::util::TimeUtil::MicrosecondsToTimestamp(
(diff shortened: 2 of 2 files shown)


collector/lib/ConfigLoader.cpp
```diff
@@ -72,11 +72,18 @@ bool ConfigLoader::LoadConfiguration(CollectorConfig& config, const YAML::Node&
...
+
   sensor::CollectorConfig runtime_config;
   runtime_config
       .mutable_networking()
       ->mutable_external_ips()
       ->set_enable(enable_external_ips);
+
+  runtime_config
+      .mutable_networking()
+      ->set_per_container_rate_limit(per_container_rate_limit);
```
Molter73, line 85: [nitpick]
```suggestion
  sensor::CollectorConfig runtime_config;
  auto* networking = runtime_config.mutable_networking();
  networking
      ->mutable_external_ips()
      ->set_enable(enable_external_ips);

  networking->set_per_container_rate_limit(per_container_rate_limit);
```

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -305,7 +305,16 @@ void NetworkStatusNotifier::AddConnections(::google::protobuf::RepeatedPtrField<
   int64_t per_container_limit = config_.PerContainerRateLimit();
   CountLimiter rate_limiter(per_container_limit);
 
+  CLOG(DEBUG) << "Per container rate limit: " << per_container_limit;
```
Molter73, line 308: Do we need to log this every time we scrape? Would it make sense to print it when it is updated instead?

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -326,15 +335,27 @@ void NetworkStatusNotifier::AddConnections(::google::protobuf::RepeatedPtrField<
...
         continue;
       }
     }
 
+    added_events++;
     updates->AddAllocated(conn_proto);
   }
+
+  if (rate_limited_containers.size() != 0) {
+    for (auto container : rate_limited_containers) {
```
Molter73, line 353: If `rate_limited_containers` is empty, the for loop will skip on its own
```suggestion
  for (auto container : rate_limited_containers) {
```

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -326,15 +335,27 @@ void NetworkStatusNotifier::AddConnections(::google::protobuf::RepeatedPtrField<
       // the case where we send a close event for a connection that it doesn't
       // know about.
       //
-      if (!rate_limiter.Allow(delta_entry.first.container())) {
-        CLOG_THROTTLED(INFO, std::chrono::seconds(5)) << "Rate limiting connections reported from container " << delta_entry.first.container();
+      std::string container = delta_entry.first.container();
```
Molter73, line 338: I don't think we need to make a copy here, the call to `UnorderedMap::operator[]` will make it on its own if it's needed.
```suggestion
      const std::string& container = delta_entry.first.container();
```

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -305,7 +305,16 @@ void NetworkStatusNotifier::AddConnections(::google::protobuf::RepeatedPtrField<
   int64_t per_container_limit = config_.PerContainerRateLimit();
   CountLimiter rate_limiter(per_container_limit);
 
+  CLOG(DEBUG) << "Per container rate limit: " << per_container_limit;
+
+  UnorderedMap<std::string, int> rate_limited_containers;
```
Molter73, line 310: Since the keys used in this container are coming from the `delta` argument and they both live for the duration of this call, could we use `const std::string&` as the key and avoid additional copies?

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -326,15 +335,27 @@ void NetworkStatusNotifier::AddConnections(::google::protobuf::RepeatedPtrField<
...
       }
     }
 
+    added_events++;
     updates->AddAllocated(conn_proto);
   }
+
+  if (rate_limited_containers.size() != 0) {
+    for (auto container : rate_limited_containers) {
+      CLOG(INFO) << "Rate limited " << container.second << " connections from container " << container.first;
```
Molter73, line 354:
```suggestion
    for (const auto& [id, events] : rate_limited_containers) {
      CLOG(INFO) << "Rate limited " << events << " connections from container " << id;
```

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -326,15 +335,27 @@ void NetworkStatusNotifier::AddConnections(::google::protobuf::RepeatedPtrField<
...
+    }
+  }
+
+  CLOG(DEBUG) << "Processed " << total_events << " events; sending " << added_events;
```
Molter73, line 358: Since the total events are the same as  all the elements in the `delta` map and a `ConnMap` is just a renamed `std::unordered_map`, you don't really need `total_events`, you can just call `delta.size()`.
```suggestion
  CLOG(DEBUG) << "Processed " << delta.size() << " events; sending " << added_events;
```

Molter73 (commented):
Most of my comments are nitpicks, but some are about avoiding string copies, so I'd like them at least ackonwledged before approving.

collector/lib/NetworkStatusNotifier.cpp
Molter73: The important part in this one is using `const auto&` to prevent more copies from being created, the binding just add a bit more readability IMO.

collector/lib/NetworkStatusNotifier.cpp
Stringy: My thinking was to keep the log message close to any rate limiting messages, and to the total vs. sent message. I can probably incorporate it into either of those, though, instead of having this extra message.

collector/lib/NetworkStatusNotifier.cpp
Stringy: Done

collector/lib/NetworkStatusNotifier.cpp
Stringy: Done

collector/lib/NetworkStatusNotifier.cpp
Stringy: Done

collector/lib/NetworkStatusNotifier.cpp
Stringy: Done

collector/lib/NetworkStatusNotifier.cpp
Stringy: Done

collector/lib/NetworkStatusNotifier.cpp
Stringy: As discussed offline, gone with a string_view here due to limitations with references as map keys.

collector/lib/NetworkStatusNotifier.cpp
```diff
@@ -326,15 +330,25 @@ void NetworkStatusNotifier::AddConnections(::google::protobuf::RepeatedPtrField<
       // the case where we send a close event for a connection that it doesn't
       // know about.
       //
-      if (!rate_limiter.Allow(delta_entry.first.container())) {
-        CLOG_THROTTLED(INFO, std::chrono::seconds(5)) << "Rate limiting connections reported from container " << delta_entry.first.container();
+      const std::string& container = delta_entry.first.container();
```
Molter73, line 333: Could we use a `std::string_view` directly? I believe it's a trivial cast back to `const std::string&` so the call to `Allow` on the next line should still work.
```suggestion
      const std::string_view container = delta_entry.first.container();
```

PR #1969, opened 2024-11-21
Fixes SLES permissions errors when logging into quay.io

Stringy:
## Description

It looks like as a non-root user, the docker login command is attempting to create a new config file at `/etc/containers/config.json<string of numbers>` but it does not have permission to write there. Part of the problem is the docker installation on SLES, so in order to avoid this issue this PR installs and uses podman on SLES VMs.

## Checklist

## Testing Performed

Tested locally by provisioning a SLES 15 machine and reproducing the bug and testing installation of podman. CI will cover the remaining parts.


Molter73 (approved):
LGTM!

PR #1970, opened 2024-11-26
Renames per container rate limit, for consistency

Stringy:
## Description

Straightforward rename, plus some extra tests and docs.


ovalenti:
This change will have an impact on https://github.com/stackrox/stackrox/pull/13419

ovalenti (approved):
Since we did not publish this feature yet, it should be fine (and better doing it now than after).

Looks good to me.

JoukoVirtanen (approved):
LGTM!

PR #1987, opened 2024-12-09
ROX-26763: Enhance the handling of failures in procfsscraper

ovalenti:
## Description

The parsing of network connections and endpoints in procfsscraper can fail for several reasons, but at least one is not easy to address : short-lived processes can lead to a concurrent modification of the /proc directory struct.

We will lower the level of the logs produced by those failures (since connections are not reported in the end), but also add troubleshooting information to increase our knowledge of the failure causes.

- A new prometheus counter is incremented for every zombie process detected while iterating over connections. There is no log message at all in this case. (`
rox_collector_counters{type="procfs_zombie_process"} 800`)
- Trace-level messages are emitted when there is an error while parsing the structure of the procfs. These are no more throttled. The corresponding process state is now part of every such message. 

## Checklist

## Testing

/proc/<pid>/stat is known to bring some locking overhead. In order to evaluate the impact, the following was attempted:

- using on a local VM with linux 6.1
- created 1024 sleeping processes (netcat listening)
- spawn a script loop doing repetitive traffic with one of the netcats (10 messages /s)
- spawn concurrently connscraper in a loop (2 times per second). With either source taken from master (without the change) or from this branch.

Result:
- with the master version: %Cpu(s): 24.7 us, 31.1 sy,  0.0 ni, 43.1 id,  0.0 wa,  0.0 hi,  1.0 si,  0.0 st
- with the branch version: %Cpu(s): 25.8 us, 29.5 sy,  0.0 ni, 43.4 id,  0.3 wa,  0.0 hi,  1.0 si,  0.0 st

There does not seem to be an easily measurable penalty.


erthalion (approved):

ovalenti:
Integration tests for runtime-config changed on master, which broke Tekton builds. Will have to rebase, again.

PR #1990, opened 2024-12-10
Allow to run builder with podman

erthalion:
## Description

Currently running with podman without root is not possible, since selinux prevents mounting of the project source directory. Disable this limitation is enough.

## Checklist

## Testing Performed

Build Collector with builder using podman.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,23 @@ BUILD_BUILDER_IMAGE ?= false
 
 export COLLECTOR_VERSION := $(COLLECTOR_TAG)
 
+PODMAN =
+# docker --version might not contain any traces of podman in the latest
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,6 +8,23 @@ BUILD_BUILDER_IMAGE ?= false
 
 export COLLECTOR_VERSION := $(COLLECTOR_TAG)
 
+podman =
```
Stringy, line 11: [nit] can we capitalise this so it's distinguishable from the `podman` command?
```suggestion
PODMAN =
```

erthalion: Yep, why not.

Stringy (approved):

PR #1993, opened 2024-12-13
ROX-27396: Collector configloader should not need to be modified when the protobuf is

JoukoVirtanen:
## Description

Currently every time that CollectorConfig protobuf definition is changed, we have to change the code for the ConfigLoader. The changes here make it so that is not necessary.

There are also changes here that make it so that collector logs are written to the correct location. Currently for tests that have sub-tests, collector logs are only written to the directory for the first sub-test. This PR fixes that.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Unit and integration tests are sufficient.

### DIFF
--- a/collector/lib/ConfigLoader.cpp
+++ b/collector/lib/ConfigLoader.cpp
@@ -85,6 +85,11 @@ Json::Value yamlNodeToJson(const YAML::Node& yamlNode) {
 }
 
 bool ConfigLoader::LoadConfiguration(CollectorConfig& config, const YAML::Node& node) {
+  // We do multiple conversions here. First convert to JSON, then a JSON string, and then
+  // to the runtime config object. There are more direct ways to do the conversion.
+  // However, those methods make it difficult to handle the different data types.
+  // Here yamlNodeToJson converts everything to strings and JsonStringToMessage then
+  // correctly converts everything to the correct data type.
   const auto jsonConfig = yamlNodeToJson(node);
   Json::StreamWriterBuilder writer;
   std::string jsonStr = Json::writeString(writer, jsonConfig);
@@ -99,6 +104,9 @@ bool ConfigLoader::LoadConfiguration(CollectorConfig& config, const YAML::Node& 
   }
 
   config.SetRuntimeConfig(runtimeConfig);
+  config.RuntimeConfigHeuristics();
+  CLOG(INFO) << "Runtime configuration:\n"
+             << config.GetRuntimeConfigStr();
 
   return true;
 }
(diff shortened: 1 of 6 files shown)


collector/lib/ConfigLoader.cpp
```diff
@@ -52,41 +56,50 @@ bool ConfigLoader::LoadConfiguration(CollectorConfig& config) {
...
-    return true;
+bool ConfigLoader::LoadConfiguration(CollectorConfig& config, const YAML::Node& node) {
+  const auto jsonConfig = yamlNodeToJson(node);
+  Json::StreamWriterBuilder writer;
+  std::string jsonStr = Json::writeString(writer, jsonConfig);
+
+  sensor::CollectorConfig runtimeConfig;
+  google::protobuf::util::JsonParseOptions parseOptions;
+  parseOptions.ignore_unknown_fields = true;
+  auto status = google::protobuf::util::JsonStringToMessage(jsonStr, &runtimeConfig, parseOptions);
```
JoukoVirtanen, line 95: It seems silly to first convert to json, then json string, and then to the protobuf class, instead of converting more directly, and it likely is silly. The reason I did this is that this way I don't have to worry as much about the types. E.g. when converting from yaml, in this case, the code assumes that all the values are strings. This is later fixed by JsonStringToMessage. However, if we tried to convert to the protobuf class directly or the json class used by protobuf, we would first have to figure out the type before doing the conversion.

collector/test/ConfigLoaderTest.cpp
```diff
@@ -42,34 +42,16 @@ TEST(CollectorConfigTest, TestYamlConfigToConfigMultiple) {
   }
```
JoukoVirtanen, line 42: Ideally this would have the same behavior as before and no changes to the unit tests would have to be made. However, there the old code had checks that the networking field was present and not blank and with the new more general approach this check is not done. Also the old allowed for the yaml to have invalid values, but `JsonStringToMessage` does not allow that.

collector/lib/ConfigLoader.cpp
JoukoVirtanen: Add comments to explain this.

Molter73: You don't need to do multiple conversions, protobuf provides reflection so that you can fill in a message from anything at runtime, I know it's extra work, but I got most of it working in less that 200 LoC. Have you looked into https://github.com/Molter73/config-much/blob/main/src/internal/parser-yaml.cpp? I've proposed we moved into using something like that library before, have you consider something like it?

Molter73: @JoukoVirtanen, as we discussed offline, I've added enum support to my config-much library, let me know if this would be enough to use it in place of this approach.
https://github.com/Molter73/config-much/commit/e0b1a70c722e316c9a2a49f4cd4c69ec0115c7ad

PR #1997, opened 2024-12-20
fix: remove duktape from civetweb to remove debugger

davdhacs:
## Description

SAST scan showed the duktape dubugger as exploitable (https://cov01.lab.eng.brq2.redhat.com/osh/task/747071/log/rhacs-collector-container-4.6.0-1/scan-results-imp.html previous version ignored manually: https://gitlab.cee.redhat.com/osh/known-false-positives/-/blob/master/rhacs-collector-container/ignore.err?ref_type=heads). Can we remove duktape from the civetweb collector uses? (expecting CI tests to fail if collector relies on the embeeded duktape js engine). Then if it is not found in the scan, we will not need to keep ignoring it after civetweb version updates.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


erthalion:
@davdhacs We can add this to silence the scanner, but AFAICT this variable is set to off by default in civetweb [[1]]. Do you know by any chance why scanner still shows that as a problem?

[1]: https://github.com/civetweb/civetweb/blob/master/CMakeLists.txt#L221

PR #2005, opened 2025-01-07
Use newest clang and llvm

erthalion:
## Description

We pin clang and llvm versions due to bugs in certain versions, which we would like to skip (see [1], [2]). Bump them to 19.1.5 .

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2209635
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=2211472

## Checklist

## Testing Performed

Multi-arch CI pass is sufficient.

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -11,8 +11,8 @@ RUN dnf -y update \
         binutils-devel \
         bison \
         ca-certificates \
-        clang-17.0.6 \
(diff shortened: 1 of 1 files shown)


ovalenti (approved):
Glad to have master back to working condition, thanks a lot !

Currently, we can't verify the Konflux images; however, better not wait forever.

PR #2014, opened 2025-01-21
Cap parallelism of dependency building

Molter73:
## Description

When building the third party dependencies, we use the nproc command to try and get as many parallel builds as possible running at once. This can bring problems on systems with large numbers of cores and not enough ram to come with them, so for those we provide the option of manually setting the number of parallel builds with an environment variable.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Test the downstream build works as expected on affected systems.

### DIFF
--- a/builder/install/install-dependencies.sh
+++ b/builder/install/install-dependencies.sh
@@ -7,7 +7,7 @@ export LICENSE_DIR="/THIRD_PARTY_NOTICES"
 mkdir -p "${LICENSE_DIR}"
 
 export NPROCS
-NPROCS="$(nproc)"
+NPROCS="${NPROCS:-$(nproc)}"
 
 # shellcheck source=SCRIPTDIR/versions.sh
 source builder/install/versions.sh


builder/install/install-dependencies.sh
```diff
@@ -7,7 +7,7 @@ export LICENSE_DIR="/THIRD_PARTY_NOTICES"
 mkdir -p "${LICENSE_DIR}"
 
 export NPROCS
-NPROCS="$(nproc)"
+NPROCS="${NPROCS:-$(nproc)}"
```
erthalion, line 10: Let maybe not rely on nproc at all, and just have a constant default, 4 or something?

Molter73: The builder build is extermely slow as it is, if I can build it with nproc locally I want to be able to do that. This way we can just cap it on the systems that are giving us trouble, do you see any problems with this approach?

Molter73: Also, GHA runners have just 2 vCPUs if I recall correctly, so capping it to 4 also doesn't sound quite right.

erthalion (approved):

PR #2025, opened 2025-02-03
Use protobuf reflection for loading runtime configuration

Molter73:
## Description

This change is based upon a toy project I have at https://github.com/Molter73/config-much.

The basic idea is to use the introspection and reflection APIs from protobuf to populate the runtime configuration structures, which should allow us to add new fields to our protobuf definitions and have them magically show up in our code to use.

This is an alternative implementation to #1993.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI should be enough to ensure no behavioral changes have been introduced.

### DIFF
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -140,7 +140,7 @@ int main(int argc, char** argv) {
 
   CollectorConfig config;
   config.InitCollectorConfig(args);
-  if (!ConfigLoader::LoadConfiguration(config)) {
+  if (!ConfigLoader(config).LoadConfiguration()) {
     CLOG(FATAL) << "Unable to parse configuration file";
   }
 
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -140,12 +140,8 @@ int main(int argc, char** argv) {
 
   CollectorConfig config;
   config.InitCollectorConfig(args);
-  switch (ConfigLoader(config).LoadConfiguration()) {
-    case collector::ConfigLoader::SUCCESS:
-    case collector::ConfigLoader::FILE_NOT_FOUND:
-      break;
-    case collector::ConfigLoader::PARSE_ERROR:
-      CLOG(FATAL) << "Unable to parse configuration file";
+  if (ConfigLoader(config).LoadConfiguration() == collector::ConfigLoader::PARSE_ERROR) {
+    CLOG(FATAL) << "Unable to parse configuration file";
   }
 
   setCoreDumpLimit(config.IsCoreDumpEnabled());
--- a/collector/collector.cpp
+++ b/collector/collector.cpp
@@ -140,8 +140,12 @@ int main(int argc, char** argv) {
 
   CollectorConfig config;
   config.InitCollectorConfig(args);
-  if (!ConfigLoader(config).LoadConfiguration()) {
-    CLOG(FATAL) << "Unable to parse configuration file";
+  switch (ConfigLoader(config).LoadConfiguration()) {
+    case collector::ConfigLoader::SUCCESS:
+    case collector::ConfigLoader::FILE_NOT_FOUND:
+      break;
+    case collector::ConfigLoader::PARSE_ERROR:
+      CLOG(FATAL) << "Unable to parse configuration file";
   }
 
   setCoreDumpLimit(config.IsCoreDumpEnabled());
--- a/collector/lib/ConfigLoader.cpp
+++ b/collector/lib/ConfigLoader.cpp
@@ -1,5 +1,7 @@
 #include "ConfigLoader.h"
 
+#include <algorithm>
+
 #include <google/protobuf/descriptor.h>
 
 #include "internalapi/sensor/collector.pb.h"
@@ -121,12 +123,15 @@ ParserResult ParserYaml::ParseArrayEnum(google::protobuf::Message* msg, const YA
 
   const EnumDescriptor* desc = field->enum_type();
   for (const auto& n : node) {
-    auto v = TryConvert<std::string_view>(n);
+    auto v = TryConvert<std::string>(n);
     if (IsError(v)) {
       errors.emplace_back(std::get<ParserError>(v));
       continue;
     }
-    const auto enum_name = std::get<std::string_view>(v);
+    auto enum_name = std::get<std::string>(v);
+    std::transform(enum_name.begin(), enum_name.end(), enum_name.begin(), [](char c) {
+      return std::toupper(c);
+    });
 
     const EnumValueDescriptor* value = desc->FindValueByName(enum_name);
     if (value == nullptr) {
@@ -331,7 +336,12 @@ ParserResult ParserYaml::Parse(google::protobuf::Message* msg, const YAML::Node&
       std::cerr << "Unsupported type BYTES" << std::endl;
       break;
     case FieldDescriptor::TYPE_ENUM: {
-      const auto enum_name = node[*name].as<std::string_view>();
+      auto enum_name = node[*name].as<std::string>();
+
+      // We assume enum definitions use UPPER_CASE nomenclature, so
+      std::transform(enum_name.begin(), enum_name.end(), enum_name.begin(), [](char c) {
+        return std::toupper(c);
+      });
 
       const EnumDescriptor* descriptor = field->enum_type();
       const EnumValueDescriptor* value = descriptor->FindValueByName(enum_name);
--- a/collector/lib/ConfigLoader.cpp
+++ b/collector/lib/ConfigLoader.cpp
@@ -164,17 +164,17 @@ ParserResult ParserYaml::FindUnkownFields(const google::protobuf::Message& msg, 
 
     const FieldDescriptor* field = descriptor->FindFieldByName(name);
     if (field == nullptr) {
-      std::stringstream ss;
-      ss << "Unknown field '" << name << "'";
-      errors.emplace_back(ss.str());
+      ParserError err;
+      err << "Unknown field '" << name << "'";
+      errors.emplace_back(err);
       continue;
     }
 
     if (it->second.IsMap()) {
       if (field->type() != FieldDescriptor::TYPE_MESSAGE) {
-        std::stringstream ss;
-        ss << file_ << ": Invalid type '" << NodeTypeToString(it->second.Type()) << "' for field " << it->first.as<std::string_view>() << ", expected '" << field->type_name() << "'";
-        errors.emplace_back(ss.str());
+        ParserError err;
+        err << file_ << ": Invalid type '" << NodeTypeToString(it->second.Type()) << "' for field " << it->first.as<std::string_view>() << ", expected '" << field->type_name() << "'";
+        errors.emplace_back(err);
         continue;
       }
 
@@ -393,17 +393,17 @@ ParserResult ParserYaml::ParseArray(google::protobuf::Message* msg, const YAML::
       return {{"Unsupport repeated type MESSAGE"}};
     } break;
     default: {
-      std::stringstream ss;
-      ss << "Unknown type " << field->type_name();
-      return {{ss.str()}};
+      ParserError err;
+      err << "Unknown type " << field->type_name();
+      return {{err}};
     }
   }
 }
 
 ParserError ParserYaml::WrapError(const std::exception& e) {
-  std::stringstream ss;
-  ss << file_ << ": " << e.what();
-  return ss.str();
+  ParserError err;
+  err << file_ << ": " << e.what();
+  return err;
 }
 
 template <typename T>
--- a/collector/lib/ConfigLoader.cpp
+++ b/collector/lib/ConfigLoader.cpp
@@ -457,7 +457,7 @@ void ConfigLoader::Stop() {
 }
 
 ConfigLoader::Result ConfigLoader::LoadConfiguration(const std::optional<const YAML::Node>& node) {
-  sensor::CollectorConfig runtime_config;
+  sensor::CollectorConfig runtime_config = NewRuntimeConfig();
   ParserResult errors;
 
   if (!node.has_value()) {
@@ -482,6 +482,15 @@ ConfigLoader::Result ConfigLoader::LoadConfiguration(const std::optional<const Y
   CLOG(DEBUG) << "Runtime configuration:\n"
               << config_.GetRuntimeConfigStr();
   return SUCCESS;
+}
+
+sensor::CollectorConfig ConfigLoader::NewRuntimeConfig() {
+  sensor::CollectorConfig runtime_config;
+
+  // Set default values that are different from the protobuf defaults
+  runtime_config.mutable_networking()->set_max_connections_per_minute(CollectorConfig::kMaxConnectionsPerMinute);
+
+  return runtime_config;
 }
 
 void ConfigLoader::WatchFile() {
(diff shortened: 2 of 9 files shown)


collector/collector.cpp
```diff
@@ -140,8 +140,12 @@ int main(int argc, char** argv) {
 
   CollectorConfig config;
   config.InitCollectorConfig(args);
-  if (!ConfigLoader::LoadConfiguration(config)) {
-    CLOG(FATAL) << "Unable to parse configuration file";
+  switch (ConfigLoader(config).LoadConfiguration()) {
```
JoukoVirtanen, line 143:
```suggestion
  if (ConfigLoader(config).LoadConfiguration()== collecotr::ConfigLoader::PARSE_ERROR) {
```

collector/lib/ConfigLoader.cpp
```diff
@@ -13,109 +19,513 @@ enum PathTags {
...
+      return ParseArrayInner<float>(msg, node, field);
+    case google::protobuf::FieldDescriptor::CPPTYPE_BOOL:
+      return ParseArrayInner<bool>(msg, node, field);
+    case google::protobuf::FieldDescriptor::CPPTYPE_ENUM:
+      return ParseArrayEnum(msg, node, field);
+    case google::protobuf::FieldDescriptor::CPPTYPE_STRING:
+      return ParseArrayInner<std::string>(msg, node, field);
+    case google::protobuf::FieldDescriptor::CPPTYPE_MESSAGE: {
+      return {{"Unsupport repeated type MESSAGE"}};
+    } break;
```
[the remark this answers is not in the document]
Molter73: I'm also going to space them out a bit, it's a little hard to read otherwise.

JoukoVirtanen:
Thanks for doing this. This is a lot of work and you have added good unit tests, but I feel that given the amount of new code, the number of unit tests is insufficient. The unit tests call `Parse`, but there are many new functions that are introduced. I would have a lot more confidence if more of these new functions were directly tested. For example `FindUnkownFields` (Fix the typo), is not tested and none of the new unit tests, have unknown fields that are checked for.

collector/lib/ConfigLoader.h
```diff
@@ -1,14 +1,167 @@
...
+                           const google::protobuf::FieldDescriptor* field, const std::string& name);
+  /**
+   * Go through all nodes in the configuration and notify of any
+   * elements that have no corresponding field in the protobuf message.
+   *
+   * @param msg The protobuf message used for validation.
+   * @param node The YAML::Node to be walked.
+   * @returns an optional vector of parser errors.
+   */
+  ParserResult FindUnkownFields(const google::protobuf::Message& msg, const YAML::Node& node);
```
JoukoVirtanen, line 133:
```suggestion
  ParserResult FindUnknownFields(const google::protobuf::Message& msg, const YAML::Node& node);
```

JoukoVirtanen: I make the same typo all the time.

collector/lib/ConfigLoader.cpp
```diff
@@ -8,114 +14,520 @@ namespace collector {
...
+  for (int i = 0; i < descriptor->field_count(); i++) {
+    const FieldDescriptor* field = descriptor->field(i);
+
+    auto err = Parse(msg, node, field);
+    if (err) {
+      errors.insert(errors.end(), err->begin(), err->end());
+    }
+  }
+
+  auto res = FindUnkownFields(*msg, node);
```
JoukoVirtanen, line 75: Do we want to treat unknown fields in the same way as other errors. What if a user reverts to an older version, but keeps the same ConfigMap? If the new version introduced new fields, this would now cause an error.

collector/lib/ConfigLoader.cpp
```diff
@@ -8,114 +14,520 @@ namespace collector {
...
+    case YAML::NodeType::Undefined:
+      return "Undefined";
+    case YAML::NodeType::Scalar:
+      return "Scalar";
+    case YAML::NodeType::Sequence:
+      return "Sequence";
+    case YAML::NodeType::Map:
+      return "Map";
+  }
+  return "";  // Unreachable
```
JoukoVirtanen, line 37:
```suggestion
  return "Unknown"; 
```

I was able to reach your unreachable code using
```
CLOG(INFO) << "NodeTypeToString= " << NodeTypeToString(static_cast<YAML::NodeType::value>(17));
```

The output was

[INFO    2025/02/13 02:22:43] NodeTypeToString=

collector/lib/ConfigLoader.cpp
JoukoVirtanen: It might be better just to warn about unknown fields.

JoukoVirtanen:
In some of the existing tests you could change `ENABLED` or `DISABLED` to lower case.

Molter73:
> This is a lot of work and you have added good unit tests, but I feel that given the amount of new code, the number of unit tests is insufficient.

I'm adding code coverage to the repo as we speak, the unit tests I added should cover most, if not all of the important branches.

> The unit tests call `Parse`, but there are many new functions that are introduced. I would have a lot more confidence if more of these new functions were directly tested.

What you are describing here is called white box testing and, while it may be useful in some cases, I prefer the black box testing approach that I took in the unit tests I added. The tests I added only call `Parse` which is the only public method the new `ParserYaml` class has, but internally, all of the added methods are called from there.

> For example `FindUnkownFields` (Fix the typo), is not tested and none of the new unit tests, have unknown fields that are checked for.

I thought I added a test for this, guess I missed it, I'll add it with my next commits.

collector/lib/ConfigLoader.cpp
Molter73: This would require `ParserYaml` to have a notion of warning vs. error, it would require it to return something like a `std::pair<ParserError, ParserError>` or implement a `ParserError` with an error level. While this is achievable, it might be overkill and it will require some more changes and complexity to how `ParserYaml` itself is configured. Are you OK to keep things as are for now and we can tackle it in a follow up?

Side note, I have implemented a method to make errors ignorable in config-much itself, but I would prefer to keep us from escalating the scope of this PR even further.

- https://github.com/Molter73/config-much/commit/80beeb90391a1d7658f9ad16eae6a65b3130c6e2

collector/lib/ConfigLoader.cpp
Molter73: This is odd, I'll have to try it, I'm fairly certain I covered all the cases for `YAML::NodeType`.

JoukoVirtanen: I think I am just being a bit nit picky. I think in practice it is unlikely that part of the code would be reached, but it is possible that either a new YAML::NodeType enum is introduced or someone writes some strange code like I did. However, I prefer that the final return be something other than an empty string and I would not comment code as being unreachable unless it was inside something like `if (false)`. But then why would I have that?

collector/lib/ConfigLoader.cpp
JoukoVirtanen: I think it could be done in a follow up PR, but I have a strong desire for that to be done before 4.8. Can you write the ticket? Thanks.

collector/lib/ConfigLoader.cpp
```diff
@@ -8,114 +14,520 @@ namespace collector {
...
-    enable_external_ips = sensor::ExternalIpsEnabled::DISABLED;
   } else {
-    CLOG(WARNING) << "Unknown value for for networking.externalIps.enabled. Setting it to DISABLED";
-    enable_external_ips = sensor::ExternalIpsEnabled::DISABLED;
+    errors = parser_.Parse(&runtime_config, *node);
   }
-  int64_t max_connections_per_minute = networking_node["maxConnectionsPerMinute"].as<int64_t>(CollectorConfig::kMaxConnectionsPerMinute);
 
+  if (errors) {
+    CLOG(ERROR) << "Failed to parse " << parser_.GetFile();
```
JoukoVirtanen, line 492: Since collector has multiple threads it is possible that this line will be logged, then other unrelated logging statements will be created, and then the errors in the loop will be logged, again possibly interspersed by unrelated logging statements. I think it would be best to save the error messages to one string and then log that string.

collector/lib/ConfigLoader.h
```diff
@@ -1,14 +1,167 @@
...
+    return os;
+  }
+
+  friend bool operator==(const ParserError& lhs, const ParserError& rhs) { return lhs.msg_ == rhs.msg_; }
+
+ private:
+  std::string msg_;
+};
+
+using ParserResult = std::optional<std::vector<ParserError>>;
```
JoukoVirtanen, line 50: After going through the code I still don't like this. I think it is simpler to not have optional. You have some if statements checking the optional, before inserting one list of errors into another. This check is not necessary if you just use a vector. Also I think there is only one place where `empty()` is needed if optional is not used. I created a branch where I made this change and the other changes that need to be made to get everything to work with just the plain vector. Please take a look and merge it into your branch if you like it. https://github.com/stackrox/collector/pull/2042

collector/lib/ConfigLoader.cpp
```diff
@@ -8,114 +14,520 @@ namespace collector {
...
+ParserResult ParserYaml::Parse(google::protobuf::Message* msg, const YAML::Node& node,
+                               const google::protobuf::FieldDescriptor* field) {
+  using namespace google::protobuf;
+
+  std::string camel;
+  const std::string* name = &field->name();
+  if (read_camelcase_) {
+    camel = SnakeCaseToCamel(*name);
+    name = &camel;
+  }
```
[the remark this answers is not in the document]
Stringy: Pretty sure the pointer is to avoid unneeded string copies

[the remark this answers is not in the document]
JoukoVirtanen: It seems like overkill when this won't get called very often.

JoukoVirtanen:
> I'm adding code coverage to the repo as we speak, the unit tests I added should cover most, if not all of the important branches.

I approve.

> The tests I added only call Parse which is the only public method the new ParserYaml class has, but internally, all of the added methods are called from there.

I realize the other methods will be called as a result of calling Parse, but many of them are complicated enough that they deserve tests of their own.

collector/lib/ConfigLoader.cpp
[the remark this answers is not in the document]
JoukoVirtanen: Keep it the way it is.

collector/lib/ConfigLoader.cpp
Stringy: I wonder if there's any benefit to officially supporting that in the logging library (presumably it can be done by adding a specialisation of the `<<` operator?)

collector/lib/ConfigLoader.cpp
Molter73: I didn't read this comment properly the first time around. This is not the way to use an enum, the whole reason enums exist is to give a limited, well defined set of values that can be used at any given time, if you are going to break that invariant then there's no code I can write that will satisfy every possible condition anyone could come up with. 

If you use the enum the way it is intended to be used, then that last line is effectively unreachable because all possible values have been accounted for in the switch-case: https://github.com/jbeder/yaml-cpp/blob/f7320141120f720aecc4c32be25586e7da9eb978/include/yaml-cpp/node/type.h#L11-L13

Now, why is the last `return` value there? Because the compiler will complain that the function reaches the end without a return statement because it is not smart enough to figure out that it is unreachable. Could I use a `default` branch in the switch case? Yes, but the advantage of doing it this way is that if you have proper linting setup for the project, (which I've wanted to implement for some time now) and you miss a value, you'll see something like this:
image

This in turn means that if we update `yaml-cpp` in the future and an enum value is added, we would know and could act on it. If the `default` case is added though, the lint check goes away, defeating the purpose.

collector/lib/ConfigLoader.h
Molter73: Answered in the PR.

TL;DR the way used in this PR is more idiomatic C++17

Molter73:
> I realize the other methods will be called as a result of calling Parse, but many of them are complicated enough that they deserve tests of their own.

I wholeheartedly disagree, most of the methods might seem complicated but are actually quite straightforward if you read them in isolation, so please provide concrete examples so we can discuss them properly.

In the end, if most of the code is tested and the output of the parsing operation is correct, what would be the benefit of testing the internals of the class? The only thing that would arise from adding such tests in my experience is that if we decide we want to change the internal implementation in the future a bunch of tests will need to be rewritten to fit, adding a lot of additional work for no added value.

collector/lib/ConfigLoader.cpp
Molter73: I'll look into it, the logging library is purposefully unsafe w.r.t threads because you don't want to be blocking proper operations with log messages. I'm also not a big fan of copying a bunch of strings around, so in order for me to be onboard with this, I'll need to figure out whether there's a way to do it by moving strings around, but there probably isn't.

collector/lib/ConfigLoader.cpp
```diff
@@ -8,114 +14,520 @@ namespace collector {
...
+      out += c;
+    }
+    capitalize = false;
+  }
+
+  return out;
+}
+
+std::string ParserYaml::CamelCaseToSnake(const std::string& s) {
+  std::string out;
```
JoukoVirtanen, line 450:
```suggestion
  std::string out;
  out.reserve(s.size());
```
Might avoid reallocation in some cases.

collector/lib/ConfigLoader.cpp
```diff
@@ -8,114 +14,520 @@ namespace collector {
...
+    return node.as<T>();
+  } catch (YAML::InvalidNode& e) {
+    return WrapError(e);
+  } catch (YAML::BadConversion& e) {
+    return WrapError(e);
   }
+}
+
+std::string ParserYaml::SnakeCaseToCamel(const std::string& s) {
+  std::string out;
```
JoukoVirtanen, line 429:
```suggestion
  std::string out;
  out.reserve(s.size());
```
Might avoid reallocation in some cases.

collector/lib/ConfigLoader.cpp
```diff
@@ -8,114 +14,520 @@ namespace collector {
...
+
+  return out;
+}
+
+std::string ParserYaml::CamelCaseToSnake(const std::string& s) {
+  std::string out;
+  bool first = true;
+
+  for (const auto& c : s) {
+    if (!first && std::isupper(c) != 0) {
```
JoukoVirtanen, line 454:
```suggestion
    if (!first && std::isupper(c)) {
```

JoukoVirtanen:
This is good work, and while I disagree with some things, I don't think they should be blockers to merging this. I have not seen any logical errors, and tests have been added beyond the ones that were originally created to test the feature. I added a couple of small suggestions, that are not needed. I can approve without those changes. The one thing I would like to see before approving, is to change the capitalization of the enum in some of the existing tests.

collector/lib/ConfigLoader.cpp
Molter73: I ended up adding the code to make it so unknown fields are ignored.

JoukoVirtanen (approved):
LGTM!

Stringy (approved):

PR #2028, opened 2025-02-07
ROX-18384: removes slim and latest images from konflux build

Stringy:
## Description

Following #1963 and https://github.com/stackrox/stackrox/pull/13350 we no longer require -latest or -slim images for Collector


tommartensen (approved):
LGTM, though not sure what `latest` was used for.

msugakov:
Now need to spoon-feed `/retest collector-on-push` a few times to rule out flakes and make sure the updated task passes.

msugakov (approved):
Approving to not block the merge once CI is green.

PR #2039, opened 2025-02-12
Add arm64 to the test matrix for the downstream tests

Molter73:
## Description

As the title specifies, we've started building Arm based images downstream, so we might as well extend our test matrix to include them.

This also fixes a discrepancy in naming now that `-slim` is no longer used.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] CI with the `run-cpaas-steps` once the 0.1 has a build with Arm images.

### DIFF
--- a/.github/workflows/cpaas.yml
+++ b/.github/workflows/cpaas.yml
@@ -64,6 +64,7 @@ jobs:
           - amd64
           - s390x
           - ppc64le
+          - arm64
     steps:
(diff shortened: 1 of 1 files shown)


ovalenti (approved):
quay has all the images and the cpaas related tests pass :tada: 

LGTM !

PR #2042, opened 2025-02-13
Remove unneeded optional from ParserResult

JoukoVirtanen:
## Description

Removes optional from `using ParserResult = std::optional<std::vector<ParserError>>;` and makes other changes so that things are consistent.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Unit tests should be sufficient.

### DIFF
--- a/collector/lib/ConfigLoader.cpp
+++ b/collector/lib/ConfigLoader.cpp
@@ -1,5 +1,7 @@
 #include "ConfigLoader.h"
 
+#include <algorithm>
+
 #include <google/protobuf/descriptor.h>
 
 #include "internalapi/sensor/collector.pb.h"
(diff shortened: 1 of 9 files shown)


collector/lib/ConfigLoader.cpp
```diff
@@ -488,9 +480,9 @@ ConfigLoader::Result ConfigLoader::LoadConfiguration(const std::optional<const Y
     errors = parser_.Parse(&runtime_config, *node);
   }
 
-  if (errors) {
+  if (!errors.empty()) {
```
Molter73, line 483: I don't really like this, `if (errors) {` is a much more idiomatic way to express that you will execute something if errors happened when compared to `if (!errors.empty()) {`. Performance wise, both options should be the same, because we create vectors in the inner methods, so not much difference there and this code is not in any hot path, so I wouldn't worry about it.

Molter73 (commented):
Hey @JoukoVirtanen! Thanks for putting this PR together.

I appreciate that removing the `std::optional` from `ParserResult` simplifies access to it very slightly, but as I mentioned in my other comment, it does so by being less idiomatic C++17 and making it harder to convey its intention, checking if there are errors is a lot easier to read with `if (errors)` than `if (!errors.empty())`, which is a C++11 idiom that was used because there was no better alternative.

With that said, if it's up to me, I wouldn't move forward with this change.

PR #2043, opened 2025-02-14
Add note about GRPC_TRACE

erthalion:
## Description

We've struggled in the past with connectivity issues and lack of information from the GRPC library, e.g. it's hard to distinguish a timeout error from a certificate error -- Collector will just say "can't connect" in both cases.

Add a note about useful workaround for that with GRPC_TRACE environment variable into the troubleshooting section.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -609,11 +609,11 @@ $ curl "http://<collector>:8080/state/runtime-config"
 
 ## Troubleshooting GRPC channel
 
-I case of network connectivity issues it can be hard to figure out the reason
-why Collector refuses to connect to Sensor. The reason for that is use
-WaitForConnected GRPC method, which doesn't betray much information. A
-workaround to get more information is to add `GRPC_TRACE='*'` environment
-variable to the DaemonSet, which will ask the GRPC library to trace all if it's
+In case of network connectivity issues it can be hard to figure out the reason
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -607,3 +607,18 @@ $ curl "http://<collector>:8080/state/network/connection?container=c6f030bc4b42&
 The runtime configuration can be obtained using
 $ curl "http://<collector>:8080/state/runtime-config"
 
+## Troubleshooting GRPC channel
+
+I case of network connectivity issues it can be hard to figure out the reason
+why Collector refuses to connect to Sensor. The reason for that is use
+WaitForConnected GRPC method, which doesn't betray much information. A
+workaround to get more information is to add `GRPC_TRACE='*'` environment
+variable to the DaemonSet, which will ask the GRPC library to trace all if it's
```
Molter73, line 616:
```suggestion
In case of network connectivity issues it can be hard to figure out the reason
why Collector refuses to connect to Sensor. The reason for that is the
WaitForConnected GRPC method, which doesn't convey much information. A
workaround to get more information is to add the `GRPC_TRACE='*'` environment
variable to the DaemonSet, which will ask the GRPC library to trace all of its
```

Molter73 (approved):
Neat! I didn't know of this gRPC variable, seems incredibly useful.

Just marked some wording that might be slightly improved, but otherwise LGTM!

PR #2048, opened 2025-02-24
Allow running integration tests with a CRI client

Molter73:
## Description

Add a new CRI executor which allows controlling containers through a standard CRI interface. This allows us to run our integration tests using the containerd engine (should also allow us to use CRI-O, but it is currently untested).

The main reason for this change is that Garden Linux has stopped shipping docker in its VMs and made them immutable, so the only way to properly test Garden kernels is to either use containerd or create our own image of it, we clearly chose the first of the two options.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Run tests on an updated image of Garden Linux


integration-tests/container/processes-listening-on-ports/plop.py
```diff
@@ -15,7 +15,7 @@ def open_port(port: int):
     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
 
     try:
-        s.bind((socket.gethostname(), port))
+        s.bind(('0.0.0.0', port))
```
Molter73, line 18: This was needed because `socket.gethostname()` was failing on Garden, binding to the wildcard IP is enough for the tests.

integration-tests/suites/duplicate_endpoints.go
```diff
@@ -95,19 +95,19 @@ func (s *DuplicateEndpointsTestSuite) TestDuplicateEndpoints() {
...
 
 	// (3) start the new endpoint, opening a different port
-	_, err = s.execContainer("socat", socatCommand)
+	_, err = s.execContainer("socat", socatCommand, true)
 	s.Require().NoError(err)
 
 	// (4) wait for the endpoint to be reported
 	// expecting two endpoints because that is the total expected for the container
-	s.Sensor().ExpectEndpointsN(s.T(), containerID, gScrapeInterval*time.Second, 2)
+	s.Sensor().ExpectEndpointsN(s.T(), containerID, 2*gScrapeInterval*time.Second, 2)
```
Molter73, line 103: This test was failing due to timing on Garden Linux.

ansible/roles/provision-vm/tasks/garden.yml
```diff
@@ -0,0 +1,132 @@
...
+            "ipMasq": true,
+            "promiscMode": true,
+            "ipam": {
+              "type": "host-local",
+              "ranges": [
+                [{
+                  "subnet": "10.88.0.0/16"
+                }],
+                [{
+                  "subnet": "2001:4860:4860::/64"
```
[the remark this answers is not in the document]
Molter73: This entire configuration file was taken from [this script](https://github.com/containerd/containerd/blob/7a0ad09a0d5edcd40555a998e0e065e6dda43acc/script/setup/install-cni), I'll add a comment so we have the reference to it. My understanding is that these are sort of default subnets used by CNIs and they are required for the CRI plugin containerd uses.

Molter73:
Looks like the konflux tests are not very happy with the bump in QA version, I'll look into it and fix it in a follow up.

Stringy (approved):

PR #2050, opened 2025-02-25
Set selinux labels to share source directory with builder.

ovalenti:
## Description

On systems where selinux is _enforcing_, docker/podman has to be told to set appropriate labels on the builder container to access the source tree from within the container.

## Testing Performed

It should be enough that the build succeeds.


robbycochran:
> @robbycochran, would this change have an impact on your workflow ?

No this should be fine.

PR #2064, opened 2025-03-18
Replace MakeUnique with std::make_unique

Molter73:
## Description

This is a small clean up, removing our home made implementation of MakeUnique (which was needed back when we were still on C++11) and replacing it with std::make_unique.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI should be enough.


Molter73:
Konflux builds seem to be randomly failing and I'm not able to check the logs for what the error actually is.

Because I've seen successful builds on all architectures and the change is pretty straight forward, I'm gonne merge the PR and deal with the aftermath if something is actually broken.

PR #2065, opened 2025-03-19
Use RAII on CollectorService and system-inspector/Service

Molter73:
## Description

This patch replaces some init methods in favour of using constructors for initializing objects in an approach closer to RAII. It also moves some logic for the CollectorService::RunForever method to the constructor and destructor of this class and moves some variables from this method to be members of the class.

Some other minor cleanups have also been applied:
- Define all default constructors, assignment operators and destructors following the rule of 5.
- Uniform formatting of cpp and header files in editorconfig.
- Generalization of handlers to be used by Civetweb server.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [x] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Manually test the prometheus endpoint, ensure stats are still being exported correctly.
- [x] Manually test the civet endpoints, ensure functionality is still correct on each one.


integration-tests/suites/log_level_endpoint.go
```diff
@@ -0,0 +1,58 @@
...
+	s.Assert().NoError(err)
+
+	var logLevel struct {
+		Status string
+		Level  string
+	}
+
+	err = json.Unmarshal(bodyV, &logLevel)
+	s.Assert().NoError(err)
+	s.Assert().Equal("INFO", logLevel.Level)
```
JoukoVirtanen, line 57: Would it be possible to test with different log levels?

integration-tests/suites/prometheus.go
```diff
@@ -0,0 +1,33 @@
...
+	response, err := http.Get("http://localhost:9090/metrics")
+	s.Require().NoError(err)
+	s.Assert().Equal(response.StatusCode, 200)
+
+	// Parse the body to check it is a valid prometheus output
+	body := response.Body
+	defer body.Close()
+	var parser expfmt.TextParser
+	_, err = parser.TextToMetricFamilies(body)
+	s.Require().NoError(err)
```
JoukoVirtanen, line 32: Maybe check that some of the fields are non-zero.

Molter73: The output for this is a bit weird, I'll add a check to verify the resulting map has some keys in it, think that should be enough.

Molter73:
@sourcery-ai review

Stringy (approved):
One largely ignorable nit, but otherwise LGTM! this feels so much cleaner

integration-tests/suites/prometheus.go
JoukoVirtanen: This is enough.

JoukoVirtanen (approved):
LGTM! A couple of CI failures look like flakes. I am rerunning them.

PR #2066, opened 2025-03-20
Fix CPU profiler

Molter73:
## Description

When we removed the /module mount from the collector daemonset we inadvertedly broke the gperftool cpu profiler, since it was using that directory for dumping its output files. This patch moves the dump directory to /var/profiles, which will be a new emptydir mount on the collector daemonset.

A new test is also added to ensure the profiles still starts, stops and produces some output.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [x] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Manually tested the CPU profiler endpoint.

### DIFF
--- a/collector/lib/ProfilerHandler.cpp
+++ b/collector/lib/ProfilerHandler.cpp
@@ -13,6 +13,9 @@
 
 namespace collector {
 
+// The following route is used by the CPU profiler to dump information
+// into. When running in read only file systems, /var/profiles needs to
+// be writable, or the CPU profiler will not work.
 const std::string ProfilerHandler::kCPUProfileFilename = "/var/profiles/cpu_profile";
 const std::string ProfilerHandler::kBaseRoute = "/profile";
 const std::string ProfilerHandler::kCPURoute = kBaseRoute + "/cpu";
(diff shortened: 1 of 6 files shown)


Molter73:
Sibling PR on the main repo: stackrox/stackrox#14689

collector/lib/ProfilerHandler.cpp
```diff
@@ -13,7 +13,7 @@
 
 namespace collector {
 
-const std::string ProfilerHandler::kCPUProfileFilename = "/module/cpu_profile";
+const std::string ProfilerHandler::kCPUProfileFilename = "/var/profiles/cpu_profile";
```
erthalion, line 16: Maybe worth adding a few lines of documentation about that?

erthalion (approved):

collector/lib/ProfilerHandler.cpp
Molter73: Makes sense, I'll add a comment.

Molter73: I also added a small section to our docs, think that should be enough.

PR #2067, opened 2025-03-20
Stop running the unit tests as part of the image build

Molter73:
## Description

When iterating on collector changes, waiting for the image build takes some time because we build the binary, then run the unit tests which also attempt to build the binary again and run the tests themselves. If we ask for the image to be built, we should only do the steps required to build the collector binary and creating the image, if we want to run unit tests, we have a separate target for it.

On CI, the unit tests are already being run as a separate step, so removing it from the image build should be a net win with no losses there.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI should be enough.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ connscrape:
 unittest:
 	make -C collector unittest
 
-image: collector unittest
+image: collector
 	make -C collector txt-files
 	docker buildx build --load --platform ${PLATFORM} \
 		--build-arg COLLECTOR_VERSION="$(COLLECTOR_TAG)" \
@@ -53,7 +53,7 @@ image: collector unittest
 		-t quay.io/stackrox-io/collector:$(COLLECTOR_TAG) \
 		$(COLLECTOR_BUILD_CONTEXT)
 
-image-dev: collector unittest container-dockerfile-dev
+image-dev: collector container-dockerfile-dev
 	make -C collector txt-files
 	docker buildx build --load --platform ${PLATFORM} \
 		--build-arg COLLECTOR_VERSION="$(COLLECTOR_TAG)" \
--- a/collector/container/konflux.Dockerfile
+++ b/collector/container/konflux.Dockerfile
@@ -90,7 +90,7 @@ RUN if [[ "$(uname -m)" == "x86_64" ]];   \
            -DADDRESS_SANITIZER="${ADDRESS_SANITIZER}" \
            -DCOLLECTOR_VERSION="${COLLECTOR_TAG}" \
            -DTRACE_SINSP_EVENTS="${TRACE_SINSP_EVENTS}"
-RUN cmake --build "${CMAKE_BUILD_DIR}" --target collector -- -j "${NPROCS:-4}"
+RUN cmake --build "${CMAKE_BUILD_DIR}" --target collector self-checks -- -j "${NPROCS:-4}"
 RUN strip -v --strip-unneeded "${CMAKE_BUILD_DIR}/collector/collector"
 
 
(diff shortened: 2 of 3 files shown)


collector/container/konflux.Dockerfile
```diff
@@ -90,8 +90,7 @@ RUN if [[ "$(uname -m)" == "x86_64" ]];   \
            -DADDRESS_SANITIZER="${ADDRESS_SANITIZER}" \
            -DCOLLECTOR_VERSION="${COLLECTOR_TAG}" \
            -DTRACE_SINSP_EVENTS="${TRACE_SINSP_EVENTS}"
-RUN cmake --build "${CMAKE_BUILD_DIR}" --target all -- -j "${NPROCS:-4}"
-RUN ctest --no-tests=error -V --test-dir "${CMAKE_BUILD_DIR}"
```
ovalenti, line 94: Unit tests are not run at all in Lonflux builds ?

Makefile
```diff
@@ -45,15 +45,15 @@ connscrape:
 unittest:
 	make -C collector unittest
 
-image: collector unittest
```
ovalenti, line 48: There is a unit-test job in CI, but the unit-tests are never run with the build options of the production binary. It happened in the past that binaries built for debug or running inside a special environment are not detecting issues.

ovalenti (commented):
It makes sense, thanks for optimizing things a bit.

Makefile
Molter73: Good call, I'll add a separate job that runs with `Release` the release build option.

collector/container/konflux.Dockerfile
Molter73: I don't really see the point, it's the same code we build and run upstream and we run integration tests on the produced images. Do you feel strongly about this? I can always just add them back.

ovalenti (changes requested):
Konflux is broken

collector/container/konflux.Dockerfile
Molter73: Well, our integration tests should catch if anything is broken, I don't think our code is so tied to compiler implementations that it would matter if something is slightly different.

Maybe we leave this one open for our @stackrox/rhtap-maintainers buddies, in case they fell otherwise.

ovalenti (approved):
Would be good to see at least one Konflux image pass integration-test, but otherwise, LGTM !

collector/container/konflux.Dockerfile
msugakov: Sorry, later here but would it be beneficial to run unit tests in Konflux builds for non-mainstream architectures? I mean you probably don't have ppc64le and s390x in GHA for running unit tests while still doing this in Konflux dockerfiles could be a smoke test.

msugakov: Here are timings from the [current build](https://github.com/stackrox/collector/pull/2067/checks?check_run_id=39299284620) (we can ignore the failure).
image

Here timings from [some `master` build](https://github.com/stackrox/collector/runs/39099990674).
image

Molter73: > Would it be beneficial to run unit tests in Konflux builds for non-mainstream architectures?

Maybe, but our unit tests are not low level enough that a change in architecture would cause issues (we don't test network or file IO, we don't test ebpf, etc.). Even on tests that do work on architecture specific things, we just mock it (see [host info unit tests](https://github.com/stackrox/collector/blob/master/collector/test/HostInfoTest.cpp)), and I would argue this is the right approach for unit tests, since we should be testing the code logic itself and not its dependency with the system/external dependencies.

Integration tests on the other hand is where I see this compatibility should be checked and we have integration tests for all archs with konflux builds.

> Seem to me in the same ballpark so maybe there's not much time gain?

Yup, this change is not meant to have a big impact on CI (either GHA or konflux), the main benefit will be on local development, since currently running the unit tests takes about 12 seconds and they run every time you need to rebuild the image. It's also not a huge gain here, but trust me, those 12 seconds when you are trying to iterate on a problem are really annoying to sit through when you only need to rebuild a single file.

All that being said, I think I'm just going to add them back in for konflux, since people seem to be hesitant to remove them at this time.

msugakov (approved):
No objections from the "Konflux side"

PR #2082, opened 2025-04-11
Stop copying hostname

Molter73:
## Description

The hostname for a given node is lazily loaded in the HostInfo and will outlast any other object attempting to use it. As such, we should try to avoid unnecessary copies and just call the GetHostname method of the HostInfo class.

The GetHostname function from our utilities has been turned into a static method of HostInfo so we can directly access it calling HostInfo::GetHostname(), otherwise we would need HostInfo::Instance().GetHostname() which is a bit too verbose.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Check CI logs to validate the hostname is properly being captured.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,7 +6,7 @@ jobs:
   lint:
     runs-on: ubuntu-24.04
     container:
-      image: fedora:40
+      image: fedora:41
     steps:
       - name: dnf install dependencies
         run: |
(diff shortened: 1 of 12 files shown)


.github/workflows/lint.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   lint:
     runs-on: ubuntu-24.04
     container:
-      image: fedora:40
+      image: fedora:41
```
Molter73, line 9: Got a lint error due to a clang-format mismatch, bumping here since we should stick to the latest versions if possible IMO.

Stringy (approved):

PR #2084, opened 2025-04-14
Use ARM GH action workflow runners for ARM builds

robbycochran:
## Description

Utilize newly added [arm64 runners](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/ ) for github actions.

* Build arm64 images natively instead of using QEMU in all workflows
* Make arm64 builds, unit-tests, and integration tests run alongside x86_64 **by default** on PR workflows. Build time performance is slightly faster on ARM so it shouldn't increase total time to finish PR steps.
* Split and refactored the builder image job into a local and remote version 

## Checklist

**Automated testing**
  - [ ] Added integration tests

## Testing Performed

- [ ] test run with `run-multiarch-builds` label -> https://github.com/stackrox/collector/pull/2094 
- [x] CI passes on PR with arm64 builds and tests
- [ ] testing on master commits, schedule runs, or release branches 🤷

### DIFF
--- a/.github/workflows/collector-builder.yml
+++ b/.github/workflows/collector-builder.yml
@@ -26,6 +26,9 @@ jobs:
       build-image: ${{ steps.check-builder-needed.outcome == 'success' }}
       collector-builder-tag: ${{ steps.builder-tag.outputs.collector-builder-tag }}
 
+    env:
+      DEFAULT_BUILDER_TAG: master
+
     steps:
       - uses: actions/checkout@v4
 
@@ -65,7 +68,7 @@ jobs:
           echo "Rebuild builder image with tag ${COLLECTOR_BUILDER_TAG}"
 
   build-builder-image:
-    name: Local builder image ${{ matrix.arch }}
+    name: Local builder image
     # Multiarch builds sometimes take for eeeeeeeeeever
     timeout-minutes: 480
     needs:
@@ -122,7 +125,7 @@ jobs:
             ansible/ci-build-builder.yml
 
   build-builder-image-remote-vm:
-    name: Remote builder image s390x
+    name: Remote builder image
     # Multiarch builds sometimes take for eeeeeeeeeever
     timeout-minutes: 480
     needs:
--- a/.github/workflows/collector-builder.yml
+++ b/.github/workflows/collector-builder.yml
@@ -61,7 +61,7 @@ jobs:
     strategy:
       matrix:
         arch: [amd64, arm64]
-    runs-on: ${{ (matrix.arch != 'arm64' && ubuntu-24.04) || 'ubuntu-24.04-arm' }}
+    runs-on: ${{ (matrix.arch == 'arm64' && 'ubuntu-24.04-arm') || 'ubuntu-24.04' }}
 
     env:
       PLATFORM: linux/${{ matrix.arch }}
@@ -118,19 +118,19 @@ jobs:
             ansible/ci-build-builder.yml
 
   build-builder-image-remote:
-    if: contains(inputs.include-arch-json, 's390x')
     name: Build the builder image
     # Multiarch builds sometimes take for eeeeeeeeeever
     timeout-minutes: 480
     needs:
     - builder-needs-rebuilding
     if: |
-      needs.builder-needs-rebuilding.outputs.build-image == 'true' ||
+      (needs.builder-needs-rebuilding.outputs.build-image == 'true' ||
       (github.event_name == 'push' && (
         github.ref_type == 'tag' || startsWith(github.ref_name, 'release-')
       )) ||
       contains(github.event.pull_request.labels.*.name, 'build-builder-image') ||
-      github.event_name == 'schedule'
+      github.event_name == 'schedule') &&
+      contains(inputs.include-arch-json, 's390x')
     outputs:
       collector-builder-tag: ${{ steps.builder-tag.outputs.collector-builder-tag }}
     strategy:
--- a/.github/workflows/collector-builder.yml
+++ b/.github/workflows/collector-builder.yml
@@ -8,6 +8,11 @@ on:
         required: true
         description: |
           The tag used to build the collector image
+      include-arch-json:
+        type: string
+        required: true
+        description: |
+          JSON for matrix.include to select architectures and runners to build
     outputs:
       collector-builder-tag:
         description: The builder tag used by the build
@@ -40,7 +45,6 @@ jobs:
 
   build-builder-image:
     name: Build the builder image
-    runs-on: ubuntu-24.04
     # Multiarch builds sometimes take for eeeeeeeeeever
     timeout-minutes: 480
     needs:
@@ -55,9 +59,9 @@ jobs:
     outputs:
       collector-builder-tag: ${{ steps.builder-tag.outputs.collector-builder-tag }}
     strategy:
-      fail-fast: false
       matrix:
-        arch: [amd64, ppc64le, s390x, arm64]
+        include: ${{ fromJSON(inputs.include-arch-json) }}
+    runs-on: ${{ matrix.runner }}
 
     env:
       PLATFORM: linux/${{ matrix.arch }}
@@ -69,24 +73,30 @@ jobs:
           submodules: true
 
       - name: Set up QEMU
+        if: matrix.remote != true
         uses: docker/setup-qemu-action@v3
         with:
           image: tonistiigi/binfmt:qemu-v8.1.5
 
       - name: Set up Docker Buildx
+        if: matrix.remote != true
         uses: docker/setup-buildx-action@v3
 
       - uses: actions/setup-python@v5
+        if: matrix.remote
         with:
           python-version: "3.10"
 
       - uses: 'google-github-actions/auth@v2'
+        if: matrix.remote
         with:
           credentials_json: '${{ secrets.GOOGLE_CREDENTIALS_COLLECTOR_SVC_ACCT }}'
 
       - uses: 'google-github-actions/setup-gcloud@v2'
+        if: matrix.remote
 
       - uses: ./.github/actions/setup-vm-creds
+        if: matrix.remote
         with:
           gcp-ssh-key: ${{ secrets.GCP_SSH_KEY }}
           gcp-ssh-key-pub: ${{ secrets.GCP_SSH_KEY_PUB }}
@@ -101,9 +111,7 @@ jobs:
           job-tag: builder
 
       - name: Create Build VMs
-        if: |
-          matrix.arch == 's390x' &&
-          (github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds'))
+        if: matrix.remote
         run: |
           make -C "${{ github.workspace }}/ansible" create-build-vms
 
@@ -133,10 +141,7 @@ jobs:
           } > ${{ github.workspace }}/ansible/secrets.yml
 
       - name: Build images
-        if: |
-          (github.event_name != 'pull_request' && matrix.arch != 's390x') ||
-          matrix.arch == 'amd64' ||
-          (contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds') && matrix.arch != 's390x')
+        if: matrix.remote != true
         timeout-minutes: 480
         run: |
           ansible-galaxy install -r ansible/requirements.yml
@@ -149,9 +154,7 @@ jobs:
             ansible/ci-build-builder.yml
 
       - name: Build s390x images
-        if: |
-          (github.event_name != 'pull_request' && matrix.arch == 's390x') ||
-          (contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds') && matrix.arch == 's390x')
+        if: matrix.remote
         timeout-minutes: 480
         run: |
           ansible-playbook \
@@ -162,7 +165,7 @@ jobs:
             ansible/ci-build-builder.yml
 
       - name: Destroy VMs
-        if: always() && matrix.arch == 's390x'
+        if: always() && matrix.remote
         run: |
           make -C ansible destroy-vms
 
@@ -173,11 +176,10 @@ jobs:
     runs-on: ubuntu-24.04
     if: |
       github.event_name != 'pull_request' ||
-      (needs.build-builder-image.outputs.collector-builder-tag != 'cache' &&
-       contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds'))
+      needs.build-builder-image.outputs.collector-builder-tag != 'cache'
     env:
       COLLECTOR_BUILDER_TAG: ${{ needs.build-builder-image.outputs.collector-builder-tag }}
-      ARCHS: amd64 ppc64le s390x arm64
+      ARCHS: ${{ join(fromJSON(inputs.include-arch-json).*.arch, ' ') }}
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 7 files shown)


.github/workflows/collector.yml
```diff
@@ -168,15 +175,19 @@ jobs:
     - build-collector-image-remote-vm
     name: Create Multiarch manifest
     runs-on: ubuntu-24.04
```
Molter73, line 177: CI is still skipping this because the `build-collector-image-remote-vm` is being skipped. Add this to force it to run:
```suggestion
    runs-on: ubuntu-24.04
    if: always()
```

Molter73: Or you can refine it a bit more so it doesn't run with failures with something like this:
```suggestion
    runs-on: ubuntu-24.04
    if: always() && !contains(join(needs.*.result, ','), 'failure')
```

.github/workflows/collector.yml
```diff
@@ -168,15 +175,19 @@ jobs:
...
 
     steps:
       - uses: actions/checkout@v4
 
+      - name: Set ARCHS
+        run: |
+          if [[ "${GITHUB_EVENT_NAME}" == "pull_request" && "${{ contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds') }}" != "true" ]]; then
+            ARCHS="amd64 arm64"
+            echo "ARCHS=$ARCHS" >> "$GITHUB_ENV"
+          fi
```
Molter73, line 189: This might be more readable.
```suggestion
      - name: Set ARCHS
        if: github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds'
        run: |
          ARCHS="amd64 arm64"
          echo "ARCHS=$ARCHS" >> "$GITHUB_ENV"
```

.github/workflows/collector.yml
```diff
@@ -168,15 +175,19 @@ jobs:
...
       ARCHS: amd64 ppc64le s390x arm64
 
     steps:
       - uses: actions/checkout@v4
 
+      - name: Set ARCHS
+        run: |
+          if [[ "${GITHUB_EVENT_NAME}" == "pull_request" && "${{ contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds') }}" != "true" ]]; then
+            ARCHS="amd64 arm64"
+            echo "ARCHS=$ARCHS" >> "$GITHUB_ENV"
```
Molter73, line 188: Don't you need to quote `$ARCHS` here? Otherwise it will get expanded before echoing out and the env file will end up looking like this: `ARCHS=amd64 arm64`, which is wrong.
```suggestion
            echo "ARCHS='$ARCHS'" >> "$GITHUB_ENV"
```

.github/workflows/collector-builder.yml
```diff
@@ -57,7 +56,16 @@ jobs:
...
+        include:
+        - arch: amd64
+          runner: ubuntu-24.04
+        - arch: ppc64le
+          runner: ubuntu-24.04
+        - arch: s390x
+          runner: ubuntu-24.04
+        - arch: arm64
+          runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
```
Molter73, line 68: Alternatively, you could do something like this:
```suggestion
        arch: [amd64, ppc64le, s390x, arm64]
    runs-on: ${{ matrix.arch != 'arm64' && 'ubuntu-24.04' || 'ubuntu-24.04-arm' }}
```

Molter73 (commented):
The arm image builds in 6 minutes 😻

.github/workflows/collector-builder.yml
robbycochran: I tried it your way and it seemed like the expression evaluation didn't work.

.github/workflows/collector.yml
```diff
@@ -86,32 +90,21 @@ jobs:
             -e @'${{ github.workspace }}/ansible/secrets.yml' \
             ansible/ci-build-collector.yml
 
-  build-collector-image-remote-vm:
```
Molter73, line 89: Please split the jobs back again. We used to have them unified, it is a nightmare to figure out what steps run on which conditions and having to add `if: something` to every single step is extremely tedious and annoying (specially when you forget one and weird errors start showing up).

.github/workflows/main.yml
```diff
@@ -66,9 +68,11 @@ jobs:
   unit-tests:
     uses: ./.github/workflows/unit-tests.yml
     needs:
-      - build-builder-image
+    - init
+    - build-builder-image
     with:
       collector-builder-tag: ${{ needs.build-builder-image.outputs.collector-builder-tag }}
+      include-arch-json: ${{ needs.init.outputs.include-arch-json }}
```
Molter73, line 75: This `include-arch-json` is being passed to 3 different workflows, because it is defined as a string, it is very opaque and hard to figure out what it is meant to hold when you are using it from one of these workflows. Do we have to do it this way? AFAICT, all it holds is a map that looks something like this:
```json
{
  "arch": {
    "remote": false,
    "runner": "ubuntu-24.04"
  }
}
```
If you split the workflows for local and remote like I commented previously, the remote field basically goes away and all that's left for figuring out is the runner, which only changes for arm.

.github/workflows/collector-builder.yml
Molter73: Reopening this one since I think it would simplify a lot of the logic.

Checked in Falco's repo, since I know they do it this way, looks like I missed a couple of parenthesis, should be:

```yaml
    runs-on: ${{ (matrix.arch != 'arm64' && ubuntu-24.04) || 'ubuntu-24.04-arm' }}
```

https://github.com/falcosecurity/libs/blob/24539f5cdd818aec2498db5c1e961a8dacc4c36c/.github/workflows/ci.yml#L19

robbycochran: Cool -- yeah that collapses a lot of logic, will try that again.

robbycochran: Ok, got this approach to work which simplifies things.

.github/workflows/konflux.yml
```diff
@@ -20,7 +20,20 @@ concurrency:
...
 jobs:
+  should-run:
+    outputs:
+      should-run: ${{ steps.should-run.outputs.should-run }}
+    runs-on: ubuntu-24.04
+    steps:
+      - id: should-run
+        run: |
+          SHOULD_RUN="${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run-konflux-steps') }}"
+          echo "should-run=${SHOULD_RUN}" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
robbycochran: I didn't think it was needed on PR builds but pulling this out.

.github/workflows/unit-tests.yml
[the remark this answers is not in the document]
robbycochran: Leaving this in for now (it's simpler) happy to remove it though. It is useful to have validation in the unit tests for another arch

.github/workflows/collector.yml
robbycochran: Done

.github/workflows/main.yml
robbycochran: Yeah this was going to cause issues, i've removed it and the ability to specify a single arch

.github/workflows/collector-builder.yml
```diff
@@ -38,30 +42,52 @@ jobs:
...
-    outputs:
-      collector-builder-tag: ${{ steps.builder-tag.outputs.collector-builder-tag }}
+      needs.builder-needs-rebuilding.outputs.build-image == 'true'
     strategy:
-      fail-fast: false
       matrix:
-        arch: [amd64, ppc64le, s390x, arm64]
+        arch: [amd64, arm64, ppc64le]
+        exclude:
+          - arch: ${{ (!contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds') && 'ppc64le') || '' }}
```
Molter73, line 84: I think this will also exclude ppc64le from workflow runs triggered from schedule and push events, since those won't have the `run-multiarch-builds` label. You might need to tweak the condition a bit:
```suggestion
          - arch: ${{ ((
              github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds')
            ) && 'ppc64le') || '' }}
```

.github/workflows/collector-builder.yml
```diff
@@ -129,29 +187,10 @@ jobs:
...
-          ansible-galaxy install -r ansible/requirements.yml
-          ansible-playbook \
-            --connection local \
-            -i localhost, \
-            --limit localhost \
-            -e arch='${{ matrix.arch }}' \
-            -e @'${{ github.workspace }}/ansible/secrets.yml' \
-            ansible/ci-build-builder.yml
-
       - name: Build s390x images
```
Molter73, line 193: We might want to rename the step now that it is a bit more generic.
```suggestion
      - name: Build images
```

.github/workflows/collector-builder.yml
```diff
@@ -162,22 +201,23 @@ jobs:
...
-      github.event_name != 'pull_request' ||
-      (needs.build-builder-image.outputs.collector-builder-tag != 'cache' &&
-       contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds'))
+      always() && !contains(join(needs.*.result, ','), 'failure') &&
+      needs.builder-needs-rebuilding.outputs.build-image == 'true'
     env:
-      COLLECTOR_BUILDER_TAG: ${{ needs.build-builder-image.outputs.collector-builder-tag }}
-      ARCHS: amd64 ppc64le s390x arm64
+      COLLECTOR_BUILDER_TAG: ${{ needs.builder-needs-rebuilding.outputs.collector-builder-tag }}
+      ARCHS: ${{ (contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds') && 'amd64 arm64 ppc64le s390x') || 'amd64 arm64' }}
```
Molter73, line 220: Similar to a previous comment on this excluding ppc64le and s390x from workflows triggered from push and schedule events.
```suggestion
      ARCHS: ${{ ((
        github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds')
      ) && 'amd64 arm64 ppc64le s390x') || 'amd64 arm64' }}
```

.github/workflows/collector.yml
```diff
@@ -29,12 +29,13 @@ env:
...
-    name: Build Collector
-    runs-on: ubuntu-24.04
+    name: Local collector image ${{ matrix.arch }}
     strategy:
-      fail-fast: false
       matrix:
-        arch: [amd64, ppc64le, arm64]
+        arch: [amd64, arm64, ppc64le]
+        exclude:
+          - arch: ${{ (!contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds') && 'ppc64le') || '' }}
```
Molter73, line 37: Same comment as in the builder workflow
```suggestion
          - arch: ${{ ((
            github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds')
          ) && 'ppc64le') || '' }}
```

.github/workflows/collector.yml
```diff
@@ -87,11 +85,10 @@ jobs:
             ansible/ci-build-collector.yml
 
   build-collector-image-remote-vm:
-    name: Build Collector on a remote VM
+    name: Remote collector image s390x
```
Molter73, line 88:
```suggestion
    name: Remote collector image
```

.github/workflows/collector.yml
```diff
@@ -168,11 +165,9 @@ jobs:
     - build-collector-image-remote-vm
     name: Create Multiarch manifest
     runs-on: ubuntu-24.04
-    if: |
-      github.event_name != 'pull_request' ||
-      contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds')
+    if: always() && !contains(join(needs.*.result, ','), 'failure')
     env:
-      ARCHS: amd64 ppc64le s390x arm64
+      ARCHS: ${{ (contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds') && 'amd64 arm64 ppc64le s390x') || 'amd64 arm64' }}
```
Molter73, line 170: Similar to the comment on the builder workflow
```suggestion
      ARCHS: ${{ ((
        github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds')
      ) && 'amd64 arm64 ppc64le s390x') || 'amd64 arm64' }}
```

.github/workflows/unit-tests.yml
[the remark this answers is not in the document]
Molter73: Now that we are just running all variations of the unit tests, I think it makes a bit more sense and it doesn't add much complexity to the workflow itself, so I'm ok with leaving it in.

.github/workflows/collector-builder.yml
```diff
@@ -8,10 +8,20 @@ on:
...
           The tag used to build the collector image
+      architectures:
+        type: string
+        required: false
+        description: |
+          Space-seperated list of architectures
+        default: |
+          ${{ ((github.event_name != 'pull_request' ||
+            contains(github.event.pull_request.labels.*.name, 'run-multiarch-builds')) &&
+            'amd64 arm64 ppc64le s390x') || 'amd64 arm64' }}
```
Molter73, line 19: I see this logic is duplicated for `collector.yml` and `collector-builder.yml`, yet the input is not actually used.

Consider making this logic an output of the `init.yml` workflow, then pass it in to these 2 workflows in `main.yml`. Also change the `required` field to `true` to ensure it is properly being provided.

See how the `collector-tag` variable is handled for an example on how to do this.

.github/workflows/collector-builder.yml
```diff
@@ -38,30 +52,52 @@ jobs:
...
+        run: |
+          COLLECTOR_BUILDER_TAG="${DEFAULT_BUILDER_TAG}"
+          if [[ "${{ github.event_name }}" == 'pull_request' || \
+                "${{ github.ref_type }}" == 'tag' || \
+                "${{ github.ref_name }}" =~ ^release- ]]; then
+            COLLECTOR_BUILDER_TAG="${{ inputs.collector-tag }}"
+          fi
+
+          echo "Rebuild builder image with tag ${COLLECTOR_BUILDER_TAG}"
+          echo "collector-builder-tag=${COLLECTOR_BUILDER_TAG}" >> "$GITHUB_OUTPUT"
```
Molter73, line 80: Could we merge these two steps?
```suggestion
      - name: Define builder tag
        id: builder-tag
        if: |
          steps.changed.outputs.builder-changed == 'true' ||
          (github.event_name == 'push' && (
            github.ref_type == 'tag' || startsWith(github.ref_name, 'release-')
          )) ||
          contains(github.event.pull_request.labels.*.name, 'build-builder-image') ||
          github.event_name == 'schedule'
        run: |
          COLLECTOR_BUILDER_TAG="${DEFAULT_BUILDER_TAG}"
          if [[ "${{ github.event_name }}" == 'pull_request' || \
                "${{ github.ref_type }}" == 'tag' || \
                "${{ github.ref_name }}" =~ ^release- ]]; then
            COLLECTOR_BUILDER_TAG="${{ inputs.collector-tag }}"
          fi

          echo "Rebuild builder image with tag ${COLLECTOR_BUILDER_TAG}"
          echo "collector-builder-tag=${COLLECTOR_BUILDER_TAG}" >> "$GITHUB_OUTPUT"
          echo "build-image=true" >> "$GITHUB_OUTPUT"
```

Then on line 36 you could just do something like this:
```yaml
      build-image: ${{ steps.builder-tag.outputs.build-image || false }}
```

Or if you want to keep the check similar to what you had:
```
      build-image: ${{ steps.builder-tag.outcome == 'success' }}
```

.github/workflows/collector-builder.yml
```diff
@@ -76,6 +112,54 @@ jobs:
...
+        run: |
+          {
+            echo "---"
+            echo "stackrox_io_username: ${{ secrets.QUAY_STACKROX_IO_RW_USERNAME }}"
+            echo "stackrox_io_password: ${{ secrets.QUAY_STACKROX_IO_RW_PASSWORD }}"
+            echo "rhacs_eng_username: ${{ secrets.QUAY_RHACS_ENG_RW_USERNAME }}"
+            echo "rhacs_eng_password: ${{ secrets.QUAY_RHACS_ENG_RW_PASSWORD }}"
+            echo "collector_git_ref: ${{ github.ref }}"
+            echo "collector_builder_tag: ${{ env.COLLECTOR_BUILDER_TAG }}"
+          } > ${{ github.workspace }}/ansible/secrets.yml
```
Molter73, line 125: Insisting on the usage of `cat` here, since I can't see my previous comment on it here
```suggestion
          cat << EOF > ${{ github.workspace }}/ansible/secrets.yml
            ---
            stackrox_io_username: ${{ secrets.QUAY_STACKROX_IO_RW_USERNAME }}
            stackrox_io_password: ${{ secrets.QUAY_STACKROX_IO_RW_PASSWORD }}
            rhacs_eng_username: ${{ secrets.QUAY_RHACS_ENG_RW_USERNAME }}
            rhacs_eng_password: ${{ secrets.QUAY_RHACS_ENG_RW_PASSWORD }}
            collector_git_ref: ${{ github.ref }}
            collector_builder_tag: ${{ env.COLLECTOR_BUILDER_TAG }}
          EOF
```

.github/workflows/collector.yml
```diff
@@ -29,12 +34,13 @@ env:
...
-    runs-on: ubuntu-24.04
+    name: Local collector image ${{ matrix.arch }}
     strategy:
-      fail-fast: false
       matrix:
-        arch: [amd64, ppc64le, arm64]
+        arch: [amd64, arm64, ppc64le]
+        exclude:
+          - arch: ${{ contains(inputs.architectures, 'ppc64le') && '' || 'ppc64le' }}
+    runs-on: ${{ (matrix.arch == 'arm64' && 'ubuntu-22.04-arm') || 'ubuntu-22.04' }}
```
Molter73, line 43: Any reason in particular why we rolled back to ubuntu-22.04? Or is it just copy-pasting from the Falco workflow?

If there's no strong reason, I would suggest sticking to 24.04, just to be on latest.

.github/workflows/unit-tests.yml
```diff
@@ -11,16 +11,18 @@ on:
...
       image: quay.io/stackrox-io/collector-builder:${{ inputs.collector-builder-tag }}
     strategy:
       fail-fast: false
       matrix:
+        arch: [amd64, arm64]
         cmake-flags:
         - -DCMAKE_BUILD_TYPE=Release
         - -DADDRESS_SANITIZER=ON -DCMAKE_BUILD_TYPE=Debug
         - -DUSE_VALGRIND=ON -DCMAKE_BUILD_TYPE=Debug
+    runs-on: ${{ (matrix.arch == 'arm64' && 'ubuntu-22.04-arm') || 'ubuntu-22.04' }}
```
Molter73, line 24: Same as my previous comment here.

Molter73 (approved):
Just a couple comments on the runner version, but other than that LGTM!

Thanks for putting the effort into this, it simplifies a lot of code in our CI and it'll be great to have arm built and tested on all PRs!

PR #2089, opened 2025-04-16
Refactor NetworkStatusNotifier

Molter73:
## Description

This change started out from the effort to unify the services used to communicate with sensor. While attempting to add an output object to the notifier it became apparent this class has some issues with ownership of its members, which in turn makes it a bit odd to extend and modify.

In order to alleviate the problem the following changes are introduced:
- The connection scraper is now owned by NetworkStatusNotifier.
- The communication channel for network info is now owned by NetworkStatusNotifier.
- Connection stats are now owned by NetworkStatusNotifier.
- Move ownership of the prometheus registry into CollectorStatsExporter.
- The process store is now owned by ConnScrapper.
- Refactor CollectorConnectionStats into its own file.

When stating "X is now owned by Y" it stands for X now is part of Y and its lifecycle is fully handled by Y (i.e X is created with Y and destroyed with Y, preferably following RAII).

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Check connection stats are still properly informed by the prometheus exporter.

### DIFF
--- a/collector/lib/CollectorConnectionStats.h
+++ b/collector/lib/CollectorConnectionStats.h
@@ -37,6 +37,12 @@ class CollectorConnectionStats {
   }
 
  private:
+  // prometheus::Registry owns all stats and only exposes references to
+  // them. However, storing references as members of a class invalidate
+  // copy and move constructors and assignment operators, making it
+  // harder to work with this type of class.
+  //
+  // To work around this, we store raw pointers instead.
   prometheus::Summary* inbound_private_summary_;
   prometheus::Summary* inbound_public_summary_;
   prometheus::Summary* outbound_private_summary_;
--- a/collector/lib/CollectorConnectionStats.h
+++ b/collector/lib/CollectorConnectionStats.h
@@ -1,0 +1,61 @@
+#ifndef _COLLECTOR_CONNECTION_STATS_
+#define _COLLECTOR_CONNECTION_STATS_
+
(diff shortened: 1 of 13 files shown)


collector/lib/CollectorConnectionStats.h
```diff
@@ -0,0 +1,61 @@
...
+    inbound_public_summary_->Observe(inbound_public);
+    outbound_private_summary_->Observe(outbound_private);
+    outbound_public_summary_->Observe(outbound_public);
+  }
+
+ private:
+  prometheus::Summary* inbound_private_summary_;
+  prometheus::Summary* inbound_public_summary_;
+  prometheus::Summary* outbound_private_summary_;
+  prometheus::Summary* outbound_public_summary_;
```
Molter73, line 43: These have been turned from references to raw pointers. The way they are created makes it so the prometheus registry owns the actual objects and we only get a reference, however, storing references implicitly makes it so the copy and move constructors and operators need to be deleted, (because references are non-nullable and they can't be assigned to point into the ether).

Storing the raw pointers makes handling of the class easier and should be safe as long as we stop the network notifier before we destroy the prometheus exposer (which we do, because they are both owned by the CollectorService object and the exposer is stopped in its destructor).

Stringy (approved):
LGTM!

PR #2103, opened 2025-04-25
Fix Garden Linux Bumper

Molter73:
## Description

Garden Linux changed how it names its GCP images, so we cannot get that
from the assets of a release directly. Instead, we need to parse the
markdown text on the release itself and get it from there.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Run the bumper from this PR and it generated another PR with the correct image name: #2146

### DIFF
--- a/utilities/gardenlinux-bumper/bumper.py
+++ b/utilities/gardenlinux-bumper/bumper.py
@@ -3,16 +3,24 @@
 import argparse
 import os
 import re
+from typing import NamedTuple
+
+from bs4 import BeautifulSoup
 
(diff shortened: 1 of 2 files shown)


Molter73:
> LGTM! sourcery does maaaybe have a point about using regex directly, but I think your impl might be slightly more robust to further changes

Yeah... I don't like the idea of parsing documents with regex, that just feels like a recipe for disaster.

PR #2111, opened 2025-05-05
Fix isCanonicalExternalIp

JoukoVirtanen:
## Description

Currently `isCanonicalIp` does not take into account the endianness of the system. This PR fixes that.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI is sufficient.

### DIFF
--- a/collector/lib/NetworkConnection.h
+++ b/collector/lib/NetworkConnection.h
@@ -139,7 +139,9 @@ class Address {
   static bool IsCanonicalExternalIp(const Address& addr) {
     switch (addr.family_) {
       case Family::IPV4:
-        return addr.data_[0] == 0xffffffffULL;
+        uint32_t ipv4_val;
+        std::memcpy(&ipv4_val, &addr.data_[0], sizeof(ipv4_val));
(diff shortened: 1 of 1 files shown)


collector/lib/NetworkConnection.h
```diff
@@ -139,7 +139,9 @@ class Address {
   static bool IsCanonicalExternalIp(const Address& addr) {
     switch (addr.family_) {
       case Family::IPV4:
-        return addr.data_[0] == 0xffffffffULL;
+        uint32_t ipv4_val;
+        std::memcpy(&ipv4_val, &addr.data_[0], sizeof(ipv4_val));
+        return ipv4_val == 0xffffffffULL;
```
Molter73, line 144: I don't like using `std::memcpy` to bit mash an array. Would something like this work?
```c++
return (ipv4_val & htonll(0xffffffffULL)) == htonll(0xffffffffULL);
```

JoukoVirtanen: Yes

PR #2125, opened 2025-05-15
Add a release workflow

Molter73:
## Description

This new workflow will enable people working in the collector repo to simply create a new release by issuing a workflow from the github actions tab. Two inputs can be provided to the workflow:
- version: Specify the <Major>.<minor> version to be tagged.
- dry-run: Run the workflow, create all tags but don't push to origin.

The intended usage is to make a first run with the dry-run flag on to check everything works as expected, then make a second run without the dry-run flag to create the release tags.

If the version input is left unchanged a new minor version will be tagged, creating the <Major>.<minor>.x tag on the master branch, setting up the release-<Major>.<minor> branch with an empty commit and the new .0 tag on top of said commit.

If a version is provided the behavior depends on what the actual value is. The workflow will check if the behavior intended by the caller is to do a major, minor or patch release by checking the provided version against the latest tag on master. In all cases, the workflow ensures the release will always have the correct version numbers.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

The logic for determining versions has been tested using [act](https://github.com/nektos/act), the actual pushing of tags needs to be tested once this PR is merged.

I also did some testing on a fork, here are some example workflows:
- [New minor version](https://github.com/Molter73/stackrox-collector/actions/runs/15066576535)
- [New patch version](https://github.com/Molter73/stackrox-collector/actions/runs/15066636886)
- [New major version](https://github.com/Molter73/stackrox-collector/actions/runs/15066841419)
- [New patch version on older major version](https://github.com/Molter73/stackrox-collector/actions/runs/15066647525)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,7 +37,7 @@ jobs:
             echo "major=${BASH_REMATCH[1]}" >> "$GITHUB_OUTPUT"
             echo "minor=${BASH_REMATCH[2]}" >> "$GITHUB_OUTPUT"
           else
-            echo >&2 "Invalid version ${{ inputs.version }}"
+            echo >&2 "Invalid version ${{ inputs.version }}. The expected format is <Major>.<minor>"
             exit 1
           fi
 
@@ -133,6 +133,26 @@ jobs:
             notice "Master tag" "${MAJOR}.${MINOR}.x"
             notice "Release branch" "release-${MAJOR}.${MINOR}"
           fi
+
+      - name: Mismatched versions
+        if: steps.required-release.outputs.major != 0 && (
+              steps.required-release.outputs.major != steps.final-values.outputs.major ||
+              steps.required-release.outputs.minor != steps.final-values.outputs.minor
+            )
+        env:
+          REQUIRED_MAJOR: ${{ steps.required-release.outputs.major }}
+          REQUIRED_MINOR: ${{ steps.required-release.outputs.minor }}
+          CALCULATED_MAJOR: ${{ steps.final-values.outputs.major }}
+          CALCULATED_MINOR: ${{ steps.final-values.outputs.minor }}
+        run: |
+          cat << EOF >&2
+          ::error title='Version mismatch'::The required version did not match the one calculated. REQUIRED: ${REQUIRED_MAJOR}.${REQUIRED_MINOR}, GOT: ${CALCULATED_MAJOR}.${CALCULATED_MINOR}
+
+          Please review the input and retrigger the workflow.
+          EOF
+
+          # Fail the workflow
+          exit 1
 
   release:
     runs-on: ubuntu-24.04
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,7 +63,7 @@ jobs:
           echo "major=${tag[0]}" >> "$GITHUB_OUTPUT"
           echo "minor=${tag[1]}" >> "$GITHUB_OUTPUT"
 
-      - name: Parse required release
+      - name: Determine release type and version
         id: final-values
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,213 @@
...
+          fetch-depth: 0
+
+      - name: Parse required release
+        id: required-release
+        run: |
+          if [[ "${{ inputs.version }}" =~ ^([[:digit:]]+)\.([[:digit:]]+)$ ]]; then
+            echo "major=${BASH_REMATCH[1]}" >> "$GITHUB_OUTPUT"
+            echo "minor=${BASH_REMATCH[2]}" >> "$GITHUB_OUTPUT"
+          else
+            echo >&2 "Invalid version ${{ inputs.version }}"
```
JoukoVirtanen, line 40:
```suggestion
            echo >&2 "Invalid version ${{ inputs.version }}. The expected format is x.y"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,213 @@
...
+        env:
+          REQUIRED_MAJOR: ${{ steps.required-release.outputs.major }}
+        run: |
+          tag=(0 0)
+          while read -r line; do
+            if [[ "$line" =~ ^([[:digit:]]+)\.([[:digit:]]+)\.x$ ]]; then
+                # If we are doing a release for a specific major
+                # version, we want to limit ourselves to that, so we
+                # ignore newer major versions.
+                if ((tag[0] < BASH_REMATCH[1] && (REQUIRED_MAJOR == 0 || REQUIRED_MAJOR >= BASH_REMATCH[1]))); then
```
JoukoVirtanen, line 55: Why is there special handling for when `REQUIRED_MAJOR` is 0?

.github/workflows/release.yml
```diff
@@ -0,0 +1,213 @@
...
+          type=$3
+          EOF
+          }
+
+          if ((REQUIRED_MAJOR==0)); then
+            add_outputs "${LATEST_MAJOR}" "$((LATEST_MINOR+1))" "minor"
+          elif ((REQUIRED_MAJOR > LATEST_MAJOR)); then
+            add_outputs "$((LATEST_MAJOR+1))" "0" "major"
+          elif ((REQUIRED_MAJOR == LATEST_MAJOR && REQUIRED_MINOR > LATEST_MINOR)); then
+            add_outputs "${LATEST_MAJOR}" "$((LATEST_MINOR+1))" "minor"
```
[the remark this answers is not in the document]
Molter73: I'll add a step for issuing warnings.

.github/workflows/release.yml
Molter73: If it is alright with you, I'll do a small variation of this to make it more clear, something like `The expected format is <Major>.<minor>`

Stringy (approved):
LGTM!

.github/workflows/release.yml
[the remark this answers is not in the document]
JoukoVirtanen: Looks good.

JoukoVirtanen (approved):
I am approving, but I would still like more documentation about using 0.0 as the input version before merging.

.github/workflows/release.yml
JoukoVirtanen: That's fine.

.github/workflows/release.yml
Molter73: Added a bit more context to the file.

Molter73:
Looks like the konflux failure is an outage, so I'm merging the PR.

PR #2130, opened 2025-05-20
ROX-29399: Directional external-IPs

ovalenti:
## Description

Add an attribute to the runtime-configuration to set in which direction the external-IPs are enabled.

The possible options for the new `direction` attribute are `INGRESS`, `EGRESS` and `BOTH`.

By default (when the attribute is missing) the behavior is to enable in both directions.

Example:
```yaml
networking:
  externalIps:
    enabled: DISABLED
    direction: INGRESS
```

The related proto changes are in https://github.com/stackrox/stackrox/pull/15373
## Checklist

**Automated testing**
  - [x] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing

Preliminary tests in https://github.com/stackrox/collector/pull/2140 tend to show that the behavior is as expected.

I did some manual testing and could show that setting `direction: INGRESS` normalized the outgoing connections.


collector/lib/CollectorConfig.cpp
```diff
@@ -545,4 +544,38 @@ std::pair<option::ArgStatus, std::string> CollectorConfig::CheckConfiguration(co
...
+      {CollectorConfig::ExternalIPsConfig::Direction::NONE, "NONE"},
+      {CollectorConfig::ExternalIPsConfig::Direction::INGRESS, "INGRESS"},
+      {CollectorConfig::ExternalIPsConfig::Direction::EGRESS, "EGRESS"},
+      {CollectorConfig::ExternalIPsConfig::Direction::BOTH, "BOTH"},
+  };
+
+  auto str = mapping.find(config.GetDirection());
+
+  return os << "direction(" << ((str == mapping.end()) ? "invalid" : (*str).second) << ")";
+}
```
Molter73, line 579: Would a switch statement be more readable? It also saves hashing on a map when we can just compare an int
```cpp
std::ostream& operator<<(std::ostream& os, const collector::CollectorConfig::ExternalIPsConfig& config) {
  os << "direction(";

  switch (config.GetDirection()) {
      case CollectorConfig::ExternalIPsConfig::Direction::NONE:
        os << "NONE";
        break;
      case CollectorConfig::ExternalIPsConfig::Direction::INGRESS:
        os << "INGRESS";
        break;
      case CollectorConfig::ExternalIPsConfig::Direction::EGRESS:
        os << "EGRESS";
        break;
      case CollectorConfig::ExternalIPsConfig::Direction::BOTH:
        os << "BOTH";
        break;
    default:
        os << "invalid";
        break;
  }

  return os << ")";
}

```

collector/lib/CollectorConfig.cpp
```diff
@@ -545,4 +544,38 @@ std::pair<option::ArgStatus, std::string> CollectorConfig::CheckConfiguration(co
...
+          break;
+        default:
+          direction_enabled_ = Direction::BOTH;
+          break;
+      }
+    }
+  } else if (default_enabled) {
+    direction_enabled_ = Direction::BOTH;
+  }
+}
```
Molter73, line 566: This might be more readable
```cpp
CollectorConfig::ExternalIPsConfig::ExternalIPsConfig(std::optional<sensor::CollectorConfig> runtime_config, bool default_enabled) {
  if (!runtime_config.has_value()) {
    direction_enabled_ = default_enabled ? Direction::BOTH : Direction::NONE;
    return;
  }

  // At this point we know runtime_config has a value, we can access it directly
  const auto& external_ips = runtime_config->networking().external_ips();
  if (external_ips.enabled() != sensor::ExternalIpsEnabled::ENABLED) {
    direction_enabled_ = Direction::NONE;
    return;
  }

  switch (external_ips.direction()) {
    case sensor::ExternalIpsDirection::INGRESS:
      direction_enabled_ = Direction::INGRESS;
      break;
    case sensor::ExternalIpsDirection::EGRESS:
      direction_enabled_ = Direction::EGRESS;
      break;
    default:
      direction_enabled_ = Direction::BOTH;
      break;
  }
}
```

collector/lib/CollectorConfig.cpp
```diff
@@ -545,4 +544,38 @@ std::pair<option::ArgStatus, std::string> CollectorConfig::CheckConfiguration(co
   return std::make_pair(ARG_OK, "");
 }
 
+CollectorConfig::ExternalIPsConfig::ExternalIPsConfig(std::optional<sensor::CollectorConfig> runtime_config, bool default_enabled)
```
Molter73, line 547: At this point, would it be worth it to split `ExternalIPsConfig` to its own source file and simply add an instance of it to `CollectorConfig`? I think that might be a bit more clear.

Molter73: Actually, never mind, we are simply creating a new copy of this type every time we call the `ExternalIPsConf` method in `CollectorConfig`, just move the `ExternalIPsConfig` to separate .h/.cpp files and leave its construction up to `CollectorConfig`.

collector/lib/ConnTracker.cpp
```diff
@@ -138,28 +138,35 @@ void ConnectionTracker::CloseConnections(ConnMap* old_conn_state, ConnMap* delta
...
   } else {
-    CloseExternalUnnormalizedConnections(old_conn_state, delta_conn);
+    CloseExternalUnnormalizedConnections(/* egress is when we are not server */ false, old_conn_state, delta_conn);
+  }
+  if (enable_external_ips_ingress_) {
+    CloseNormalizedConnections(/* ingress is when we are server */ true, old_conn_state, delta_conn);
+  } else {
+    CloseExternalUnnormalizedConnections(/* ingress is when we are server */ true, old_conn_state, delta_conn);
   }
 }
```
Molter73, line 171: This might be more verbose, but I think it's more readable than the current implementation, I'm pretty sure some more iterating can get it back to a reasonable place, WDYT?
```cpp
void ConnectionTracker::CloseConnectionsOnRuntimeConfigChange(ConnMap* old_conn_state, ConnMap* delta_conn) {
  /* egress is when we are not server */
  if (enable_external_ips_egress_) {
    CloseConnections(old_conn_state, delta_conn, [](const Connection* conn) {
      if (!conn->is_server()) {
        return false;
      }
      return Address::IsCanonicalExternalIp(conn->remote().address());
    });
  } else {
    CloseConnections(old_conn_state, delta_conn, [this](const Connection* conn) {
      if (!conn->is_server()) {
        return false;
      }
      return ShouldNormalizeConnection(conn) && !Address::IsCanonicalExternalIp(conn->remote().address());
    });
  }

  /* ingress is when we are not server */
  if (enable_external_ips_ingress_) {
    CloseConnections(old_conn_state, delta_conn, [](const Connection* conn) {
      if (conn->is_server()) {
        return false;
      }
      return Address::IsCanonicalExternalIp(conn->remote().address());
    });
  } else {
    CloseConnections(old_conn_state, delta_conn, [this](const Connection* conn) {
      if (conn->is_server()) {
        return false;
      }
      return ShouldNormalizeConnection(conn) && !Address::IsCanonicalExternalIp(conn->remote().address());
    });
  }
}
```

Molter73 (commented):
Only a partial review, mostly refactorings that would be nice to see

collector/lib/CollectorConfig.cpp
ovalenti: Absolutely, I updated with your proposal :+1:

collector/lib/CollectorConfig.cpp
ovalenti: Applied as well !

collector/lib/CollectorConfig.cpp
ovalenti: I moved the ExternalIPsConfig to its own file.

Molter73 (commented):
Finished going through the PR, changes look good apart from the pending comment in my previous review.

Molter73 (approved):
LGTM!

docs/references.md
```diff
@@ -151,6 +151,36 @@ Whenever the configuration file is updated or created, collector will update
...
+content of the file changes.
+
+The following attributes can be set using the runtime-configuration file :
+
+* `networking.externalIPs.enabled: ENABLED | DISABLED`: enable/disable the external-IPs feature.
+Default value: `DISABLED`
+
+* `networking.externalIPs.direction: INGRESS | EGRESS | BOTH`: when external-IPs are enabled,
+this attribute can restrict the direction in which it is effective. For instance, using `EGRESS`
+will aggregate all the incoming connections and give all details for the outgoing ones. This can
```
Molter73, line 167: Is a bit weird that you mention `EGRESS` but then explain the effect on ingress first, maybe doing it the other way around is easier to understand?
```suggestion
will give all details for all the outgoing connections and aggregate the incoming ones. This can
```

Molter73 (approved):
Still LGTM, just a minor comment on ordering in an example.

JoukoVirtanen:
Please explain what manual testing you did in the PR description.

collector/lib/ExternalIPsConfig.h
```diff
@@ -0,0 +1,40 @@
...
+ public:
+  enum Direction {
+    NONE = 0,
+    INGRESS = 1 << 0,
+    EGRESS = 1 << 1,
+    BOTH = INGRESS | EGRESS,
+  };
+
+  // Are External-IPs enabled in the provided direction ?
+  bool IsEnabled(Direction direction) const { return (direction & direction_enabled_) == direction; }
```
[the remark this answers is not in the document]
JoukoVirtanen: I hate to admit it, but the AI is correct.

collector/lib/ConnTracker.h
```diff
@@ -202,7 +203,7 @@ class ConnectionTracker {
 
   UnorderedSet<Address> known_public_ips_;
   NRadixTree known_ip_networks_;
-  bool enable_external_ips_ = false;
+  ExternalIPsConfig external_IPs_config_;
```
JoukoVirtanen, line 206: [nit]
```suggestion
  ExternalIPsConfig external_ips_config_;
```

collector/lib/CollectorConfig.h
```diff
@@ -97,15 +98,10 @@ class CollectorConfig {
   // EnableExternalIPs will check for the existence
   // of a runtime configuration, and defer to that value
   // otherwise, we rely on the feature flag (env var)
-  bool EnableExternalIPs() const {
+  ExternalIPsConfig ExternalIPsConf() const {
```
JoukoVirtanen, line 101: [nit]
```suggestion
  ExternalIPsConfig GetExternalIPsConfig() const {
```

collector/test/ConnTrackerTest.cpp
```diff
@@ -1651,80 +1651,143 @@ TEST(ConnTrackerTest, TestConnectionStats) {
...
-  EXPECT_THAT(old_state, expected_old_state);
-  EXPECT_THAT(delta, IsEmpty());
+  struct TestStep {
+    ExternalIPsConfig::Direction previous_direction;
+    ExternalIPsConfig::Direction new_direction;
+    ConnMap old_state;
+    ConnMap resulting_old_state;
+    ConnMap expected_delta;
+  } testSteps[] = {
+      {// No change (enabled)
```
JoukoVirtanen, line 1677: There are 16 possible combinations for `previous_direction` and `new_direction`. It would not be unreasonable to have all of them. I added a couple of them in the PR I opened.

collector/lib/ExternalIPsConfig.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <internalapi/sensor/collector.pb.h>
+
+namespace collector {
+
+// Encapsulates the configuration of the External-IPs feature
+class ExternalIPsConfig {
+ public:
+  enum Direction {
+    NONE = 0,
```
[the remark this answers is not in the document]
Molter73: The AI is not aware that `IsEnabled(NONE)` is not something a human being would write, it makes more sense to simply make that invariant explicit in `IsEnabled` as I explained in a comment on your PR. `NONE` should still be 0 and having this additional enum makes it easier to test configurations with bitwise operations, so I would vote for keeping it.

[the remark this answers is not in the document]
ovalenti: There are some reasons why I did not use the proto-defined enum type:
- while the protobuf enum type reflects the `direction` attribute in the yaml file, the enum type in ExternalIPsConfig encompasses the full configuration of external-IPs, including the fact that it is globally enabled or not. Reusing the same enum type could be misleading here.
- the proto enum is part of the implementation choice we made for the configuration. Users of the config object should not have to be aware of this.

Regarding the comment that IsEnabled(NONE) returns always true, I honestly was considering that this would be an undefined behavior. In order to make sure that we don't inadvertently do this, I introduced an assert() statement.

collector/test/ConnTrackerTest.cpp
ovalenti: I cherry-picked/adapted the tests you added on your PR. I will prioritize some manual testing over adding all the combinations for now.

collector/lib/ExternalIPsConfig.h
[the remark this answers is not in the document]
JoukoVirtanen: Those points make sense, but what assert statement are you talking about?

collector/test/ConnTrackerTest.cpp
JoukoVirtanen: Okay.

PR #2142, opened 2025-05-27
Ovalenti/rox 29399 directional ext i ps jv 2

JoukoVirtanen:
## Description

Fixes a problem where `IsEnabled(Direction::NONE)` always returns true.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI is sufficient.

### DIFF
--- a/collector/test/ConnTrackerTest.cpp
+++ b/collector/test/ConnTrackerTest.cpp
@@ -1740,6 +1740,40 @@ TEST(ConnTrackerTest, TestExternalIPsConfigChangeEnableEgress) {
        .expected_delta = {
            {conn_ingress_normalized, ConnStatus(connection_time, false)},
        }},
+      {// Disable EGRESS enable INGRESS
+       .previous_direction = ExternalIPsConfig::Direction::EGRESS,
+       .new_direction = ExternalIPsConfig::Direction::INGRESS,
+       .old_state = {
+           {conn_ingress, ConnStatus(connection_time, true)},
+           {conn_ingress_normalized, ConnStatus(connection_time, true)}, /* closed */
+           {conn_egress, ConnStatus(connection_time, true)} /* closed */,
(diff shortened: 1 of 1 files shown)


collector/test/CollectorConfigTest.cpp
```diff
@@ -155,4 +156,95 @@ TEST(CollectorConfigTest, TestEnableExternalIpsRuntimeConfig) {
   EXPECT_TRUE(config.ExternalIPsConf().IsEnabled(Direction::BOTH));
 }
 
+TEST(CollectorConfigTest, TestIsEnabledIngress) {
```
JoukoVirtanen, line 159: These tests could all be done in one test in a loop.

collector/lib/ExternalIPsConfig.h
```diff
@@ -11,7 +11,7 @@ namespace collector {
 class ExternalIPsConfig {
  public:
   enum Direction {
-    NONE = 0,
+    NONE = 1 << 2,
```
Molter73, line 14: I think it makes more sense to change `IsEnabled` like this:
```cpp
bool IsEnabled(Direction direction) const { return direction != Direction::None && (direction & direction_enabled_) == direction; }
```

On a different note, I don't think `IsEnabled(Direction::NONE)` is something anyone would write, so I'm not convinced the extra overhead for the check is worth it. Maybe we can do a static check instead?
```cpp
bool IsEnabled(Direction direction) const {
  static_assert(direction != Direction::NONE);
  return (direction & direction_enabled_) == direction;
}
```

Or, if a compile time check is not possible, then just use good old `assert` to make the invariant explicit.
```cpp
bool IsEnabled(Direction direction) const {
  assert(direction != Direction::NONE);
  return (direction & direction_enabled_) == direction;
}
```

PR #2190, opened 2025-06-23
fix(ansible): use inventory_hostname for s390x destruction

Stringy:
## Description

Connecting to the VM to get the hostname is redundant, and can cause issues where VMs are not deleted if we are unable to SSH to them.


Molter73 (approved):
Makes sense, thanks for making this change!

PR #2196, opened 2025-06-25
Update Falco libs to 0.21.0

Molter73:
## Description

The latest version of Falco has a number of changes that are incompatible with collector, biggest ones are:
- Removal of the container manager code in favor of a plugin.
- Major refactoring of sinsp.

In order to make collector compatible again, we had to drop the ContainerEngine that we implemented in favor of a method in the event extractor that will get the container id from the cgroups when it is called. The ContainerMetadata is also essentially dead in the water, since we can't get container metadata without the container plugin.

Filtering of events that used to happen in the inspector itself has been moved into collector, since we can't filter events by container id without the container engine.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

TODO(replace-me)


Molter73:
Superseded by #2976

PR #2297, opened 2025-07-29
Bump Falco commit with 6.15 kernel fix

Stringy:
## Description

Includes fix for 6.15 kernel changes, which have been causing CI failures in fedora-coreos.


Molter73 (approved):
Approving but we should probably merge the Falco PR and update this branch before merging here.

PR #2326, opened 2025-08-05
backport: Use P and Z GHA runners (#2215)

Stringy:
Backport P&Z runners

### DIFF
--- a/.github/workflows/collector-builder.yml
+++ b/.github/workflows/collector-builder.yml
@@ -8,6 +8,7 @@ on:
         required: true
         description: |
           The tag used to build the collector image
+
     outputs:
       collector-builder-tag:
         description: The builder tag used by the build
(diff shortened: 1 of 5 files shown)


Molter73 (commented):
This will not be as straightforward, on top of #2215 you will need #2106 and #2210, as seen in the backport to 3.21: https://github.com/stackrox/collector/pull/2254/commits

Those other commits probably need even more cherry-picks for them to work. Unless you desperately need this on 3.20, I would just give up and deal with the 8 hour builds for a couple more weeks/months until it goes EOL (heck, that's what I did).

ovalenti:
@Stringy, @Molter73 found out why the Z VM was failing to checkout (something to do with annotated tags), and so far the build is going smoothly :crossed_fingers:

PR #2329, opened 2025-08-06
Downgrade network status notifier messages to trace

erthalion:
## Description

Currently two log messages are at debug level, which makes this level useless, since information in it is covered by an avalanche of repeating network status notifier messages. Downgrade to trace.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Run Collector manually with the new level.


erthalion:
> we probably want to have a per-component level at some point ?
Yeah, that would be cool.

PR #2338, opened 2025-08-08
Specify registry for nginx (#2321)

erthalion:
## Description

Currently no registry is set for nginx, with leads to a few issues:

* It's being pulled from docker.io and sometimes bails out due to rate limiting.
* With podman it can cause an issue with short-name resolution:

    ```
    short-name resolution enforced but cannot prompt without a TTY
    ```

Promote nginx from a non_qa to qa image.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests


JoukoVirtanen:
Why was `GetContainerPort` changed to `GetContainerPorts` in the master branch, but not this branch?

erthalion:
> Why was GetContainerPort changed to GetContainerPorts in the master branch, but not this branch?

It's a refactoring, which has nothing to do in a back-patch PR.

> QA_TAG was changed to 2.0.3 in master and here. If the desired images for 3.20 and master have differences, could that lead to problems?

Theoretically yes, practically this would be visible in broken CI tests. No failures, no problem.

JoukoVirtanen (approved):
LGTM!

PR #2358, opened 2025-08-18
Miscellaneous resilience improvements

Molter73:
## Description

Some improvements have been made to our ansible playbooks, namely:
- Remove `us-central1-c` zone, since this one doesn't have Arm VM support.
- Update ansible to use the newer `break_when` loop control, this is closer to our original intent.
- Ensure the compute instance is removed when creation fails.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI should be enough.

### DIFF
--- a/ansible/ci/group_vars/all.yml
+++ b/ansible/ci/group_vars/all.yml
@@ -7,6 +7,7 @@ gcp_instance_prefix: ci
 gcp_available_zones:
   - us-central1-a
   - us-central1-b
+  - us-central1-d
   - us-central1-f
 ansible_user: runner
 gcp_network_name: collector-ci


ansible/ci/group_vars/all.yml
```diff
@@ -7,7 +7,6 @@ gcp_instance_prefix: ci
 gcp_available_zones:
   - us-central1-a
   - us-central1-b
-  - us-central1-c
   - us-central1-f
```
robbycochran, line 10:
```suggestion
  - us-central1-f
  - us-central1-d
```

robbycochran:
```
$ gcloud compute machine-types list --filter="name=t2a-standard-2 AND zone~'^us-'" --format="value(zone)"
us-central1-a
us-central1-b
us-central1-f
us-central1-d
```

Molter73:
So the changes in this PR are not really doing what I intended them to do.

The current state on master is that ansible will create a VM on `us-central1-a` attempt to start it and fail at this point, leaving the VM created and causing the subsequent zones to fail.

The intention of this PR was to attempt to remove the VM once it fails, this can be achieved with a `block`, but you can't combine loops and blocks in ansible, so I created a separate file with tasks and attempted to include that one in a loop, when doing this, the task will be added once per zones and run concurrently, which means whichever creates the VM first will be used and all others will fail. I believe this will not solve our initial problem, since the actual error happens when the VM is attempted to be started and can still cause the same issue we are seeing.

PR #2501, opened 2025-09-17
ROX-30928: Disable garden-linux integration tests

ovalenti:
## Description

Disable garden-linux:
- the kernel bumper is broken and difficult to fix (for more than a month)
- process-listening-on-ports tests are failing for a reason specific to garden-linux

Let's simply disable the tests.

## Checklist

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -59,7 +59,7 @@ jobs:
           - rhel-sap
           # TODO(ROX-30765): re-enable once provisioning is fixed
           # - sles
-          - garden-linux
+          # - garden-linux
     with:
       vm_type: ${{ matrix.vm_type }}
       collector-tag: ${{ inputs.collector-tag }}


.github/workflows/integration-tests.yml
```diff
@@ -59,7 +59,7 @@ jobs:
           - rhel-sap
           # TODO(ROX-30765): re-enable once provisioning is fixed
           # - sles
-          - garden-linux
+          # - garden-linux
```
JoukoVirtanen, line 62: Perhaps delete all together.

JoukoVirtanen (approved):
Left a suggestion, but feel free to ignore and just merge.

.github/workflows/integration-tests.yml
ovalenti: I had the same temptation, but it is interesting to make it easy restore the test locally. Also, deleting the top level entry would mean to delete all the other garden-linux bits (ansible).

PR #2540, opened 2025-09-24
Remove SSH access to builder container

Molter73:
## Description

Z builds are currently failing when building the builder image with the following error:

```
Dockerfile:72
--------------------
  71 |     # Add remote development user
  72 | >>> RUN useradd -m remoteuser \
  73 | >>>   && yes c0llectah | passwd remoteuser || if [[ $? -eq 141 ]]; then true; else exit $?; fi
  74 |
--------------------
ERROR: failed to build: failed to solve: process "/bin/sh -c useradd -m remoteuser   && yes c0llectah | passwd remoteuser || if [[ $? -eq 141 ]]; then true; else exit $?; fi" did not complete successfully: exit code: 1
gmake: *** [Makefile:27: builder] Error 1
```

In the past, SSH was used with CLion for development, but devcontainers are now properly supported by that IDE, so there is no point in keeping the SSH access method in the builder container.

The documentation section on developing with CLion has been removed since it is no longer supported. I have not documented how to use a devcontainer with that environment because I have no idea how that works and don't have access to CLion for testing either, contributions to add the section back in are welcomed.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

- [x] Z builder build succeeds in CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,9 +20,6 @@ cmake-build-*/
 # Vim swap files
 *.swp
 
-# Development SSH Server keys
-.collector_dev_ssh_host_ed25519_key*
-
 # vscode configuration files
 .vscode/
 .devcontainer/
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,6 @@ include Makefile-constants.mk
 
 NPROCS ?= $(shell nproc)
 
-DEV_SSH_SERVER_KEY ?= $(CURDIR)/.collector_dev_ssh_host_ed25519_key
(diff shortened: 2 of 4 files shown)


Stringy (approved):

ovalenti (approved):
TIL: by default, a docker image inherits its parent CMD value

Thanks for the cleanup :fist_oncoming:

PR #2976, opened 2026-02-25
ROX-33614: Update Falco to 0.23.1

Stringy:
## Description

Falco PR: https://github.com/stackrox/falcosecurity-libs/pull/97 

The changes in this PR relate to the uplift of Falco to the latest tagged version, 0.23.1. This is a significant upgrade, jumping from 0.18.1 up to this latest version and contains the following notable changes to Collector:

* Added TOCTOU BPF programs to exclude list (we are not interested in these syscalls and they caused verifier issues)
* Moved Container ID parsing into Collector
  * Perhaps the most significant change in upstream Falco is the container plugin, which is a Go shared library that must be loaded at runtime and dynamically populates fields for each process. The complexity of such a plugin in terms of building, loading, and utilising it was deemed a little too much for the small amount of data we actually need from it. As a result, we don't use the plugin at all.
  * One side effect of this is we don't have k8s namespaces automatically populated, which means in the future if/when we want namespace-level runtime configuration we may have to revisit these changes and either implement this in Collector or resign ourselves to using the container plugin.
* The container filter has been changed from "container.id != 'host'" to "proc.pid != proc.vpid" - these are functionally identical but the former is not usable because we're not using the container plugin. Strictly speaking the new filter does not work on (very) old kernels e.g. RHEL 7, but we do not need to support them.
* Numerous API changes to interact correctly with the new Falco version. 

This work was performed primarily by Claude, with oversight from me. This was something of an experiment and so I got Claude to do as much of the work as possible, with my role being that of the driver; steering Claude when it got muddled or hyper focused on the wrong fix.

The update and rebase itself was relatively straight forward, resulting in new locally-built and locally-verified builds in just a couple of hours (i.e. builds that built and ran locally, passing the integration tests on my Fedora 42 x86 machine.) The bottle neck became the CI and getting that work fed back into Claude to perform the diagnostics and fixes.

As a result, I have added two Claude skills as part of this PR which allow Claude to (1) perform Falco updates in the future and (2) inspect the state of CI and investigate test failures or BPF verifiers issues.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Built and tested locally, all unit & integration tests passing. CI handles the remainder of our test matrix.


.claude/commands/analyze-ci.md
Molter73, line 1: Can we move the `.claude/commands` and `AGENTS.md` to a separate PR so we can focus the review efforts?

collector/lib/NetworkSignalHandler.cpp
```diff
@@ -68,24 +69,30 @@ NetworkSignalHandler::~NetworkSignalHandler() = default;
...
-      // connect() failed or getsockopt(SO_ERROR) returned a failure
-      return std::nullopt;
-    }
-
-    if (fd_info->is_socket_pending()) {
-      // connect() returned E_INPROGRESS
-      return std::nullopt;
+    bool is_send_recv = (strncmp(evt_name, "send", 4) == 0 ||
+                         strncmp(evt_name, "recv", 4) == 0);
+    if (!is_send_recv) {
```
Molter73: I'm fairly certain directly doing the checks is faster than doing these string checks. If we don't want to do this check for process information, I'm fairly certain we can check the syscall number instead and that will also be faster. Is there some other reason this check would be here I'm not seeing?

collector/lib/system-inspector/EventExtractor.cpp
```diff
@@ -5,6 +5,10 @@ namespace collector::system_inspector {
 void EventExtractor::Init(sinsp* inspector) {
   for (auto* wrapper : wrappers_) {
     std::unique_ptr<sinsp_filter_check> check = FilterList().new_filter_check_from_fldname(wrapper->event_name, inspector, true);
+    if (!check) {
+      CLOG(WARNING) << "Filter check not available for field: " << wrapper->event_name;
+      continue;
+    }
```
Molter73: Fairly certain this check should just be an `assert`.

collector/lib/system-inspector/EventExtractor.h
```diff
@@ -41,9 +41,12 @@ class EventExtractor {
 #define FIELD_RAW(id, fieldname, type)                                                                     \
  public:                                                                                                   \
   const type* get_##id(sinsp_evt* event) {                                                                 \
-    uint32_t len;                                                                                          \
-    auto buf = filter_check_##id##_->extract_single(event, &len);                                          \
-    if (!buf) return nullptr;                                                                              \
+    if (!filter_check_##id##_.filter_check) return nullptr;                                                \
+    std::vector<extract_value_t> vals_##id;                                                                \
+    if (!filter_check_##id##_->extract(event, vals_##id)) return nullptr;                                  \
```
Molter73: Why are we using `extract` when the falco PR still has the definition of `extract_single`?

collector/lib/system-inspector/EventExtractor.h
```diff
@@ -41,9 +41,12 @@ class EventExtractor {
 #define FIELD_RAW(id, fieldname, type)                                                                     \
  public:                                                                                                   \
   const type* get_##id(sinsp_evt* event) {                                                                 \
-    uint32_t len;                                                                                          \
-    auto buf = filter_check_##id##_->extract_single(event, &len);                                          \
-    if (!buf) return nullptr;                                                                              \
+    if (!filter_check_##id##_.filter_check) return nullptr;                                                \
```
Molter73: Should this be an `assert`?

collector/lib/system-inspector/EventExtractor.h
```diff
@@ -63,9 +66,12 @@ class EventExtractor {
   const std::optional<type> get_##id(sinsp_evt* event) {                                                   \
     static_assert(std::is_trivially_copyable_v<type>,                                                      \
                   "Attempted to create FIELD_RAW_SAFE on non trivial type");                               \
-    uint32_t len;                                                                                          \
-    auto buf = filter_check_##id##_->extract_single(event, &len);                                          \
-    if (!buf) return {};                                                                                   \
+    if (!filter_check_##id##_.filter_check) return {};                                                     \
```
Molter73: Should this be an assert?

collector/lib/system-inspector/EventExtractor.h
```diff
@@ -63,9 +66,12 @@ class EventExtractor {
...
+    std::vector<extract_value_t> vals_##id;                                                                \
+    if (!filter_check_##id##_->extract(event, vals_##id)) return {};                                       \
```
Molter73: `extract_single` still exists.

collector/lib/system-inspector/EventExtractor.h
```diff
@@ -80,39 +86,40 @@ class EventExtractor {
...
-    auto buf = filter_check_##id##_->extract_single(event, &len); \
-    if (!buf) return nullptr;                                     \
-    return reinterpret_cast<const char*>(buf);                    \
-  }                                                               \
-                                                                  \
- private:                                                         \
+#define FIELD_CSTR(id, fieldname)                                         \
+ public:                                                                  \
+  const char* get_##id(sinsp_evt* event) {                                \
+    if (!filter_check_##id##_.filter_check) return nullptr;               \
```
Molter73: assert?

collector/lib/system-inspector/EventExtractor.h
```diff
@@ -80,39 +86,40 @@ class EventExtractor {
...
+    std::vector<extract_value_t> vals_##id;                               \
+    if (!filter_check_##id##_->extract(event, vals_##id)) return nullptr; \
```
Molter73: extract_single?

collector/lib/system-inspector/Service.cpp
```diff
@@ -62,32 +55,22 @@ Service::Service(const CollectorConfig& config)
...
-
-    // k8s naming conventions specify that max length be 253 characters
-    // (the extra 2 are just for a nice 0xFF).
-    inspector_->set_container_labels_max_len(255);
-  } else {
-    auto engine = std::make_shared<ContainerEngine>(inspector_->m_container_manager);
-    auto* container_engines = inspector_->m_container_manager.get_container_engines();
-    container_engines->push_back(engine);
+  default_formatter_ = std::make_unique<sinsp_evt_formatter>(
+      inspector_.get(), DEFAULT_OUTPUT_STR, EventExtractor::FilterList());
```
Molter73: Why was this moved from the initializer list to this point in the constructor?

collector/lib/system-inspector/Service.cpp
```diff
@@ -62,32 +55,22 @@ Service::Service(const CollectorConfig& config)
...
+
+  // Filter out host processes. In containers, pid != vpid due to PID
+  // namespacing. This is a built-in sinsp field that doesn't require
+  // any plugin.
```
Molter73: What happens if you run a container in the host PID namespace? Do we completely ignore it as well? Might be worth adding an integration test for this.

collector/lib/system-inspector/Service.cpp
```diff
@@ -62,32 +55,22 @@ Service::Service(const CollectorConfig& config)
...
+  try {
+    auto factory = std::make_shared<sinsp_filter_factory>(
+        inspector_.get(), EventExtractor::FilterList());
+    sinsp_filter_compiler compiler(factory, "proc.pid != proc.vpid");
+    inspector_->set_filter(compiler.compile(), "proc.pid != proc.vpid");
+  } catch (const sinsp_exception& e) {
+    CLOG(WARNING) << "Could not set container filter: " << e.what()
+                  << ". Container filtering will not be active.";
```
Molter73: Since ACS has no idea what to do with processes outside of a container, it might be better to just let the exception bubble up and kill collector, rather than flood sensor with events it will not process.

collector/lib/Process.cpp
```diff
@@ -32,7 +33,11 @@ std::string Process::container_id() const {
   WaitForProcessInfo();
 
   if (system_inspector_threadinfo_) {
-    return system_inspector_threadinfo_->m_container_id;
+    for (const auto& [subsys, cgroup_path] : system_inspector_threadinfo_->cgroups()) {
+      if (auto id = ExtractContainerIDFromCgroup(cgroup_path)) {
+        return std::string(*id);
+      }
+    }
```
Molter73: Why can't we use `GetContainerID` here?

collector/lib/ProcessSignalFormatter.cpp
```diff
@@ -190,7 +193,7 @@ ProcessSignal* ProcessSignalFormatter::CreateProcessSignal(sinsp_evt* event) {
   }
 
   CLOG(DEBUG) << "Process (" << signal->container_id() << ": " << signal->pid() << "): "
-              << signal->name() << "[" << container_metadata_.GetNamespace(event) << "] "
+              << signal->name()
```
Molter73, line 194:
```suggestion
              << signal->name() << "[" << container_id << "] "
```

collector/lib/Utility.cpp
```diff
@@ -57,9 +57,29 @@ const char* SignalName(int signum) {
...
+  if (!tinfo) {
+    return {};
+  }
+  return GetContainerID(*tinfo);
+}
+
 std::ostream& operator<<(std::ostream& os, const sinsp_threadinfo* t) {
   if (t) {
-    os << "Container: \"" << t->m_container_id << "\", Name: " << t->m_comm << ", PID: " << t->m_pid << ", Args: " << t->m_exe;
+    os << "Name: " << t->m_comm << ", PID: " << t->m_pid << ", Args: " << t->m_exe;
```
Molter73: Fairly certain you can do something like this:
```suggestion
    os << "Container: \"" << (t == nullptr ? "null" : GetContainerID(*t)) << "\", Name: " << t->m_comm << ", PID: " << t->m_pid << ", Args: " << t->m_exe;
```

Or just create an overloaded version of `GetContainerID` that takes a pointer, checks it, then calls the version with the reference.

integration-tests/suites/udp_networkflow.go
Molter73, line 1: Why do the timeouts in this test need to be bumped from 5 to 30 seconds?

Molter73 (commented):
I haven't gone through the changes under `.claude` or `AGENTS.md`, I think those probably should moved to their own PR.

Even without those there are plenty of comments, the falco side looks fine though.

collector/lib/system-inspector/EventExtractor.cpp
```diff
@@ -5,6 +5,9 @@ namespace collector::system_inspector {
 void EventExtractor::Init(sinsp* inspector) {
   for (auto* wrapper : wrappers_) {
     std::unique_ptr<sinsp_filter_check> check = FilterList().new_filter_check_from_fldname(wrapper->event_name, inspector, true);
+    if (!check) {
+      CLOG(FATAL) << "Filter check not available for field: " << wrapper->event_name;
+    }
```
Molter73: [nitpick]
```suggestion
    CLOG_IF(!check, FATAL) << "Filter check not available for field: " << wrapper->event_name;
```

collector/lib/system-inspector/Service.cpp
```diff
@@ -62,31 +59,13 @@ Service::Service(const CollectorConfig& config)
...
-  }
-
-  inspector_->set_filter("container.id != 'host'");
+  // Filter out host processes. In containers, pid != vpid due to PID
+  // namespacing. The val() transformer is required so the parser treats
+  // proc.vpid as a field reference rather than a bare string value.
+  auto factory = std::make_shared<sinsp_filter_factory>(
+      inspector_.get(), EventExtractor::FilterList());
+  sinsp_filter_compiler compiler(factory, "proc.pid != val(proc.vpid)");
+  inspector_->set_filter(compiler.compile(), "proc.pid != val(proc.vpid)");
```
Molter73: This should be equivalent:
```suggestion
  inspector_->set_filter("proc.pid != val(proc.vpid)");
```

collector/lib/system-inspector/Service.cpp
Molter73: I think this may still need addressing, was a manual test at least run to verify if the behavior is consistent?

collector/lib/NetworkSignalHandler.cpp
```diff
@@ -75,17 +75,23 @@ std::optional<Connection> NetworkSignalHandler::GetConnection(sinsp_evt* evt) {
...
+    bool is_send_recv = (type >= PPME_SOCKET_SENDTO_E && type <= PPME_SOCKET_RECVFROM_X) ||
+                        (type >= PPME_SOCKET_SENDMSG_E && type <= PPME_SOCKET_RECVMMSG_X);
+    if (!is_send_recv) {
+      if (fd_info->is_socket_failed()) {
+        return std::nullopt;
+      }
+
+      if (fd_info->is_socket_pending()) {
+        return std::nullopt;
+      }
```
Molter73: Might want to do something closer to what falco does:
```suggestion
    switch (evt->get_type()) {
      case PPME_SOCKET_SENDTO_E:
      case PPME_SOCKET_SENDTO_X:
      case PPME_SOCKET_RECV_E:
      case PPME_SOCKET_RECV_X:
      case PPME_SOCKET_RECVFROM_E:
      case PPME_SOCKET_RECVFROM_X:
      case PPME_SOCKET_SENDMSG_E:
      case PPME_SOCKET_SENDMSG_X:
      case PPME_SOCKET_SENDMMSG_E:
      case PPME_SOCKET_SENDMMSG_X:
      case PPME_SOCKET_RECVMSG_E:
      case PPME_SOCKET_RECVMSG_X:
      case PPME_SOCKET_RECVMMSG_E:
      case PPME_SOCKET_RECVMMSG_X:
        if (fd_info->is_socket_failed()) {
          return std::nullopt;
        }

        if (fd_info->is_socket_pending()) {
          return std::nullopt;
        }
        break;
      default:
        break;
```

It's more verbose, but also more explicit about what operations should make these checks. The compiler should still be able to optimize the case conditions to range checks.

collector/lib/ProcessSignalFormatter.cpp
```diff
@@ -2,6 +2,8 @@
 
 #include <uuid/uuid.h>
 
+#include <libsinsp/sinsp.h>
```
Molter73: Oh lord, no, please let's not start polluting the code base with sinsp headers again, we even have a proposal to stop doing this (#3502).

AFAICT, this is only needed for the `sinsp_thread_manager` definition, can we use a forward definition in the header file like we do with `sinsp` and `sinsp_threadinfo`?

collector/lib/ProcessSignalFormatter.cpp
Molter73: Actually, I'm an idiot, the container ID is clearly printed in the line above this one 🤦🏻 
Can we go back to not having this here?

collector/lib/ProcessSignalFormatter.cpp
Molter73: Actually, I tried locally removing it and it just compiled, so maybe it is not needed at all.

Molter73 (commented):
Overall looks good, I would like to get confirmation on the sinsp header removal comment and I believe there is still one comment from my previous review that has not been addressed.

collector/lib/system-inspector/Service.cpp
Stringy: Yeah sorry, missed this. I've done some checks and it looks like this _does_ introduce a gap where a container is in the host pid namespace. I need to do some more investigation here.

collector/lib/system-inspector/Service.cpp
```diff
@@ -182,10 +182,14 @@ bool Service::FilterEvent(const sinsp_threadinfo* tinfo) {
...
-       tinfo->m_exepath == "/usr/bin/runc") &&
-      tinfo->m_comm == "6") {
+  // Exclude host processes that leak through the sinsp filter.
+  // The sinsp filter uses a cgroup regex to catch containers in
+  // the host PID namespace, but this can also match container
+  // runtime helpers (crun, runc, conmon, podman) that run on the
+  // host within cgroup paths containing container IDs. Checking
+  // GetContainerID catches all such cases without maintaining a
+  // list of runtime helper names.
+  if (GetContainerID(*tinfo).empty()) {
```
Molter73, line 192: I'm fairly certain most runtimes work by forking a process into the container (meaning they have a container ID), are we 100% sure we don't need any of these checks anymore?

collector/lib/ProcessSignalFormatter.cpp
```diff
@@ -341,13 +342,13 @@ void ProcessSignalFormatter::GetProcessLineage(sinsp_threadinfo* tinfo,
...
-    // not contain namespace information, so `m_container_id` is checked
-    // instead. `m_container_id` is not enough on its own to identify
+    // not contain namespace information, so the container ID is checked
+    // instead. The container ID is not enough on its own to identify
     // containerized processes, because it is not guaranteed to be set on
     // all platforms.
     //
     if (pt->m_vpid == 0) {
-      if (pt->m_container_id.empty()) {
+      if (GetContainerID(*pt).empty()) {
```
Molter73, line 351: This is why we should try to setup our own container plugin, when visiting the process lineage we are computing the container ID for each process every time, with a plugin we could just cache the result and use that instead like Falco did before this update.

Not a blocker, just something to keep in mind.

collector/lib/system-inspector/Service.cpp
Molter73: Seems to be addressed in the latest commit.

integration-tests/suites/udp_networkflow.go
Molter73: This question seems to be pending still

Molter73 (approved):
Just a few more small comments, nothing that should prevent merging this.

integration-tests/suites/udp_networkflow.go
Stringy: just fiddled with this, and the raised timeouts have been reverted. the thing that does affect this test, however, is reducing the number of messages from 8 -> 4. Without that change, the test is incredibly flaky.

collector/lib/system-inspector/Service.cpp
Stringy: the new host process testing covers this quite well, and ensures we're only getting the expected in-container processes and nothing else. I think the old checks were a bit too specific anyway, so this should be a little more thorough.

PR #3270, opened 2026-04-24
chore(deps): Bump gRPC to v1.80.0

vladbologa:
## Description

Update gRPC from v1.67.0 to v1.80.0, along with its required dependency bumps: 
* `protobuf` v28.3 to v31.1
* `abseil-cpp` 20240722.0 to 20250512.1.

The `abseil-cpp` bump triggers [GCC bug #71962](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=71962) (tracked in [abseil#1634](https://github.com/abseil/abseil-cpp/issues/1634)): `-fsanitize=undefined` breaks `constexpr` evaluation of function pointer comparisons in abseil's `hash_policy_traits.h`, causing a build failure in `ConfigLoaderTest.cpp` when building with address sanitizer enabled. Worked around by disabling UBSan for that file only.

Adapt collector code to protobuf v31 breaking changes:
- `Arena::CreateMessage<T>()` removed, replaced with `Arena::Create<T>()`
- `google::protobuf::uint64` removed, replaced with `uint64_t`
- `FieldDescriptor::name()` now returns `absl::string_view` instead of `const std::string&`

This bump will make it easier to upgrade to v1.81.0 when it'll be available (which will include required PQC fixes). It could also be merged before https://github.com/stackrox/collector/pull/3269, since in that PR I used some APIs that are now deprecated.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI is sufficient.


collector/lib/ConfigLoader.cpp
```diff
@@ -141,7 +141,7 @@ ParserResult ParserYaml::ParseArrayEnum(google::protobuf::Message* msg, const YA
     const EnumValueDescriptor* value = desc->FindValueByName(enum_name);
     if (value == nullptr) {
       ParserError err;
-      err << file_ << ": Invalid enum value '" << enum_name << "' for field " << (read_camelcase_ ? SnakeCaseToCamel(field->name()) : field->name());
+      err << file_ << ": Invalid enum value '" << enum_name << "' for field " << (read_camelcase_ ? SnakeCaseToCamel(std::string(field->name())) : std::string(field->name()));
```
Molter73, line 144: This is not a huge deal because it is in the error side of a configuration reload which should not happen very often, but do we need to build `std::string` because the `field->name()` method is returning an abseil string_view? Can we create `std::string_view` instead so we don't have to copy the entire string over?

collector/lib/ConfigLoader.cpp
```diff
@@ -202,43 +202,41 @@ ParserResult ParserYaml::Parse(google::protobuf::Message* msg, const YAML::Node&
                                const google::protobuf::FieldDescriptor* field, const std::string& path) {
   using namespace google::protobuf;
 
-  std::string camel;
-  const std::string* name = &field->name();
+  std::string name(field->name());
```
Molter73, line 205: Similar to my previous comment.

collector/lib/ConfigLoader.cpp
vladbologa: Yes, `field->name()` used to return `std::string`, and after the protobuf bump it's a `absl:string_view` (which I think it's just a typedef to `std::string_view`).

But I'd also have to modify `SnakeCaseToCamel` to take `std::string_view` instead of `const std::string&`. Should I do that?

collector/lib/ConfigLoader.cpp
vladbologa: Also here, `SnakeCaseToCamel`, `ConcatPath` and `ParseScalar` should then be modified to use `std::string_view` instead of `const std::string&`.

And it looks like we'd still need a copy on the `read_camelcase_` path:
```
  std::string name(field->name());
  if (read_camelcase_) {
    name = SnakeCaseToCamel(name);
  }
```

collector/lib/ConfigLoader.cpp
Molter73: I think it could be worth it, right now the case where `read_camelcase_` is true creating a new `std::string` before calling `SnakeCaseToCamel` and this in turn will create a separate string with the case changed (the compiler might be smart enough to not do these multiple allocations, but I would rather not rely on it).

We still need to create a `std::string` for the false case because `SnakeCaseToCamel` needs to return that type and the compiler will get upset if we have different types as the result of the branches in a ternary operator.

I'll leave it up to you, I think I'm just putting too much effort on a piece of code that will most likely not be hit and is not in a hot path.

collector/lib/ConfigLoader.cpp
Molter73: Same as my answer to the previous comment, I'll leave it up to you, it might not be worth the effort since this is not code that is on any hot path.

collector/lib/ConfigLoader.cpp
vladbologa: I'm also not sure if it's worth it.  I can make the changes in a follow-up PR and see how they look.

Because I'm changing the builder image, it looks like it literally takes one day to get CI to pass, so I wouldn't tinker with this one too much anymore, unless I have to. 😄

Molter73: > I'm also not sure if it's worth it. I can make the changes in a follow-up PR and see how they look.

Sounds good, thanks!



> Because I'm changing the builder image, it looks like it literally takes one day to get CI to pass,

Weird, I thought we had managed to get the builds down to about an hour, I'll take a quick look and see if we need to do something about this.

Molter73: So I just checked [this build](https://github.com/stackrox/collector/actions/runs/24987043570/attempts/2?pr=3270) that says it took 8 hours to run, but the builder jobs are at about 45 minutes, the collector build is about 10 minutes and the tests are 45 minutes at the longest, so worst case scenario the full run time should have been about 2 hours, not 8.

I'll post on slack so we can discuss with the team, there might be some known issue with some of the runners we are using I'm not aware of.

Molter73 (approved):

PR #3296, opened 2026-04-29
fix: don't fail integration test jobs on artifact upload errors

Stringy:
## Description

The Store artifacts step uploads debug logs for human inspection but is not consumed by any downstream job. Network errors (ECONNRESET) during upload were causing passing test runs to be marked as failed — 9 false positives in the last week alone.

Test failures still fail the job via the Run Tests step.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Trivial CI fix.

### DIFF
--- a/.github/workflows/integration-tests-vm-type.yml
+++ b/.github/workflows/integration-tests-vm-type.yml
@@ -142,6 +142,10 @@ jobs:
 
       - name: Store artifacts
         if: always()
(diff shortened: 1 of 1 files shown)


Molter73 (approved):
LGTM!

Could you make this same change in stackrox/fact? I've seen this same problem a bunch but have not had time to look into it.

PR #3328, opened 2026-05-05
chore: Bump Konflux image expiration to 52 weeks

msugakov:
## Description

Similar to: https://github.com/stackrox/scanner/pull/3147
Related thread: https://redhat-internal.slack.com/archives/CELUQKESC/p1777919339423939

If Konflux-built images expire, they are gone and this can be a problem when COLLECTOR_VERSION is not updated for too long. Bumping the image expiration should make that less probable.

I'm open to bargain on the duration (to make it lower)!

It looks like GHA-built images aren't set to expire at all. At least, I wasn't able to find the `quay.expires-after` label.

## Checklist

**Automated testing**
No change.

## Testing Performed

None. Should just work.

### DIFF
--- a/.tekton/collector-build.yaml
+++ b/.tekton/collector-build.yaml
@@ -27,7 +27,7 @@ spec:
   - name: git-url
     value: '{{source_url}}'
   - name: image-expires-after
-    value: '13w'
+    value: '52w'
   - name: output-image-repo
     value: quay.io/rhacs-eng/release-collector
   - name: path-context


Molter73:
> I'm open to bargain on the duration (to make it lower)!

I think a year is probably fine, we try to keep the collector image updated in the main repo to catch any problems we might have early.

We could also consider setting the automated bump version PRs to auto-merge in the main repo.

Molter73 (approved):

PR #3381, opened 2026-05-20
Add Claude AI test failure analysis to Slack notifications

robbycochran:
## Summary

Adds claude CI failure analysis that runs before posting failure to slack.

### DIFF
--- a/.github/scripts/README.md
+++ b/.github/scripts/README.md
@@ -1,0 +1,232 @@
+# Test Failure Analysis with Claude
+
+Automatically analyzes test failures using Claude AI and includes intelligent insights in Slack notifications.
+
+## Architecture
+
+```
+Integration Tests Run
+  ├── amd64-integration-tests (may fail)
+  ├── arm64-integration-tests (may fail)
+  ├── s390x-integration-tests (may fail)
+  ├── ppc64le-integration-tests (may fail)
+  │
+  ├── collect-failures
+  │    └── Determine which jobs failed
+  │
+  └── analyze-and-notify (reusable workflow)
+       ├── analyze-failures
+       │    ├── Download test artifacts
+       │    ├── Execute /analyze-test-failures skill
+       │    └── Upload analysis-report.md
+       │
+       └── notify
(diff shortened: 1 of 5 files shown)


.github/workflows/analyze-and-notify.yml
Molter73, line 1: I'm confused, why do we need a completely separate workflow for this? The integration-tests one already has GCP authenticated, it has all the reports downloaded... Can't we just add the couple steps that do the analisis in there and update the notify step? If you want the analisis step to be reusable I'd suggest using an action instead of a full on workflow, but I don't see anything particularly complicate, I would venture to say all you need is to add the `Analyze test failures with Claude` step into the integration tests.

.github/scripts/README.md
Molter73, line 1: Why is this in `.github/scripts`? Also, is this just a description of what `analyze-test-failures.md` does? Do we need a 200+ lines of markdown to explain what a separate 100+ line markdown file does?

.github/workflows/analyze-and-notify.yml
```diff
@@ -0,0 +1,136 @@
...
+          credentials_json: '${{ secrets.GCP_CLAUDE_SERVICE_ACCOUNT_KEY }}'
+        continue-on-error: true
+
+      - name: Analyze test failures with Claude
+        id: analyze
+        uses: anthropics/claude-code-base-action@beta
+        continue-on-error: true
+        env:
+          ANTHROPIC_VERTEX_PROJECT_ID: ${{ secrets.GCP_CLAUDE_PROJECT_ID }}
+          CLOUD_ML_REGION: us-east5
```
Molter73, line 61: Should this be a GHA variable instead? You know... So we can change it without opening a PR if we have to update it.

Molter73 (commented):
I've only skimmed through the PR, I'm not very happy about it honestly, looks like a lot more work than it needs to be. ~I'm assuming this was vibe coded and~ (Just noticed the commit author) This needs a lot of cleanup before it can go in IMO.

PR #3021, opened 2026-03-04
ROX-33559: Migrate collector images from UBI-minimal to UBI-micro

janisz:
## Description

Reduce container image size and improve security posture by migrating from UBI minimal base images to UBI micro base images.

  - 31% smaller image (183 MB → 127 MB, saved 56 MB)
  - 40% fewer packages (111 → 67, removed 44 packages)
  - 41% fewer libraries (239 → 141 library files, removed 98)
  - 17% fewer binaries (195 → 162 in /usr/bin, removed 33)

Removed Packages
  
  - GPG/Signing stack (7): gnupg2, gpgme, libgcrypt, libgpg-error, libassuan, libksba, npth
  - GNOME/GObject libs (8): glib2, gobject-introspection, json-glib, libpeas, systemd-libs, libxml2, libunistring, libidn2
  - Package management (8): libmodulemd, librepo, librhsm, lua-libs, lz4-libs, libreport-filesystem, python-unversioned-command, popt
  - Fonts/Localization (5): dejavu-sans-fonts, fonts-filesystem, langpacks-*
  - Storage/Mount (4): libblkid, libmount, libsmartcols, libusbx
  - Network/Security (4): curl-minimal, cyrus-sasl-lib, openldap, gnutls
  - System utilities (5): audit-libs, shadow-utils, libsemanage, libxcrypt-compat, rootfiles
  - Misc (3): file-libs, libevent, libtool-ltdl, nettle, json-c

Refs: 

- https://github.com/stackrox/stackrox/pull/17406

## Checklist
- [ ] Investigated and inspected CI test results
- [ ] Updated documentation accordingly

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

If any of these don't apply, please comment below.

## Testing Performed

CI

PR #3449, opened 2026-06-11
ROX-33133: Re-enable post-quantum crypto-policies for Collector

vladbologa:
## Description

Back-port of https://github.com/stackrox/collector/pull/3448 to the `release-3.25` branch. Required for https://github.com/stackrox/collector/pull/3434 to actually have an effect.

## Checklist
- [x] Investigated and inspected CI test results
- [x] Updated documentation accordingly

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

If any of these don't apply, please comment below.

## Testing Performed

See https://github.com/stackrox/collector/pull/3448

PR #3463, opened 2026-06-16
ROX-35107: Add Konflux pipeline check for post-quantum crypto policy

vladbologa:
## Description

Verify that the built collector image has `X25519MLKEM768` in `/etc/crypto-policies/back-ends/opensslcnf.config`, guarding against regressions of the `DEFAULT:PQ` crypto-policy setting.

I added this check because:
- I added PQC support in https://github.com/stackrox/collector/pull/2977
- it was subsequently removed by mistake in https://github.com/stackrox/collector/pull/3021
- since we currently do not have E2E tests for PQC support, I only discovered this months later, via manual testing
- until we have E2E tests in place, I'd like to have a regression test that is ideally not located in the same place as the code that enables PQC (i.e. not in the Dockerfile), so it's harder to remove it by mistake

Update: Extracted the script into a task here: https://github.com/stackrox/konflux-tasks/pull/105

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [x] Added regression tests

## Testing Performed

CI is sufficient.

### DIFF
--- a/.tekton/collector-component-pipeline.yaml
+++ b/.tekton/collector-component-pipeline.yaml
@@ -580,7 +580,9 @@ spec:
           if grep -q "${EXPECTED_GROUP}" "${CONFIG_FILE}"; then
             echo "PASS: ${EXPECTED_GROUP} found in ${CONFIG_FILE}"
           else
+            echo "Contents of ${CONFIG_FILE}:"
             cat "${CONFIG_FILE}"
+            echo
             echo "FAIL: The image's OpenSSL is not configured to support ML-KEM."
             echo "FAIL: ${EXPECTED_GROUP} not found in ${CONFIG_FILE}"
             exit 1
--- a/.tekton/collector-component-pipeline.yaml
+++ b/.tekton/collector-component-pipeline.yaml
@@ -555,34 +555,34 @@ spec:
       values: [ "false" ]
 
   - name: verify-crypto-policies
-    params:
-    - name: IMAGE_URL
-      value: $(tasks.build-image-index.results.IMAGE_URL)
-    - name: IMAGE_DIGEST
-      value: $(tasks.build-image-index.results.IMAGE_DIGEST)
     when:
     - input: $(params.skip-checks)
       operator: in
       values: [ "false" ]
     taskSpec:
-      params:
-      - name: IMAGE_URL
-        type: string
-      - name: IMAGE_DIGEST
-        type: string
+      description: >-
+        Verify that OpenSSL in the built collector image is configured to support post-quantum
+        cryptography (more specifically, the ML-KEM algorithm group). Regressions can occur e.g.
+        if the crypto-policy setting is removed from the Dockerfile, or the base image is changed
+        to one that does not support post-quantum cryptography.
       steps:
       - name: check-pq-crypto-policy
-        image: $(params.IMAGE_URL)@$(params.IMAGE_DIGEST)
+        image: $(tasks.build-image-index.results.IMAGE_URL)@$(tasks.build-image-index.results.IMAGE_DIGEST)
+        env:
+        - name: EXPECTED_GROUP
+          value: X25519MLKEM768
+        - name: CONFIG_FILE
+          value: /etc/crypto-policies/back-ends/opensslcnf.config
         script: |
           #!/bin/bash
           set -euo pipefail
           echo "Verifying post-quantum crypto policy configuration..."
-          if grep -q X25519MLKEM768 /etc/crypto-policies/back-ends/opensslcnf.config; then
-            echo "PASS: X25519MLKEM768 found in /etc/crypto-policies/back-ends/opensslcnf.config"
+          if grep -q "${EXPECTED_GROUP}" "${CONFIG_FILE}"; then
+            echo "PASS: ${EXPECTED_GROUP} found in ${CONFIG_FILE}"
           else
-            echo "FAIL: X25519MLKEM768 not found in /etc/crypto-policies/back-ends/opensslcnf.config"
-            echo "The post-quantum crypto policy (DEFAULT:PQ) may not be properly applied."
-            cat /etc/crypto-policies/back-ends/opensslcnf.config
+            cat "${CONFIG_FILE}"
+            echo "FAIL: The image's OpenSSL is not configured to support ML-KEM."
+            echo "FAIL: ${EXPECTED_GROUP} not found in ${CONFIG_FILE}"
             exit 1
           fi
 
--- a/.tekton/collector-component-pipeline.yaml
+++ b/.tekton/collector-component-pipeline.yaml
@@ -560,19 +560,19 @@ spec:
       value: $(tasks.build-image-index.results.IMAGE_URL)
     - name: IMAGE_DIGEST
       value: $(tasks.build-image-index.results.IMAGE_DIGEST)
-    when:
-    - input: $(params.skip-checks)
-      operator: in
-      values: [ "false" ]
     taskRef:
       params:
       - name: name
         value: verify-pq-crypto-policies
       - name: bundle
-        value: quay.io/rhacs-eng/konflux-tasks:pr-105@sha256:bbda5ad1b1ffdffb4918d0c12542c0d603060a1690338a6592b0eaa46cf2119b
+        value: quay.io/rhacs-eng/konflux-tasks:latest@sha256:4d05c7ad1bcf63015b6b67787e9f024466fd2c864b69f7939a1925e307afb9b0
       - name: kind
         value: task
       resolver: bundles
+    when:
+    - input: $(params.skip-checks)
+      operator: in
+      values: [ "false" ]
 
   - name: push-dockerfile
     params:
--- a/.tekton/collector-component-pipeline.yaml
+++ b/.tekton/collector-component-pipeline.yaml
@@ -554,39 +554,25 @@ spec:
       operator: in
       values: [ "false" ]
 
-  - name: verify-crypto-policies
+  - name: verify-pq-crypto-policies
+    params:
+    - name: IMAGE_URL
+      value: $(tasks.build-image-index.results.IMAGE_URL)
+    - name: IMAGE_DIGEST
+      value: $(tasks.build-image-index.results.IMAGE_DIGEST)
     when:
     - input: $(params.skip-checks)
       operator: in
       values: [ "false" ]
-    taskSpec:
-      description: >-
-        Verify that OpenSSL in the built collector image is configured to support post-quantum
-        cryptography (more specifically, the ML-KEM algorithm group). Regressions can occur e.g.
-        if the crypto-policy setting is removed from the Dockerfile, or the base image is changed
-        to one that does not support post-quantum cryptography.
-      steps:
-      - name: check-pq-crypto-policy
-        image: $(tasks.build-image-index.results.IMAGE_URL)@$(tasks.build-image-index.results.IMAGE_DIGEST)
-        env:
-        - name: EXPECTED_GROUP
-          value: X25519MLKEM768
-        - name: CONFIG_FILE
-          value: /etc/crypto-policies/back-ends/opensslcnf.config
-        script: |
-          #!/bin/bash
-          set -euo pipefail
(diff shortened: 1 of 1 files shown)


janisz (approved):

.tekton/collector-component-pipeline.yaml
```diff
@@ -554,6 +554,38 @@ spec:
       operator: in
       values: [ "false" ]
 
+  - name: verify-crypto-policies
```
tommartensen: Custom tasks need to be implemented in https://github.com/stackrox/konflux-tasks, otherwise we will get a Conforma violation and cannot release.

I have to assume that you didn't add this as a script after https://github.com/stackrox/collector/blob/master/collector/container/konflux.Dockerfile#L141 because you don't want that in the final image? 

Why is the regression test necessary in the first place?

tommartensen (changes requested):

.tekton/collector-component-pipeline.yaml
```diff
@@ -554,6 +554,38 @@ spec:
...
+    params:
+    - name: IMAGE_URL
+      value: $(tasks.build-image-index.results.IMAGE_URL)
+    - name: IMAGE_DIGEST
+      value: $(tasks.build-image-index.results.IMAGE_DIGEST)
```
msugakov, line 562: Since this is inline task, you can shorten it.
Here's part 1 of 3.

```suggestion
```

.tekton/collector-component-pipeline.yaml
```diff
@@ -554,6 +554,38 @@ spec:
...
+    when:
+    - input: $(params.skip-checks)
+      operator: in
+      values: [ "false" ]
+    taskSpec:
+      params:
+      - name: IMAGE_URL
+        type: string
+      - name: IMAGE_DIGEST
+        type: string
```
msugakov: Part 2 of 3.

```suggestion
```

.tekton/collector-component-pipeline.yaml
```diff
@@ -554,6 +554,38 @@ spec:
...
+      steps:
+      - name: check-pq-crypto-policy
+        image: $(params.IMAGE_URL)@$(params.IMAGE_DIGEST)
```
msugakov: Part 3 of 3.

```suggestion
        image: $(tasks.build-image-index.results.IMAGE_URL)@$(tasks.build-image-index.results.IMAGE_DIGEST)
```

.tekton/collector-component-pipeline.yaml
```diff
@@ -554,6 +554,38 @@ spec:
...
+    params:
+    - name: IMAGE_URL
+      value: $(tasks.build-image-index.results.IMAGE_URL)
+    - name: IMAGE_DIGEST
+      value: $(tasks.build-image-index.results.IMAGE_DIGEST)
+    when:
+    - input: $(params.skip-checks)
+      operator: in
+      values: [ "false" ]
+    taskSpec:
```
msugakov: Please add a short `description:` under `taskSpec` explaining why this task exists. Something in the spirit to prevent regressions.
Example https://github.com/stackrox/stackrox/blob/ca6005ff6c1ffb5b172328edebafa6fa712ce8bf/.tekton/create-custom-snapshot.yaml#L125-L130 , but not necessarily as long.

.tekton/collector-component-pipeline.yaml
```diff
@@ -554,6 +554,38 @@ spec:
...
+      - name: IMAGE_DIGEST
+        type: string
+      steps:
+      - name: check-pq-crypto-policy
+        image: $(params.IMAGE_URL)@$(params.IMAGE_DIGEST)
+        script: |
+          #!/bin/bash
+          set -euo pipefail
+          echo "Verifying post-quantum crypto policy configuration..."
+          if grep -q X25519MLKEM768 /etc/crypto-policies/back-ends/opensslcnf.config; then
```
msugakov: `X25519MLKEM768` is repeated 3 times and `/etc/crypto-policies/back-ends/opensslcnf.config` is repeated four. You can introduce env vars for both, but it makes this tiny code 2 lines longer so I wouldn't insist.

.tekton/collector-component-pipeline.yaml
```diff
@@ -554,6 +554,38 @@ spec:
...
+            echo "PASS: X25519MLKEM768 found in /etc/crypto-policies/back-ends/opensslcnf.config"
+          else
+            echo "FAIL: X25519MLKEM768 not found in /etc/crypto-policies/back-ends/opensslcnf.config"
+            echo "The post-quantum crypto policy (DEFAULT:PQ) may not be properly applied."
+            cat /etc/crypto-policies/back-ends/opensslcnf.config
```
msugakov: 1. WDYT about doing `cat /etc/crypto-policies/back-ends/opensslcnf.config` before `echo`-s?
When something fails, I usually like to find the error cause directly where the screen loads. If it loads at the end of the log, it would make sense to do `echo`-s at the end. If the screen loads the log from the beginning, then it's fine as you have it. 
2. Do you have an example run where this job failed? Please share one, it would be good to see if it fails as expected.

.tekton/collector-component-pipeline.yaml
msugakov: Not necessarily Conforma will fail. If the task executes after the build, we can do whatever, per my recollection of what Konflux folks said.
Only if the task is in the chain before buildah, it matters for Conforma.

Though I agree adding the check to the Dockerfile may be less verbose way to do this.

vladbologa: @tommartensen The PQC policies were removed by mistake during a refactor of the Konflux Dockerfile. See https://github.com/stackrox/collector/pull/3449

So I want to prevent something similar from happening again, since right now we don't have e2e tests for PQC compliance (and will not have for a while).

That's also why I would prefer to have the test somewhere else, and not in the Dockerfile itself. Basically my reasoning was that it's much less likely to remove the code from
here by mistake.

vladbologa: I added an explanation in the PR description (why I want to have this tests, and why I didn't put it in the Dockerfile).

.tekton/collector-component-pipeline.yaml
vladbologa: [Failed run](https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/rh-acs-tenant/applications/acs/pipelineruns/collector-on-push-7zhsm/logs) example:

```
place-scripts :-
2026/06/16 17:44:42 Decoded script /tekton/scripts/script-0-stqgj
prepare :-
2026/06/16 17:44:42 Entrypoint initialization
step-check-pq-crypto-policy :-
Verifying post-quantum crypto policy configuration...
CipherString = @SECLEVEL=2:kEECDH:kRSA:kEDH:kPSK:kDHEPSK:kECDHEPSK:kRSAPSK:-aDSS:-3DES:!DES:!RC4:!RC2:!IDEA:-SEED:!eNULL:!aNULL:!MD5:-SHA384:-CAMELLIA:-ARIA:-AESCCM8
Ciphersuites = TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256:TLS_AES_128_CCM_SHA256
TLS.MinProtocol = TLSv1.2
TLS.MaxProtocol = TLSv1.3
DTLS.MinProtocol = DTLSv1.2
DTLS.MaxProtocol = DTLSv1.2
SignatureAlgorithms = ECDSA+SHA256:ECDSA+SHA384:ECDSA+SHA512:ed25519:ed448:rsa_pss_pss_sha256:rsa_pss_pss_sha384:rsa_pss_pss_sha512:rsa_pss_rsae_sha256:rsa_pss_rsae_sha384:rsa_pss_rsae_sha512:RSA+SHA256:RSA+SHA384:RSA+SHA512:ECDSA+SHA224:RSA+SHA224
Groups = *X25519:secp256r1:X448:secp521r1:secp384r1:ffdhe2048:ffdhe3072:ffdhe4096:ffdhe6144:ffdhe8192
FAIL: The image's OpenSSL is not configured to support ML-KEM.
FAIL: X25519MLKEM768 not found in /etc/crypto-policies/back-ends/opensslcnf.config
```

.tekton/collector-component-pipeline.yaml
```diff
@@ -554,6 +554,38 @@ spec:
...
+        - name: CONFIG_FILE
+          value: /etc/crypto-policies/back-ends/opensslcnf.config
+        script: |
+          #!/bin/bash
+          set -euo pipefail
+          echo "Verifying post-quantum crypto policy configuration..."
+          if grep -q "${EXPECTED_GROUP}" "${CONFIG_FILE}"; then
+            echo "PASS: ${EXPECTED_GROUP} found in ${CONFIG_FILE}"
+          else
+            cat "${CONFIG_FILE}"
```
msugakov: Looking at the log that you shared (https://github.com/stackrox/collector/pull/3463#discussion_r3423750916), I'd suggest adding a couple of things:

```suggestion
            echo "Contents of ${CONFIG_FILE}:"
            cat "${CONFIG_FILE}"
            echo
```

.tekton/collector-component-pipeline.yaml
msugakov: WRT Conforma, I'd suggest that Vlad could do the following: merge this PR, find or create a PR that bumps `COLLECTOR_VERSION` in the StackRox repo, make sure that `COLLECTOR_VERSION` file points to the merge commit or later, and re-run Konflux builds in that PR.
IIRC, we still have `acs-conforma-prod-like` enabled for PRs and so it should show if the Collector image is not kosher. If not, Vlad would need to create a revert of this PR and we need to find another way to solve it (e.g. via setting up `IntegrationTestScenario`).

Would this work for both of you?

msugakov (approved):

.tekton/collector-component-pipeline.yaml
tommartensen: I agree with the plan.

tommartensen (approved):

.tekton/collector-component-pipeline.yaml
vladbologa: But it looks like the `acs-conforma-prod-like` job is failing consistently in all commits in the stackrox repo since ~May 18th.

msugakov: The question would be whether this Collector image will trigger new violations compared to that current baseline.

vladbologa: Ok, let's try this.

.tekton/collector-component-pipeline.yaml
```diff
@@ -554,6 +554,26 @@ spec:
...
+    when:
+    - input: $(params.skip-checks)
+      operator: in
+      values: [ "false" ]
+    taskRef:
+      params:
+      - name: name
+        value: verify-pq-crypto-policies
+      - name: bundle
+        value: quay.io/rhacs-eng/konflux-tasks:pr-105@sha256:bbda5ad1b1ffdffb4918d0c12542c0d603060a1690338a6592b0eaa46cf2119b
```
msugakov: Please don't forget to update it to

```yaml
        value: quay.io/rhacs-eng/konflux-tasks:latest@sha256:<digest>
```

taking `<digest>` value from the build of merge commit of https://github.com/stackrox/konflux-tasks/pull/105 (or any later commit from `master`).

.tekton/collector-component-pipeline.yaml
```diff
@@ -554,6 +554,26 @@ spec:
...
+  - name: verify-pq-crypto-policies
+    params:
+    - name: IMAGE_URL
+      value: $(tasks.build-image-index.results.IMAGE_URL)
+    - name: IMAGE_DIGEST
+      value: $(tasks.build-image-index.results.IMAGE_DIGEST)
+    when:
+    - input: $(params.skip-checks)
+      operator: in
+      values: [ "false" ]
```
msugakov: I'd suggest to move `when:` after `taskRef:` for consistency with others, but it's so minor that you may just skip it.

msugakov (approved):
Looks good. Please update tag and digest before merging.

PR #3849, opened 2026-08-21
feat(ansible): Improve ansible create and destroy vms for ppc64le

mdafsanhossain:
This PR introduces the following improvements for the ppc64le ansible create and destroy tasks

## Description

- Split rhel-ppc64le into rhel-ppc64le-9 and rhel-ppc64le-10 so GitHub Actions can run RHEL 9.6 and RHEL 10 PowerVS tests in parallel (fail-fast: false).
- Move the PowerVS SSH cipher workaround to group_vars/vm_arch_ppc64le.yml so it still applies after the VM type split.
- Wrap PowerVS create in block/always so a failed create still lands in inventory and teardown can delete IBM resources.
- Stop caching PowerVS facts on localhost so instance/network/key IDs cannot leak across VMs.
- Replace cockpit ignore_errors with failed_when that only ignores a missing cockpit.socket unit.
- Make instance delete fail the destroy play if it cannot remove the VM, and delete SSH keys by id instead of passing the whole resource.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

Tested using CI.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -122,8 +122,16 @@ jobs:
 
   ppc64le-integration-tests:
     uses: ./.github/workflows/integration-tests-vm-type.yml
+    strategy:
+      # ensure that if one part of the matrix fails, the
+      # rest will continue
+      fail-fast: false
+      matrix:
+        vm_type:
+          - rhel-ppc64le-9
+          - rhel-ppc64le-10
     with:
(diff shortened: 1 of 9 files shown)


Stringy (approved):
LGTM! thanks @mdafsanhossain

PR #3939, opened 2026-09-08
fix: include custom container plugin and filtering

Stringy:
## Description

To fix the performance regression following the falco update, this PR adds a custom container plugin which is responsible for computing and caching the container ID for each thread. By supporting the plugin system in this way, we can reintroduce the container.id != host filter, which ensures process filtering much earlier, before they get to Collector itself.

## Checklist

**Automated testing**
  - [ ] Added unit tests
  - [ ] Added integration tests
  - [ ] Added regression tests

## Testing Performed

CI tests should be enough; particularly host process test and test process network. Manually run these tests locally (x86) with success.

### DIFF
--- a/collector/container-plugin/ContainerID.h
+++ b/collector/container-plugin/ContainerID.h
@@ -1,0 +1,39 @@
+#pragma once
+
+#include <algorithm>
+#include <cctype>
+#include <optional>
+#include <string_view>
+
+namespace collector::container_plugin {
+
+constexpr size_t kContainerIDLength = 64;
+constexpr size_t kShortContainerIDLength = 12;
+
+inline std::optional<std::string_view> ExtractContainerIDFromCgroup(std::string_view cgroup) {
+  const auto scope = cgroup.rfind(".scope");
+  if (scope != std::string_view::npos) {
+    cgroup.remove_suffix(cgroup.size() - scope);
+  }
+  if (cgroup.size() < kContainerIDLength + 1) {
+    return {};
+  }
+  const auto id_start = cgroup.size() - kContainerIDLength;
+  const char separator = cgroup[id_start - 1];
+  if (separator != '/' && separator != '-' && separator != ':') {
+    return {};
+  }
+  const std::string_view parent = cgroup.substr(0, id_start - 1);
+  constexpr std::string_view kConmonSuffix = "-conmon";
+  if (parent.size() >= kConmonSuffix.size() &&
+      parent.substr(parent.size() - kConmonSuffix.size()) == kConmonSuffix) {
+    return {};
+  }
+  const std::string_view id = cgroup.substr(id_start);
+  if (!std::all_of(id.begin(), id.end(), [](char c) { return std::isxdigit(static_cast<unsigned char>(c)); })) {
+    return {};
+  }
+  return id.substr(0, kShortContainerIDLength);
+}
+
+}  // namespace collector::container_plugin
--- a/collector/container-plugin/ContainerPlugin.cpp
+++ b/collector/container-plugin/ContainerPlugin.cpp
@@ -1,0 +1,260 @@
+#include <cctype>
+#include <cstdint>
+#include <ppm_events_public.h>
+#include <string>
+#include <string_view>
+
+#include <plugin/plugin_api.h>
+
+#include "ContainerID.h"
+
+namespace {
+
+constexpr std::string_view kPluginName = "collector-container";
+constexpr std::string_view kPluginVersion = "0.1.0";
+constexpr std::string_view kHostContainerID = "host";
+
+struct plugin_state {
+  std::string last_error;
+  ss_plugin_table_t* threads = nullptr;
+  ss_plugin_table_field_t* cgroups = nullptr;
+  ss_plugin_table_field_t* cgroup_path = nullptr;
+  ss_plugin_table_field_t* container_id = nullptr;
+  std::string extracted_container_id;
+  const char* extracted_container_id_ptr = nullptr;
+};
+
+std::string ContainerIDFromCgroup(std::string_view cgroup) {
+  return std::string(collector::container_plugin::ExtractContainerIDFromCgroup(cgroup).value_or(std::string_view{}));
+}
+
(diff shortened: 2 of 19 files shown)


collector/lib/system-inspector/Service.cpp
```diff
@@ -119,6 +100,13 @@ bool Service::InitKernel(const CollectorConfig& config) {
     return false;
   }
 
+  sinsp_filter_check_list filter_list;
+  filter_list.add_filter_check(inspector_->new_generic_filtercheck());
+  filter_list.add_filter_check(sinsp_plugin::new_filtercheck(container_plugin_));
+  auto filter_factory = std::make_shared<sinsp_filter_factory>(inspector_.get(), filter_list);
+  sinsp_filter_compiler filter_compiler(filter_factory, "container.id != host");
+  inspector_->set_filter(filter_compiler.compile(), "container.id != host");
```
Molter73, line 108: What's the reason we can't initialize the filter in the same place we used to? Does the kernel driver need to be initialized first?

collector/lib/system-inspector/Service.cpp
```diff
@@ -174,25 +162,14 @@ sinsp_evt* Service::GetNext() {
 bool Service::FilterEvent(sinsp_evt* event) {
   const auto* tinfo = event->get_thread_info();
 
-  return FilterEvent(tinfo);
+  return FilterEvent(*event->get_inspector(), tinfo);
 }
 
-bool Service::FilterEvent(const sinsp_threadinfo* tinfo) {
+bool Service::FilterEvent(sinsp&, const sinsp_threadinfo* tinfo) {
```
Molter73, line 168: What's up with the unused `sinsp&` argument?

collector/lib/CollectorConfig.h
```diff
@@ -225,6 +226,7 @@ class CollectorConfig {
   std::optional<std::string> grpc_server_;
 
   bool disable_process_arguments_ = false;
+  std::string container_plugin_path_ = "/usr/local/lib/collector/collector-container-plugin.so";
```
Molter73, line 229: I see this path is also hardcoded in `collector/lib/CollectorConfig.cpp`, probably worth putting in a global constant somewhere.

collector/lib/Process.cpp
```diff
@@ -32,8 +32,8 @@ const std::shared_ptr<IProcess> ProcessStore::Fetch(uint64_t pid) {
 std::string Process::container_id() const {
   WaitForProcessInfo();
 
-  if (system_inspector_threadinfo_) {
-    auto id = GetContainerID(*system_inspector_threadinfo_);
+  if (system_inspector_ && system_inspector_threadinfo_) {
```
Molter73, line 35: How many system inspectors do we need? Can't we just access the `system_inspector_threadinfo_` indirectly via `system_inspector_`?

collector/lib/Utility.cpp
```diff
@@ -195,50 +205,8 @@ void TryUnlink(const char* path) {
...
-    return {};
-  }
-
-  container_id_part.remove_prefix(1);
-
-  if (!IsContainerID(container_id_part)) {
-    return {};
-  }
-  return std::make_optional(container_id_part.substr(0, SHORT_CONTAINER_ID_LENGTH));
+  return container_plugin::ExtractContainerIDFromCgroup(cgroup);
```
Molter73, line 209: Why do we even need `ExtractContainerIDFromCgroup` here anymore? Why not just call `container_plugin::ExtractContainerIDFromCgroup` wherever it is needed?

collector/container-plugin/ContainerID.h
```diff
@@ -0,0 +1,39 @@
...
+
+namespace collector::container_plugin {
+
+constexpr size_t kContainerIDLength = 64;
+constexpr size_t kShortContainerIDLength = 12;
+
+inline std::optional<std::string_view> ExtractContainerIDFromCgroup(std::string_view cgroup) {
+  const auto scope = cgroup.rfind(".scope");
+  if (scope != std::string_view::npos) {
+    cgroup.remove_suffix(cgroup.size() - scope);
```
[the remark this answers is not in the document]
Molter73: @coderabbit this code has been run in production over multiple years and we have never seen the described behavior from any container engine.

collector/container-plugin/ContainerPlugin.cpp
```diff
@@ -0,0 +1,260 @@
...
+      PPME_SYSCALL_CLONE_20_X,
+      PPME_SYSCALL_FORK_20_X,
+      PPME_SYSCALL_VFORK_20_X,
+      PPME_SYSCALL_CLONE3_X,
+      PPME_SYSCALL_EXECVE_16_X,
+      PPME_SYSCALL_EXECVE_17_X,
+      PPME_SYSCALL_EXECVE_18_X,
+      PPME_SYSCALL_EXECVE_19_X,
+      PPME_SYSCALL_EXECVEAT_X,
+      PPME_SYSCALL_CHROOT_X,
```
Molter73, line 180: Why do we need chroot? Can it somehow change the cgroups a process belongs to?

collector/test/SystemInspectorServiceTest.cpp
```diff
@@ -1,37 +1,44 @@
...
+  const char* plugin_path = std::getenv("ROX_COLLECTOR_CONTAINER_PLUGIN_PATH");
+  ASSERT_NE(plugin_path, nullptr);
+  auto plugin = inspector->register_plugin(plugin_path);
+  std::string error;
+  ASSERT_TRUE(plugin->init("{}", error)) << error;
+  sinsp_filter_check_list filter_list;
+  filter_list.add_filter_check(inspector->new_generic_filtercheck());
+  filter_list.add_filter_check(sinsp_plugin::new_filtercheck(plugin));
+  auto filter_factory = std::make_shared<sinsp_filter_factory>(inspector.get(), filter_list);
+  sinsp_filter_compiler filter_compiler(filter_factory, "container.id != host");
```
Molter73, line 24: If we can't initialize the filter where we used to as I asked on a previous comment, can we at least create a helper method?

